PR #3, opened 2026-02-18
feat: implement replace mode (detection, replacement strategies, visualization)

lipikaramaswamy:
## Summary

Implements the complete replace-mode anonymization pipeline: entity detection (GLiNER + LLM validation/augmentation), four replacement strategies (Redact, Label, Hash, LLM), preview with `display_record()` visualization, and entity occurrence propagation. Ported from research code with architecture aligned to DataDesigner patterns.

## Type of Change
- [x] New feature

## Testing
- [x] Tests pass locally
        - 157 tests, 90% coverage
        - Edge case coverage: malformed JSON, position bounds, overlapping entities, word-boundary matching, unicode, tag notation fallback
        - Pre-commit hooks pass (trailing whitespace, ruff lint, ruff format)

## What's included

**Model defaults** — Default model configs point to `integrate.api.nvidia.com` 
with `NIM_API_KEY`. Users call `Anonymizer()` with no arguments for the default 
setup. Custom endpoints supported via provider YAML. See docs/ for an example.

**Config** — `AnonymizerConfig`, `ReplaceStrategy` (Redact/Label/Hash/LLM with configurable templates), `RewriteParams` + `EvaluationCriteria` (forward-looking for rewrite mode), role-level `ModelSelection`

**Detection** — GLiNER NER → LLM augmentation → LLM validation → entity occurrence propagation. Custom NDD column generators, postprocess with overlap resolution and name splitting

**Replacement** — Local strategies (pure Python) and LLM-backed replacement map generation. `ReplaceRunner` dispatches between them

**Interface** — `Anonymizer` facade with `.run()` / `.preview()`, `display_record()` HTML visualization with entity highlights and replacement map table

**Docs** — contains three scripts that are structured using jupytext for version control and ease of review purposes. These can be run in interactive windows in VSCode / Cursor. They may also be run as python scripts, but record displays will not be satisfying. 

## Workflow
```
Input CSV / DataFrame
       │
       ▼
┌──────────────────────────────┐
│  Entity Detection            │  NemotronPII → Augment → Validate
└──────────────┬───────────────┘
               │
               ▼
┌──────────────────────────────┐
│  Entity Replacement          │  Redact / Label / Hash / LLM
└──────────────┬───────────────┘
               │
               ▼
      PreviewResult / AnonymizerResult
```
## What's not included
1. Documentation - we'll add documentation for entity detection and entity replacement in a separate PR. 
2. Refinement in logging (what granularity of DD logs to pass through by default, how much additional Anonymizer logging to add for clarity)
3. Rewrite mode (full text rewriting with evaluation/repair loop)
4. NemotronPII on build.nvidia.com — currently requires a separate GLiNER endpoint. 
   Notebooks point to a temporary hosted instance.
5. Domain classification (forward-looking config exists but no classifier step as this feeds into rewrite only)
6. Any further algorithmic improvements that @asteier2026 is researching - those will come in smaller PRs once validated.


lipikaramaswamy:

asteier2026 (commented):
Looks great to me. Instigated a few minor questions I sent in slack.

lipikaramaswamy:
Screen grab of the results when running docs/notebook_source in an interactive window on Cursor.
https://github.com/user-attachments/assets/d0ab5925-17fc-4451-915a-1b59e8fbeb23

lipikaramaswamy:
Also not sure why DCO assistant is failing, I've added signed off by to all my commits, but maybe the issue is there is no signatures branch? @binaryaaron any idea why this is behaving differently from NSS?

docs/notebook_source/data/synth_bios_sample10.csv
```diff
@@ -0,0 +1,31 @@
+occupation,bio
+bartender,"Mara Delgado, a 34‑year‑old bartender, has become a familiar face behind the polished mahogany bar at The Lantern, a cozy cocktail lounge tucked into the historic downtown district of Asheville, North Carolina. Born and raised in a small town outside of Raleigh, she moved to Asheville after graduating with a degree in hospitality management, chasing the vibrant food and drink scene she’d only ever read about. Over the past decade, Mara has honed her craft, mixing classic concoctions and inventing her own seasonal drinks that draw regulars and tourists alike. She’s known for her quick wit, a knack for remembering each patron’s favorite order, and an easy‑going demeanor that makes the Lantern feel like a living living room. When she’s not shaking up a new batch of rosemary‑infused gin fizz, you’ll find her at the local farmers’ market, scouting for fresh herbs and local honey to keep her recipes as home‑grown as possible.
+
```
kendrickb-nvidia, line 3: nit: Why are there blank lines between the records? Doesn't that mess up reading this file?

README.md
```diff
@@ -1,154 +1,118 @@
...
+| **RedactReplace** | `[REDACTED_FIRST_NAME]` | `format_template` |
+| **LabelReplace** | `<Alice, first_name>` | `format_template` |
+| **HashReplace** | `<HASH_FIRST_NAME_3bc51062973c>` | `format_template`, `algorithm`, `digest_length` |
+| **LLMReplace** | `Maya` | `instructions` |
+
+```python
+from anonymizer.config.replace_strategies import RedactReplace, LabelReplace, HashReplace, LLMReplace
+
+# Constant redaction
+AnonymizerConfig(replace=RedactReplace(redact_template="****"))
```
kendrickb-nvidia:
```suggestion
AnonymizerConfig(replace=RedactReplace(format_template="****"))
```

src/anonymizer/engine/replace/replace_runner.py
```diff
@@ -0,0 +1,59 @@
...
+    def run(
+        self,
+        dataframe: pd.DataFrame,
+        *,
+        replace_strategy: ReplaceStrategy,
+        model_configs: list[ModelConfig] | str | Path,
+        model_providers: list[Any] | str | Path | None,
+        selected_models: ReplaceModelSelection,
+        preview_num_records: int | None = None,
+    ) -> tuple[pd.DataFrame, list[FailedRecord]]:
```
kendrickb-nvidia: suggestion (from cursor review): Consider a typed dataclass for the return. Good practice and matches the pattern throughout the rest of the code.

src/anonymizer/config/replace_strategies.py
```diff
@@ -0,0 +1,131 @@
...
+        digest = hashlib.new(self.algorithm, text.encode("utf-8")).hexdigest()[: self.digest_length]
+        return self.format_template.format(text=text, label=label.upper(), digest=digest)
+
+
+class LLMReplace(BaseModel):
+    """Marker config for LLM-backed replacement workflow execution."""
+
+    kind: Literal["llm"] = "llm"
+    model_alias: str = Field(default="text")
+    workflow_id: str = Field(default="replace")
```
kendrickb-nvidia: question (from cursor review): `model_alias` and `workflow_id` appear to be unused, should we remove them. Also okay if these are forward looking fields that will be used later. And I haven't dug into if they get passed through to DD and used there so none of the anonyizer code uses them directly, maybe worth a comment if that's the case.

src/anonymizer/engine/detection/detection_workflow.py
```diff
@@ -0,0 +1,532 @@
...
+    def identify_latent_entities(
+        self,
+        dataframe: pd.DataFrame,
+        *,
+        model_configs: list[ModelConfig] | str | Path,
+        model_providers: list[Any] | str | Path | None,
+        selected_models: DetectionModelSelection,
+        gliner_detection_threshold: float,
+        entity_labels: list[str] | None = None,
+        privacy_goal: Any | None,
```
kendrickb-nvidia: question: Does this really accept `Any`? Is there an abstract base class we could use for the type? Or maybe a class or two or string?

src/anonymizer/engine/ndd/adapter.py
```diff
@@ -0,0 +1,175 @@
...
+            return self._data_designer
+
+        resolved_providers = self._load_model_providers(model_providers)
+        return DataDesigner(
+            artifact_path=self._data_designer._artifact_path,
+            model_providers=resolved_providers,
+            secret_resolver=self._data_designer.secret_resolver,
+        )
+
+    def _load_model_providers(
```
kendrickb-nvidia: suggestion (from cursor review): Use a shared method for `_load_model_providers` here and `_resolve_model_providers` in anonymizer.py since they do the same thing at first glance.

src/anonymizer/config/default_model_configs/detection.yaml
kendrickb-nvidia, line 1: question: Are these defaults? Can they be overridden, where an why are they used, a README.md might be helpful. Also maybe use `default_model_configs/` directory name?

src/anonymizer/config/anonymizer_config.py
```diff
@@ -0,0 +1,69 @@
...
+
+class AnonymizerInput(BaseModel):
+    """Input source definition for the anonymizer pipeline.
+
+    Format is inferred from file extension (.csv or .parquet).
+    """
+
+    source: str | Path
+    text_column: str = Field(default="text", min_length=1)
+    id_column: str | None = None
```
kendrickb-nvidia: comment:  Okay to skip for this PR, but I expect we'll want pydantic field descriptions for all the config objects that users may interact with.

src/anonymizer/config/anonymizer_config.py
```diff
@@ -0,0 +1,69 @@
...
+    data_summary: str | None = None
+    selected_models: ModelSelection = Field(default_factory=ModelSelection)
+    gliner_detection_threshold: float = Field(default=0.5, ge=0.0, le=1.0)
+
+    # Replace configuration
+    replace: ReplaceStrategy
+
+    # Rewrite/evaluation configuration
+    rewrite: RewriteParams | None = None
+    privacy_goal: PrivacyGoal | None = None
```
kendrickb-nvidia: question: Ah, there is a PrivacyGoal object, curious why this class isn't used as the type at my later comment about the `privacy_goal: Any | None` type annotation being very broad.

src/anonymizer/config/models.py
```diff
@@ -0,0 +1,36 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+from pydantic import BaseModel, Field
+
+
+class DetectionModelSelection(BaseModel):
```
kendrickb-nvidia, line 9: question: These sort of look like the yaml files in `model_configs/` but not quite (and have similar default values), what's the relationship with config objects (and which config objects) and those yaml files?

src/anonymizer/engine/detection/detection_workflow.py
```diff
@@ -0,0 +1,532 @@
...
+
+
+class EntityDetectionWorkflow:
+    """Detection workflow using NDD LLM + custom-column steps."""
+
+    def __init__(self, adapter: NddAdapter, config_dir: Path | None = None) -> None:
+        self._adapter = adapter
+        self._config_dir = config_dir
+
+    def detect_and_validate_entities(
```
kendrickb-nvidia, line 131: question: Should code using `EntityDetectionWorkflow` call `detect_and_validate_entities` directly instead of using `run`? I know you're adding documentation in a followup PR, but some more context in pydocs on these methods would be helpful. At some point we should have a goal that all non-private (ie don't start with `_`) methods are a pydoc comment at least.

Existing pydocs like "Detection workflow using NDD LLM + custom-column steps." on the class tends towards stating what, which you can often gather from the name, and doesn't provide much context for future readers (llm or human).

src/anonymizer/engine/io/reader.py
```diff
@@ -0,0 +1,50 @@
...
+def _load_dataframe(input_data: pd.DataFrame | AnonymizerInput) -> pd.DataFrame:
+    if isinstance(input_data, pd.DataFrame):
+        return input_data.copy()
+    source = Path(str(input_data.source))
+    suffix = source.suffix.lower()
+    if suffix == ".csv":
+        return pd.read_csv(source)
+    if suffix == ".parquet":
+        return pd.read_parquet(source)
+    raise InvalidInputError(f"Unsupported input format: {suffix}. Use .csv or .parquet.")
```
kendrickb-nvidia, line 43: nit: No json or jsonl support yet?

src/anonymizer/engine/io/reader.py
```diff
@@ -0,0 +1,50 @@
...
+import pandas as pd
+
+from anonymizer.config.anonymizer_config import AnonymizerInput
+from anonymizer.engine.detection.constants import COL_TEXT
+from anonymizer.interface.errors import InvalidInputError
+
+
+def read_input(input_data: pd.DataFrame | AnonymizerInput) -> pd.DataFrame:
+    """Load input into a normalized dataframe with canonical internal text column."""
+    dataframe = _load_dataframe(input_data=input_data)
```
kendrickb-nvidia: nit: I find this a bit of a weird structure to have `_load_dataframe()` both load or copy when we immediately branch on the same condition in line 18. Would prefer `_load_dataframe(input_data: AnonymizerInput)` and do the dataframe copy in this method with an else.

Also looks like lines 22-24 and 26-28 are basically the same. What about something like:

```
selected_text_column = None
if isinstance(input_data, AnonymizerInput):
    dataframe = load_dataframe(input_data)
    selected_text_column = input_data.text_column
else:
    dataframe = input_data.copy()
    if "test" in dataframe.columns:
        selected_text_column = "text"

if selected_text_columns:
   _validate_internal_column_collision(...)
   ...

```

Needs some tidying to match logic, but I think that's easier to see what's happening, what's shared and what's different for the two different input types, etc.

src/anonymizer/engine/io/reader.py
```diff
@@ -0,0 +1,50 @@
...
+        if selected_text_column not in dataframe.columns:
+            raise InvalidInputError(f"Input text column '{selected_text_column}' not found.")
+        _validate_internal_column_collision(dataframe, selected_text_column=selected_text_column)
+        dataframe = dataframe.rename(columns={selected_text_column: COL_TEXT})
+        dataframe.attrs["original_text_column"] = selected_text_column
+    elif "text" in dataframe.columns:
+        _validate_internal_column_collision(dataframe, selected_text_column="text")
+        dataframe = dataframe.rename(columns={"text": COL_TEXT})
+        dataframe.attrs["original_text_column"] = "text"
+    return dataframe
```
kendrickb-nvidia, line 24: question: What happens if a dataframe is provided and it doesn't contain a column called "text"? Will later code crash, or does it handle that okay if there's no column named `COL_TEXT` (in which case what was all this setup needed for)?

src/anonymizer/engine/ndd/adapter.py
```diff
@@ -0,0 +1,175 @@
...
+            return df.copy()
+
+        dataframe = df.copy()
+        dataframe[RECORD_ID_COLUMN] = dataframe.apply(self._compute_record_id, axis=1)
+        return dataframe
+
+    @staticmethod
+    def _compute_record_id(row: pd.Series) -> str:
+        serialized = json.dumps(row.to_dict(), default=str, sort_keys=True, separators=(",", ":"))
+        return uuid.uuid5(uuid.NAMESPACE_URL, serialized).hex
```
kendrickb-nvidia, line 114: question (from cursor review): What happens if there are duplicate rows and this produces identical record ids? I guess we want them to map to the same record id since we're using uuid5 instead of uuid4? Probably worth a comment on what's going on.

src/anonymizer/engine/ndd/model_loader.py
```diff
@@ -0,0 +1,73 @@
...
+
+from pathlib import Path
+from typing import Any
+
+from data_designer.config.utils.io_helpers import load_config_file
+
+DEFAULT_CONFIG_DIR = Path(__file__).resolve().parents[2] / "config" / "model_configs"
+
+
+def load_models_config(config_dir: Path = DEFAULT_CONFIG_DIR) -> dict[str, Any]:
```
kendrickb-nvidia: nit: Good python hygiene to not use an object as a default value (due to concerns about mutability), though Path should be immutable so maybe this is okay. It's just jarring when we've been trained to not use an object as a default.

Possibly prefer default of None:

```
def load_models_config(config_dir: Path | None = None) -> dict[str, Any]:
    if not config_dir:
        config_dir = DEFAULT_CONFIG_DIR`
```

src/anonymizer/engine/ndd/model_loader.py
```diff
@@ -0,0 +1,73 @@
...
+    """Load model definitions from models.yaml."""
+    models_file = config_dir / "models.yaml"
+    return _load_yaml_dict(models_file)
+
+
+def load_workflow_selections(workflow_name: str, config_dir: Path = DEFAULT_CONFIG_DIR) -> dict[str, str]:
```
kendrickb-nvidia: nit: There will be a fixed set of defined workflows, right? Consider making `workflow_name` an enum, or do we need to support arbitrary strings?

src/anonymizer/engine/ndd/model_loader.py
```diff
@@ -0,0 +1,73 @@
...
+    """Load selected model aliases for a workflow."""
+    workflow_file = config_dir / f"{workflow_name}.yaml"
+    workflow_config = _load_yaml_dict(workflow_file)
+    selected_models = workflow_config.get("selected_models", {})
+    if not isinstance(selected_models, dict):
+        raise ValueError(f"{workflow_file} must define a top-level 'selected_models' mapping.")
+    return {str(key): str(value) for key, value in selected_models.items()}
+
+
+def load_workflow_config(workflow_name: str, config_dir: Path = DEFAULT_CONFIG_DIR) -> dict[str, Any]:
```
kendrickb-nvidia: question: (For all methods in this file) Is using `dict[str, Any]` a pattern from DD? Would it make sense to have pydantic objects defined for these configs. I guess I'm a bit fuzzy on the relationship between these "configs" and what's in the `config/*.py` files (and my questions about the yaml files and dir name are part of that).

src/anonymizer/engine/replace/llm_replace_workflow.py
```diff
@@ -0,0 +1,218 @@
...
+                    ),
+                    model_alias=replace_alias,
+                    output_format=ReplacementMap,
+                )
+            ],
+            workflow_name="replace-map-generation",
+            preview_num_records=preview_num_records,
+        )
+        output_df = run_result.dataframe.copy()
+        output_df.attrs = {**run_result.dataframe.attrs, **dataframe.attrs}
```
kendrickb-nvidia, line 88: suggestion: (Generally for all code using data frame .attrs) These `.attrs` are not a standard thing (at least that I'm aware of), explaining/documenting somewhere what and why we're passing around through attrs would be helpful. E.g., why wouldn't these attrs be additional fields in the LlmReplaceResult and the run_result type from run_workflow? Why do we need to use basically an untyped arbitrary dictionary in another library's type? What happens when you write a dataframe with .attrs, does anything unusual happen then?

src/anonymizer/engine/replace/llm_replace_workflow.py
```diff
@@ -0,0 +1,218 @@
...
+
+3. Maintain format and type:
+   - Same structure, length patterns, character types
+   - Same demographic characteristics (Indian name → Indian name)
+
+CRITICAL: Every entity MUST have a different synthetic value. Never return original=synthetic.
+""" % (
+        instruction_block,
+        entities_column,
+    )
```
kendrickb-nvidia: comment (also from cursor): I'm struggling to parse these templates and what's happening thoughout the prompt templating code, not specific to this one method.

The mix of jinja and % replacement is messy. I think the jinja is required (and is really for DD). But would recommend using f-strings instead of the legacy % replacement throughout this PR. Although maybe f-strings also don't play nicely with jinja's usage of `{{ }}`. Since % replacement is a legacy method in python I was super confused for a while about whether instruction_block and entities_column were even being used here, before remembering I need to look for the first and second instances of `%s`.

src/anonymizer/interface/anonymizer.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+class Anonymizer:
+    """Public facade for anonymization workflows."""
+
+    def __init__(
+        self,
+        *,
+        data_designer: DataDesigner | None = None,
+        artifact_path: str | Path = ".anonymizer-artifacts",
```
kendrickb-nvidia: nit: Another place where it's more pythonic in my opinion to use a default of `None` and then have an if to use ".anonymizer-artifacts" in the method code.

src/anonymizer/interface/anonymizer.py
```diff
@@ -0,0 +1,180 @@
...
+        model_providers: list["ModelProvider"] | str | Path | None = None,
+        detection_workflow: EntityDetectionWorkflow | None = None,
+        replace_runner: ReplaceRunner | None = None,
+    ) -> None:
```
kendrickb-nvidia, line 42: suggestion: Definitely need pydocs for the public methods of the public interface. What are all these arguments I can provide, should I provide them?

src/anonymizer/interface/anonymizer.py
```diff
@@ -0,0 +1,180 @@
...
+    """Public facade for anonymization workflows."""
+
+    def __init__(
+        self,
+        *,
+        data_designer: DataDesigner | None = None,
+        artifact_path: str | Path = ".anonymizer-artifacts",
+        model_providers: list["ModelProvider"] | str | Path | None = None,
+        detection_workflow: EntityDetectionWorkflow | None = None,
+        replace_runner: ReplaceRunner | None = None,
```
kendrickb-nvidia: nit: Just struck me here seeing these side by side, aren't detection and replace similar concepts, different steps in the anonymizer flow? Then using a noun for one and a verb for the other, plus workflow and runner names seems confusing. Or if they are conceptually different then cool, but maybe need to explain somewhere why, and thus what naming pattern to use for future pieces of the engine.

src/anonymizer/interface/anonymizer.py
```diff
@@ -0,0 +1,180 @@
...
+        self._detection_workflow = detection_workflow or EntityDetectionWorkflow(adapter=self._adapter)
+        self._replace_runner = replace_runner or ReplaceRunner(llm_workflow=LlmReplaceWorkflow(adapter=self._adapter))
+
+    def run(
+        self,
+        *,
+        config: AnonymizerConfig,
+        data: pd.DataFrame | AnonymizerInput,
+        model_configs: list[ModelConfig] | str | Path | None = None,
+        model_providers: list[ModelProvider] | str | Path | None = None,
```
kendrickb-nvidia: question: Did we think about which args to pass to `run()` and which to `__init__`()? If yes, then great.

Do we expect folks to call `run()` with different model configs and providers? Though I see they're also optional. Would model config and providers fit better with __init__ and you make a new Anonymizer object if you need to change them?

kendrickb-nvidia (approved):
I read through most of the PR (though skimmed tests) plus had cursor to do some review, but not the most thorough review.

Nothing I have is really blocking, so going ahead and approving from me. And feel free to ignore any comments or address things in a follow-up PR.

docs/notebook_source/data/synth_bios_sample10.csv
lipikaramaswamy: These are carriage breaks in the text field -- bios are two paragraphs.

src/anonymizer/engine/ndd/adapter.py
```diff
@@ -0,0 +1,175 @@
...
+        )
+        return WorkflowRunResult(dataframe=output_df, failed_records=failed_records)
+
+    def _resolve_data_designer(self, model_providers: list[ModelProvider] | str | Path | None) -> DataDesigner:
+        if model_providers is None:
+            return self._data_designer
+
+        resolved_providers = self._load_model_providers(model_providers)
+        return DataDesigner(
+            artifact_path=self._data_designer._artifact_path,
```
andreatnvidia: suggestion: this reaches into `self._data_designer._artifact_path`, which is a private attribute on DD with no public accessor. it works today, but it's a bit fragile if DD ever renames or restructures internals.

I think the underlying reason this is needed is that we're constructing a new `DataDesigner` when `model_providers` differs at run time. in DD itself this pattern doesn't actually exist — providers are set once at init and never swapped.

wdyt about just making `model_providers` an init-only concern on `Anonymizer`? that would remove the need to recreate DD instances entirely, and also takes care of the duplicated provider-loading logic between here and `anonymizer.py`. if per-run override is truly needed, maybe the `Anonymizer` (which already knows `artifact_path`) could be the one doing the reconstruction instead of the adapter.

src/anonymizer/engine/ndd/adapter.py
andreatnvidia: nit: this is basically the same as `_resolve_model_providers` in `anonymizer.py:136-147`. probably worth extracting into a shared utility so the parsing logic lives in one place. (also related to the comment above — if providers become init-only, this method might not be needed here at all.)

src/anonymizer/engine/ndd/adapter.py
```diff
@@ -0,0 +1,175 @@
...
+        model_configs: list[ModelConfig] | str | Path,
+        model_providers: list[ModelProvider] | str | Path | None,
+        columns: list[ColumnConfigT],
+        workflow_name: str,
+        preview_num_records: int | None = None,
+    ) -> WorkflowRunResult:
+        """Run one workflow and return output with missing-record tracking."""
+        workflow_input_df = self._attach_record_ids(df=df)
+        workflow_data_designer = self._resolve_data_designer(model_providers=model_providers)
+        resolved_model_configs = load_model_configs(model_configs)
```
andreatnvidia: question: `run_workflow` calls `load_model_configs(model_configs)` here, but callers like `_inject_detector_params` already resolve configs before passing them in. so for detection workflows, configs get loaded twice — once in the workflow to mutate them, and once here again (as a no-op since it's already a list). not a bug, but maybe `run_workflow` could accept `list[ModelConfig]` only and push the resolution responsibility to the caller? might make the data flow clearer.

src/anonymizer/engine/ndd/adapter.py
```diff
@@ -0,0 +1,175 @@
...
+
+        with tempfile.TemporaryDirectory(prefix=f"anonymizer_{workflow_name}_") as tmp_dir:
+            tmp_path = Path(tmp_dir)
+            seed_path = tmp_path / "seed.parquet"
+            workflow_input_df.to_parquet(seed_path, index=False)
```
andreatnvidia: suggestion: I think DD already has `LocalFileSeedSource.from_dataframe(df, path)` which does exactly this write-to-parquet-and-return-source dance. might simplify this block a bit:

```python
seed_source = LocalFileSeedSource.from_dataframe(workflow_input_df, str(seed_path))
config_builder.with_seed_dataset(seed_source, sampling_strategy=SamplingStrategy.ORDERED)
```

src/anonymizer/engine/detection/detection_workflow.py
```diff
@@ -0,0 +1,532 @@
...
+- Prefer ssn over national_id or account_number if ambiguous
+- Prefer phone_number over fax_number unless "fax" is explicit
+- VIN length 17 is vehicle_identifier, shorter is license_plate (check context for regional variations)
+- Numbers preceded by '$' are monetary values, not entities like age or account number
+- If classified as date_of_birth, verify context is appropriate; otherwise use date
+
+Use the candidates list as the authoritative source for entity values and labels.
+
+Example:
+{%% if %s == "xml" -%%}
```
andreatnvidia: nit (echoing Kendrick's comment): the mix of `%s` positional formatting and Jinja `{%% %%}` is pretty hard to parse — you have to mentally track which `%s` maps where, and the `%%` escaping adds noise. I think f-strings won't work directly because of the `{ }` clash with Jinja, but maybe `.replace()` with named placeholders (like you're already doing for `<<VALID_CLASSES>>` and `<<DATA_SUMMARY>>` lower down) would be more readable? something like using `<<TAG_NOTATION>>` and `<<TAGGED_TEXT>>` consistently instead of the `%s` substitutions. wdyt?

src/anonymizer/engine/replace/llm_replace_workflow.py
andreatnvidia, line 88: question: I see `df.attrs` is used to carry `original_text_column` through the pipeline. I think pandas considers `.attrs` experimental and it's not preserved through many operations (merge, concat, groupby, etc.). it works here because you're mostly doing copies, but it's probably a bit fragile as the pipeline evolves. have you considered wrapping the dataframe + metadata in a small dataclass or namedtuple instead? might be more explicit about what state is being threaded through. no strong opinion though — if it works for the current use case and you're aware of the limitation, maybe fine for now.

src/anonymizer/interface/anonymizer.py
```diff
@@ -0,0 +1,180 @@
...
+        self._adapter = NddAdapter(data_designer=self._data_designer)
+        self._detection_workflow = detection_workflow or EntityDetectionWorkflow(adapter=self._adapter)
+        self._replace_runner = replace_runner or ReplaceRunner(llm_workflow=LlmReplaceWorkflow(adapter=self._adapter))
+
+    def run(
+        self,
+        *,
+        config: AnonymizerConfig,
+        data: pd.DataFrame | AnonymizerInput,
+        model_configs: list[ModelConfig] | str | Path | None = None,
```
andreatnvidia: question (building on Kendrick's comment): I'm wondering about the split between init-time and run-time arguments. `model_configs` and `model_providers` can be passed to both `__init__` and `run()`/`preview()`, which means there are two levels of override. the run-level providers in particular trigger new DD instance creation per workflow (see adapter comment above).

I think maybe `model_providers` makes more sense as init-only (you'd create a new `Anonymizer` if you need different providers), while `model_configs` at the run level is probably fine since it's lightweight. wdyt?

src/anonymizer/__init__.py
```diff
@@ -0,0 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+__version__ = "0.1.0"
```
andreatnvidia, line 6: suggestion: the quickstart requires 3 import lines to get going:

```python
from anonymizer.config.anonymizer_config import AnonymizerConfig, AnonymizerInput
from anonymizer.config.replace_strategies import RedactReplace
from anonymizer.interface.anonymizer import Anonymizer
```

wdyt about re-exporting the main public symbols from `anonymizer/__init__.py`? something like `from anonymizer import Anonymizer, AnonymizerConfig, AnonymizerInput, RedactReplace` would lower the barrier for new users. not blocking, just a nice-to-have for DX.

src/anonymizer/engine/io/reader.py
```diff
@@ -0,0 +1,50 @@
...
+    """Load input into a normalized dataframe with canonical internal text column."""
+    dataframe = _load_dataframe(input_data=input_data)
+    if isinstance(input_data, AnonymizerInput):
+        selected_text_column = input_data.text_column
+        if selected_text_column not in dataframe.columns:
+            raise InvalidInputError(f"Input text column '{selected_text_column}' not found.")
+        _validate_internal_column_collision(dataframe, selected_text_column=selected_text_column)
+        dataframe = dataframe.rename(columns={selected_text_column: COL_TEXT})
+        dataframe.attrs["original_text_column"] = selected_text_column
+    elif "text" in dataframe.columns:
```
andreatnvidia: bug: when a user passes a DataFrame directly (not via `AnonymizerInput`), there's no way to specify which column contains the text. if the DataFrame doesn't have a column named exactly `"text"`, neither branch sets the internal text column, and DD fails later with a generic validation error ("Your configuration contains validation errors") that gives no hint about the actual problem.

I ran into this during smoke testing — `anon.preview(config=config, data=df_with_bio_column)` fails silently. two things probably worth addressing:
1. raise an explicit error here when the DataFrame has no `"text"` column (e.g. `InvalidInputError("DataFrame input must contain a 'text' column, or use AnonymizerInput to specify text_column")`)
2. maybe also accept `AnonymizerInput` wrapping a DataFrame so users can specify `text_column` for in-memory data too? or add a `text_column` param to `run()`/`preview()`

src/anonymizer/engine/detection/detection_workflow.py
```diff
@@ -0,0 +1,532 @@
...
+from anonymizer.engine.ndd.model_loader import resolve_model_alias
+
+
+class ValidationChoice(str, Enum):
+    keep = "keep"
+    reclass = "reclass"
+    drop = "drop"
+
+
+class ValidationDecision(BaseModel):
```
lipikaramaswamy, line 53: @asteier2026 flagged that it'd be useful to see the entity value alongside validation decisions. 

IMO rather than adding it to the ValidationDecision schema here, we should enrich it post-hoc via a CustomColumnConfig step that joins value from COL_VALIDATION_CANDIDATES by id. Keeps the LLM schema lean, avoids echoed/hallucinated values. Tracking here to address.

src/anonymizer/engine/replace/replace_runner.py
```diff
@@ -0,0 +1,59 @@
...
+        if isinstance(replace_strategy, (LabelReplace, RedactReplace, HashReplace)):
```
lipikaramaswamy, line 47: @asteier2026 requested that we perform replace to a user-specified set of labels: this is straightforward. Detection already produces labeled entities; we just need a config field (replace_labels) and a filter in ReplaceRunner before dispatch. Applies uniformly to local and LLM strategies. TBD on including here or fast follow

src/anonymizer/config/default_model_configs/detection.yaml
lipikaramaswamy: Done: renamed, added README, added validation of aliases and tests

src/anonymizer/__init__.py
lipikaramaswamy: Agreed. We held off on top-level re-exports while the public API was still in flux. Now that the interface has stabilized (model_configs on init, AnonymizerConfig is pure workflow behavior, AnonymizerInput for data), I've added re-exports in __init__.py. Thanks for flagging!

src/anonymizer/engine/ndd/adapter.py
lipikaramaswamy: Yup. No longer needed. `model_providers` is now init-only, so the adapter never reconstructs DataDesigner instances. Removed the `_resolve_data_designer` and `_load_model_providers` methods entirely

src/anonymizer/engine/ndd/adapter.py
lipikaramaswamy: Removed from adapter, providers are resolved once at `Anonymizer.__init__`

src/anonymizer/engine/ndd/adapter.py
lipikaramaswamy: Fixed — `parse_model_configs()` runs once at `Anonymizer.__init__`. Workflows receive `list[ModelConfig]`. No more `load_model_configs` calls in the adapter or detection workflow

src/anonymizer/engine/ndd/adapter.py
lipikaramaswamy: Done, thanks!

src/anonymizer/engine/detection/detection_workflow.py
lipikaramaswamy: All %s replaced with named <<PLACEHOLDER>> tags.

src/anonymizer/engine/replace/llm_replace_workflow.py
lipikaramaswamy: Documented with a comment and TODO. Agreed the right fix is wrapping df + metadata in a typed container — tracking for a follow-up.

src/anonymizer/engine/io/reader.py
lipikaramaswamy: Fixed by removing the DataFrame input path entirely. `run()`/`preview()` now only accept `AnonymizerInput`, which always specifies text_column. No more silent failures.

src/anonymizer/engine/replace/replace_runner.py
lipikaramaswamy: Makes sense, added typed dataclass returns throughout: ParsedModelConfigs for parse_model_configs() and ReplacementResult for ReplacementWorkflow.run().

src/anonymizer/config/replace_strategies.py
lipikaramaswamy: Removed both fields. Model alias resolution is handled by selected_models (from YAML), and workflow naming is hardcoded in the adapter.

src/anonymizer/engine/detection/detection_workflow.py
lipikaramaswamy: Changed to `PrivacyGoal | None` throughout.

src/anonymizer/engine/ndd/adapter.py
lipikaramaswamy: Removed `_load_model_providers` from the adapter entirely. `model_providers` is now init-only on Anonymizer, so the adapter no longer needs its own provider resolution.

src/anonymizer/config/anonymizer_config.py
lipikaramaswamy: Added `description=` to all user-facing pydantic fields. Only `RewriteParams` remains to be fleshed out - will do when we add that pipeline.

src/anonymizer/config/anonymizer_config.py
lipikaramaswamy: Done done!

src/anonymizer/config/models.py
lipikaramaswamy: Resolved by making YAML the single source of truth. Pydantic *Selection classes are now pure schema (no hardcoded defaults). `load_default_model_selection()` reads YAML and constructs the models. Added documentation to `default_model_configs/README.md`.

src/anonymizer/engine/detection/detection_workflow.py
lipikaramaswamy: Added docstrings to `detect_and_validate_entities`, `identify_latent_entities`, and `run` explaining what each does and how they relate. Also added docstrings to `Anonymizer.__init__`, `run()`, `preview()`.

src/anonymizer/engine/io/reader.py
lipikaramaswamy: Not yet — tracking for a follow-up.

src/anonymizer/engine/io/reader.py
lipikaramaswamy: The DataFrame code path has been removed entirely. `read_input()` now only accepts `AnonymizerInput`, which always specifies `text_column`. No more branching or implicit text column assumption.

src/anonymizer/engine/io/reader.py
lipikaramaswamy: No longer possible

src/anonymizer/engine/ndd/adapter.py
lipikaramaswamy: Row index is now included in the hash input so duplicate rows get distinct record IDs. Added a comment explaining the uuid5 determinism rationale and a TODO to validate and use `id_column` from `AnonymizerInput` if provided.

src/anonymizer/engine/ndd/model_loader.py
lipikaramaswamy: Fixed.

src/anonymizer/engine/ndd/model_loader.py
lipikaramaswamy: Added `WorkflowName` enum with detection, replace, rewrite. Used throughout `model_loader.py`.

src/anonymizer/engine/ndd/model_loader.py
lipikaramaswamy: Kept `dict[str, Any]` with docstrings explaining why — these are raw YAML dicts passed to DD's `load_model_configs()` for deserialization

src/anonymizer/engine/replace/llm_replace_workflow.py
lipikaramaswamy: Added a comment documenting what .attrs carries (original_text_column), its experimental pandas status, and a TODO to wrap df + metadata in a typed container.

src/anonymizer/engine/replace/llm_replace_workflow.py
lipikaramaswamy: Replaced all %s substitutions with named <<PLACEHOLDER>> tags and .replace() in all three prompt functions. Consistent with the existing <<VALID_CLASSES>>/<<DATA_SUMMARY>> pattern. Also removed the %% escaping since we no longer use % string formatting.

src/anonymizer/interface/anonymizer.py
lipikaramaswamy: Fixed — `artifact_path` defaults to `None` with internal fallback to `".anonymizer-artifacts"`.

src/anonymizer/interface/anonymizer.py
lipikaramaswamy: Added docstrings with parameter descriptions

src/anonymizer/interface/anonymizer.py
lipikaramaswamy: Renamed `ReplaceRunner` - which is an orchestrator - to `ReplacementWorkflow` for consistency with `EntityDetectionWorkflow.`

src/anonymizer/interface/anonymizer.py
lipikaramaswamy: Redesigned: `model_configs` (unified YAML with model pool + role selections) and `model_providers` live on `Anonymizer.__init__`. `AnonymizerConfig` is pure workflow behavior. `run()`/`preview()` take only config + data.

andreatnvidia (approved):
Approved — this is in good shape and the latest updates addressed my review feedback.

I did a final pass and smoke-tested the current head (including live GLiNER endpoint usage), and behavior looks solid: tests/lint are clean, detection+replacement flows run successfully, and the API/data-flow changes are clearer now (especially init-only provider/model setup and workflow consistency).

For follow-up architecture (non-blocking for this PR), I’d suggest we track a next step to harden the contract toward DataDesigner production posture:
- add an explicit `Anonymizer.validate(...)` preflight (model alias/provider/input/schema/strategy compatibility checks),
- replace implicit `DataFrame.attrs` metadata threading with a typed pipeline context object,
- wire `AnonymizerInput.id_column` into record tracking (with deterministic hash fallback),
- externalize long prompts into versioned files, and
- add a lightweight extension interface for detection/replacement workflow components.

That would reduce coupling and make this easier to evolve as runtime/plugin complexity grows.

lipikaramaswamy:
> Approved — this is in good shape and the latest updates addressed my review feedback.
> 
> I did a final pass and smoke-tested the current head (including live GLiNER endpoint usage), and behavior looks solid: tests/lint are clean, detection+replacement flows run successfully, and the API/data-flow changes are clearer now (especially init-only provider/model setup and workflow consistency).
> 
> For follow-up architecture (non-blocking for this PR), I’d suggest we track a next step to harden the contract toward DataDesigner production posture:
> 
> * add an explicit `Anonymizer.validate(...)` preflight (model alias/provider/input/schema/strategy compatibility checks),
> * replace implicit `DataFrame.attrs` metadata threading with a typed pipeline context object,
> * wire `AnonymizerInput.id_column` into record tracking (with deterministic hash fallback),
> * externalize long prompts into versioned files, and
> * add a lightweight extension interface for detection/replacement workflow components.
> 
> That would reduce coupling and make this easier to evolve as runtime/plugin complexity grows.

Thanks! And all except the last two are tracked in TODOs. I'll start populating our board too so we can knock these items out quickly with smaller PRs.

lipikaramaswamy:

src/anonymizer/engine/detection/detection_workflow.py
lipikaramaswamy: tracking for follow up

src/anonymizer/engine/replace/replace_runner.py
lipikaramaswamy: tracking for follow up

PR #17, opened 2026-02-27
fix: resolve replaced highlight drift

### LINKED ISSUE #15: fix: entity highlights misaligned in replaced text display due to cursor drift in _build_replaced_entities
### Priority Level

High (Major functionality broken)

### Describe the bug

Entity highlight spans in the **Replaced** section of `display_record()` are visually misaligned — boxes start mid-word and include trailing punctuation. The second paragraph is more severely affected than the first, consistent with cumulative drift.

Example of what's observed:
- `So|fia. | first_name` instead of `Sofia | first_name`
- `Di|ego, | first_name` instead of `Diego | first_name`
- `Is|abel, | first_name` instead of `Isabel | first_name`

### Root Cause

The replaced text is built in `src/anonymizer/engine/replace/strategies.py:93` (`_apply_replacement_map_to_text`) using entity positions from `COL_DETECTED_ENTITIES`. `_build_replaced_entities` in `src/anonymizer/interface/display.py:119` then tries to reconstruct entity positions in the replaced text by replaying the same cursor logic.

Both functions perform a `by_value` lookup to find the synthetic value for each entity. Their fallbacks differ when an entity is **not found in the replacement map**:

```python
# strategies.py:125 — used to BUILD the replaced text
synthetic = by_value_label.get((value, label), by_value.get(value, text[start:end]))
#                                                                        ^^^^^^^^^^^^
#                                                          actual text span (may include whitespace)

# display.py:156 — used to COMPUTE highlight positions
synthetic = by_value_label.get((value, label), by_value.get(value, value))
#                                                                    ^^^^^
#                                                       entity.value (stripped by EntitySpan)
```

If `entity.value` and `text[start:end]` have different lengths (e.g. due to whitespace, or a case variant produced by `expand_entity_occurrences` using `re.IGNORECASE` in `postprocess.py:393`), and that entity is absent from the replacement map (e.g. LLM partial output, or an expanded occurrence whose value doesn't match any map key), the two cursors diverge. Drift accumulates with each such entity, explaining why the second paragraph is more severely affected.

A related regression is already documented and tested in `tests/interface/test_display.py:234-263` (`test_render_record_html_replaced_tags_positioned_correctly_with_case_mismatch`), which covers the case where `entity.value` differs in casing from the text. The current bug appears to be a variant of this same drift mechanism.

Run `display_record()` on a result produced by `LLMReplace` strategy on a multi-paragraph text where entity values appear multiple times with mixed casing (names like "Sofia" that also appear lowercased).

### Expected behavior

Entity highlight boxes in the Replaced section should align exactly with the synthetic replacement values in the replaced text — no mid-word splits, no drift across paragraphs.

### Additional context

- Bug visible in screenshot from early tester — second paragraph is more severely misaligned than the first.
- `_apply_replacement_map_to_text` (`strategies.py:93`) and `_build_replaced_entities` (`display.py:119`) must be kept in sync. The fix should either unify the fallback logic or have `_build_replaced_entities` directly locate synthetic values in the replaced text rather than replaying cursor arithmetic.
- Existing regression test: `tests/interface/test_display.py:234`.

lipikaramaswamy:
## Summary

- Fix replaced-text highlight drift in `display_record()` by resolving replacement spans via forward search in replaced text instead of replaying cursor arithmetic.
- Add case-insensitive replacement-map lookup/fallback so propagated entities (for example, "the Lantern" vs "The Lantern") map correctly without label offset corruption.
- Add regression tests for drift scenarios: case mismatch and entities absent from replacement map.
- Fix notebook source path handling in `docs/notebook_source/01_replace.py` and `docs/notebook_source/02_detection_debug.py` so .ipynb execution resolves data from `docs/notebook_source/data`.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing
- [x] Tests pass locally
- [x] Manually verify preview rendering in replace notebook

image

## Related Issues
Closes #15

### DIFF
--- a/docs/notebook_source/01_replace.py
+++ b/docs/notebook_source/01_replace.py
@@ -28,9 +28,10 @@ import tempfile
 from pathlib import Path
 
 try:
-    NOTEBOOK_DIR = Path(__file__).resolve().parent
+    NOTEBOOK_SOURCE_DIR = Path(__file__).resolve().parent
 except NameError:
-    NOTEBOOK_DIR = Path.cwd()
+    # Running as .ipynb — cwd is docs/notebooks/, data lives in docs/notebook_source/
+    NOTEBOOK_SOURCE_DIR = Path.cwd().parent / "notebook_source"
 
 from anonymizer import Annotate, Anonymizer, AnonymizerConfig, AnonymizerInput, Hash, Redact, Substitute
 
@@ -93,7 +94,7 @@ anonymizer = Anonymizer(model_configs=MODEL_CONFIGS_YAML, model_providers=provid
 
 # %%
 input_data = AnonymizerInput(
-    source=str(NOTEBOOK_DIR / "data" / "synth_bios_sample10.csv"),
+    source=str(NOTEBOOK_SOURCE_DIR / "data" / "synth_bios_sample10.csv"),
     text_column="bio",
     data_summary="Biographical profiles",
 )
--- a/docs/notebook_source/02_detection_debug.py
+++ b/docs/notebook_source/02_detection_debug.py
@@ -28,9 +28,10 @@ from collections import Counter
 from pathlib import Path
 
 try:
-    NOTEBOOK_DIR = Path(__file__).resolve().parent
+    NOTEBOOK_SOURCE_DIR = Path(__file__).resolve().parent
 except NameError:
-    NOTEBOOK_DIR = Path.cwd()
+    # Running as .ipynb — cwd is docs/notebooks/, data lives in docs/notebook_source/
+    NOTEBOOK_SOURCE_DIR = Path.cwd().parent / "notebook_source"
 
 import pandas as pd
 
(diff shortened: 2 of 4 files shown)


andreatnvidia (approved):
Nice fix! The find-based approach is much more robust than the cursor arithmetic.

One small nit: `_resolve_synthetic` takes 4 separate dict params which makes the signature pretty wide. Could be bundled into a dataclass or namedtuple, but totally not a blocker.

asteier2026 (approved):

asteier2026 (approved):

lipikaramaswamy:
> Nice fix! The find-based approach is much more robust than the cursor arithmetic.
> 
> One small nit: `_resolve_synthetic` takes 4 separate dict params which makes the signature pretty wide. Could be bundled into a dataclass or namedtuple, but totally not a blocker.

Thanks! Implemented this cleanup in display.py by introducing a small frozen `_SyntheticLookupMaps` dataclass and updating `_resolve_synthetic()` accordingly. Tests pass and the display is as expected!

Issue #14, opened 2026-02-25
feat: add user-facing logging with pipeline stage progress and entity stats


lipikaramaswamy:
### Priority Level

High (Major improvement)

### Is your feature request related to a problem?

When running `Anonymizer.run()` or `Anonymizer.preview()` on a non-trivial dataset, users have no visibility into what's happening. The pipeline is silent until it either returns or raises — making it impossible to tell if a long-running job is progressing normally or has stalled.

### Describe the solution you'd like

Add a curated logging layer (using Python's standard `logging` module under an `anonymizer` logger) that surfaces meaningful progress signals at each stage of `_run_internal` in `src/anonymizer/interface/anonymizer.py:108`.

NDD already emits detailed debug logs internally. The goal here is not to replicate those, but to select the subset of signals that give an Anonymizer user confidence that things are moving along. Suggested log points:

| Stage | Log message |
|---|---|
| After `read_input` (line 115) | `Loaded N records from <source>` |
| Before detection workflow (line 117) | `Running entity detection on N records` |
| After detection workflow (line 127) | `Detection complete — X entities found across N records (Y failed)` |
| Before replacement (line 129) | `Running entity replacement` |
| After replacement (line 135) | `Replacement complete (Y failed)` |
| End of `_run_internal` | `Pipeline complete — N records processed, F total failures` |

Failed record counts should come from `WorkflowRunResult.failed_records` already returned by `NddAdapter.run_workflow` in `src/anonymizer/engine/ndd/adapter.py:97`.

### Describe alternatives you've considered

- Tqdm progress bars — rejected because meaningful progress tracking requires NDD to expose batch-level callbacks, which is a DataDesigner concern.
- Printing directly — rejected in favour of `logging` so users can control log level and routing.

### Additional context

The logger should be namespaced (e.g. `logging.getLogger("anonymizer")`) so library users can opt in or out cleanly without global side effects.

PR #18, opened 2026-02-27
fix: harden io path validation and error handling

### LINKED ISSUE #12: fix: add input path validation in reader and writer
### Priority Level

Medium (Annoying but has workaround)

### Describe the bug

`_load_dataframe` in `src/anonymizer/engine/io/reader.py:37` and `write_output` in `src/anonymizer/engine/io/writer.py:15` accept file paths without any validation. The path is wrapped in `Path()` and passed directly to pandas — no canonicalization, no existence check on read, and no restriction on write destination.

```python
# reader.py:37
source = Path(str(input_data.source))

# writer.py:15
path = Path(output_path)
```

```python
anonymizer.run(
    config=config,
    data=AnonymizerInput(source="../../sensitive_dir/data.csv", text_column="text"),
)
```

No error or warning is raised before pandas attempts to open the path.

### Expected behavior

- On read: validate that the path exists and has a supported extension before attempting to open it, with a clear `InvalidInputError` if not.
- On write: validate the output directory exists (or create it), and surface a clear error if the path is not writable, rather than letting pandas raise a raw `OSError`.

### Additional context

Both `reader.py` and `writer.py` already import `InvalidInputError` and use it for unsupported format errors — path validation would fit naturally in `_load_dataframe` (line 36) and `write_output` (line 13).

lipikaramaswamy:
## Summary
- Add input path validation in `read_input` flow: fail fast when path does not exist or is not a file.
- Normalize read/write filesystem and pandas failures into clear `InvalidInputError` messages with path context.
- Ensure output parent directories are created before writing, and add regression tests for these edge cases.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing
- [x] Tests pass locally
  - Verified new failure modes return `InvalidInputError` instead of raw OS/pandas errors

## Related Issues
Closes #12

### DIFF
--- a/src/anonymizer/config/anonymizer_config.py
+++ b/src/anonymizer/config/anonymizer_config.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import logging
+from pathlib import Path
 
 from pydantic import BaseModel, Field, field_validator, model_validator
 
@@ -30,6 +31,16 @@ class AnonymizerInput(BaseModel):
     data_summary: str | None = Field(
         default=None, description="Short description of the data. Improves LLM detection accuracy."
     )
+
+    @field_validator("source")
+    @classmethod
+    def validate_source_path(cls, value: str) -> str:
+        source = Path(value)
+        if not source.exists():
+            raise ValueError(f"Input path does not exist: {source}")
+        if not source.is_file():
+            raise ValueError(f"Input path is not a file: {source}")
+        return value
 
 
 class Detect(BaseModel):
--- a/src/anonymizer/engine/io/constants.py
+++ b/src/anonymizer/engine/io/constants.py
@@ -1,0 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+SUPPORTED_IO_FORMATS: tuple[str, ...] = (".csv", ".parquet")
--- a/src/anonymizer/engine/io/reader.py
+++ b/src/anonymizer/engine/io/reader.py
@@ -9,7 +9,8 @@ import pandas as pd
 
 from anonymizer.config.anonymizer_config import AnonymizerInput
 from anonymizer.engine.constants import COL_TEXT
-from anonymizer.interface.errors import InvalidInputError
+from anonymizer.engine.io.constants import SUPPORTED_IO_FORMATS
+from anonymizer.interface.errors import AnonymizerIOError, InvalidInputError
 
 
 def read_input(input_data: AnonymizerInput) -> pd.DataFrame:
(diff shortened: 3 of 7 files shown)


andreatnvidia (commented):
**`src/anonymizer/engine/io/reader.py:49`**

`InvalidInputError` is used for both validation problems and IO failures here. If a caller catches it to handle bad config, it'll also catch unrelated IO problems like permission errors or corrupt files.

Could be better to add an `AnonymizerIOError(AnonymizerError)` for the try/except blocks and keep `InvalidInputError` for the validation cases. Not blocking, just a thought for a follow-up.

---

**`src/anonymizer/engine/io/reader.py:38`**

nit: these `exists()` + `is_file()` checks could live as a Pydantic `@field_validator` on `AnonymizerInput.source`, so bad paths get caught at config time instead of read time. Keeps `_load_dataframe` focused on just parsing.

---

**`src/anonymizer/engine/io/reader.py:44` / `writer.py:21`**

suggestion: adding `logger.debug` before reads/writes and `logger.exception` in the except blocks would help with debugging without adding noise.

---

**`src/anonymizer/engine/io/writer.py:17`**

nit: `{".csv", ".parquet"}` shows up in both reader and writer. Could pull it into a shared constant like `SUPPORTED_FORMATS` so adding a format later is a one-line change.

asteier2026 (approved):

lipikaramaswamy:
> **`src/anonymizer/engine/io/reader.py:49`**
> 
> `InvalidInputError` is used for both validation problems and IO failures here. If a caller catches it to handle bad config, it'll also catch unrelated IO problems like permission errors or corrupt files.
> 
> Could be better to add an `AnonymizerIOError(AnonymizerError)` for the try/except blocks and keep `InvalidInputError` for the validation cases. Not blocking, just a thought for a follow-up.
> 
> **`src/anonymizer/engine/io/reader.py:38`**
> 
> nit: these `exists()` + `is_file()` checks could live as a Pydantic `@field_validator` on `AnonymizerInput.source`, so bad paths get caught at config time instead of read time. Keeps `_load_dataframe` focused on just parsing.
> 
> **`src/anonymizer/engine/io/reader.py:44` / `writer.py:21`**
> 
> suggestion: adding `logger.debug` before reads/writes and `logger.exception` in the except blocks would help with debugging without adding noise.
> 
> **`src/anonymizer/engine/io/writer.py:17`**
> 
> nit: `{".csv", ".parquet"}` shows up in both reader and writer. Could pull it into a shared constant like `SUPPORTED_FORMATS` so adding a format later is a one-line change.

Thanks @andreatgretel for the thoughtful review. I've addressed 3/4 items - 

- Added shared IO format constants so reader/writer don’t duplicate suffix definitions.
- Moved `source` path checks (`exists` / `is_file`) into `AnonymizerInput` via Pydantic `@field_validator`, so invalid paths fail at config/input creation time.
- Introduced `AnonymizerIOError` and now use it for read/write runtime failures, while keeping `InvalidInputError` for validation-style issues.

Let's defer logging changes to a dedicated follow-up PR so we can apply a consistent logging policy across the codebase in one diff. Can capture in #14 or a follow up to it.

PR #20, opened 2026-02-27
fix: surface JSON parse failures for observability

### LINKED ISSUE #13: fix: surface silent JSON parse failures in postprocessing pipeline
### Priority Level

High (Major functionality broken)

### Describe the bug

`_safe_json_loads` in `src/anonymizer/engine/detection/postprocess.py:353-362` silently returns `{}` on any `json.JSONDecodeError`. Two critical functions depend on it and silently fall back to no-ops when the NDD model returns malformed output:

- `apply_validation_decisions` (line 113): if `_safe_json_loads` returns `{}`, `decisions` becomes `[]` and all entities are kept unchanged — as if validation never ran.
- `apply_augmented_entities` (line 162): if parsing fails, `augmented` becomes `[]` and augmentation is silently skipped.

```python
# postprocess.py:353-362
def _safe_json_loads(value: dict | str) -> dict:
    ...
    except json.JSONDecodeError:
        return {}  # callers have no way to know this happened

# postprocess.py:113-114
payload = _safe_json_loads(validation_output)
decisions = payload.get("decisions", []) if isinstance(payload, dict) else []
```

If an LLM model returns a malformed or truncated JSON response for the validation step, the pipeline completes without error and the result looks normal — but no validation decisions were actually applied.

### Expected behavior

Parse failures should be distinguishable from legitimate empty responses. Options:
- Raise a structured exception that the caller can catch and log/track as a failed record.
- At minimum, add a boolean flag or warning that indicates the parse failed, so the trace dataframe can surface it (e.g., a `_validation_parse_failed` column).

### Additional context

The `FailedRecord` dataclass in `src/anonymizer/engine/ndd/adapter.py:28-33` already exists to track per-record failures — JSON parse failures in postprocessing are a natural fit for that mechanism.

lipikaramaswamy:
## Summary

Malformed JSON responses from the LLM previously failed silently in key postprocessing steps (`apply_validation_decisions` and `apply_augmented_entities`), making diagnosis difficult.  
This change improves observability by:
- Adding warning logs when `_safe_json_loads` encounters malformed JSON in post processing, so failures are visible instead of silently acting as no-ops.
- Preserving existing fallback behavior (`{}`) to avoid changing runtime control flow.
- Redacting payload content from logs by emitting only safe metadata (`error` and `length`) instead of raw JSON previews.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing
- [x] Tests pass locally

Closes #13

### DIFF
--- a/src/anonymizer/engine/detection/postprocess.py
+++ b/src/anonymizer/engine/detection/postprocess.py
@@ -361,8 +361,12 @@ def _safe_json_loads(value: dict | str) -> dict:
     try:
         parsed = json.loads(value)
         return parsed if isinstance(parsed, dict) else {}
-    except json.JSONDecodeError:
-        logger.warning("Failed to parse JSON in postprocessing pipeline: %.120r", value)
+    except json.JSONDecodeError as exc:
+        logger.warning(
+            "Failed to parse JSON in postprocessing pipeline (error=%s, length=%d)",
+            exc.msg,
+            len(value),
+        )
         return {}
 
 
--- a/src/anonymizer/engine/detection/postprocess.py
+++ b/src/anonymizer/engine/detection/postprocess.py
@@ -4,10 +4,13 @@
 from __future__ import annotations
 
 import json
+import logging
 import re
 from dataclasses import dataclass
 from enum import Enum
 from hashlib import sha1
+
+logger = logging.getLogger(__name__)
 
 VALIDATION_CONTEXT_WINDOW = 32
 
@@ -359,6 +362,7 @@ def _safe_json_loads(value: dict | str) -> dict:
         parsed = json.loads(value)
         return parsed if isinstance(parsed, dict) else {}
     except json.JSONDecodeError:
+        logger.warning("Failed to parse JSON in postprocessing pipeline: %.120r", value)
         return {}
 
 
--- a/tests/engine/test_detection_postprocess.py
+++ b/tests/engine/test_detection_postprocess.py
@@ -636,13 +636,11 @@ def test_expand_case_insensitive_matching() -> None:
     assert len(expanded) == 2
 
 
-def test_safe_json_loads_logs_warning_on_invalid_json(caplog: pytest.LogCaptureFixture) -> None:
-    from anonymizer.engine.detection.postprocess import _safe_json_loads
-
+def test_parse_raw_entities_logs_warning_on_malformed_json(caplog: pytest.LogCaptureFixture) -> None:
     payload = '{"name":"Alice","ssn":"123-45-6789",invalid}'
     with caplog.at_level(logging.WARNING, logger="anonymizer.engine.detection.postprocess"):
-        result = _safe_json_loads(payload)
-    assert result == {}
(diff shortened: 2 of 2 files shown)


asteier2026 (approved):

tests/engine/test_detection_postprocess.py
```diff
@@ -631,3 +634,15 @@ def test_expand_case_insensitive_matching() -> None:
...
+def test_safe_json_loads_logs_warning_on_invalid_json(caplog: pytest.LogCaptureFixture) -> None:
+    from anonymizer.engine.detection.postprocess import _safe_json_loads
+
+    payload = '{"name":"Alice","ssn":"123-45-6789",invalid}'
+    with caplog.at_level(logging.WARNING, logger="anonymizer.engine.detection.postprocess"):
+        result = _safe_json_loads(payload)
+    assert result == {}
+    assert any("Failed to parse JSON" in m for m in caplog.messages)
+    assert any("length=" in m for m in caplog.messages)
+    assert payload not in "\n".join(caplog.messages)
```
andreatnvidia, line 646: nit: might be worth testing this through a public function instead of importing `_safe_json_loads` directly. `parse_raw_entities` passes its input straight to `_safe_json_loads`, and there's already a malformed JSON test at line 211 you could piggyback on:

```python
def test_parse_raw_entities_logs_warning_on_malformed_json(caplog: pytest.LogCaptureFixture) -> None:
    with caplog.at_level(logging.WARNING, logger="anonymizer.engine.detection.postprocess"):
        result = parse_raw_entities(raw_response="not json", text="hello")
    assert result == []
    assert any("Failed to parse JSON" in m for m in caplog.messages)
```

Also the `_safe_json_loads` import is inside the function body -- could move it to the top with the other imports from that module.

src/anonymizer/engine/detection/postprocess.py
```diff
@@ -358,7 +361,12 @@ def _safe_json_loads(value: dict | str) -> dict:
     try:
         parsed = json.loads(value)
         return parsed if isinstance(parsed, dict) else {}
-    except json.JSONDecodeError:
+    except json.JSONDecodeError as exc:
+        logger.warning(
+            "Failed to parse JSON in postprocessing pipeline (error=%s, length=%d)",
+            exc.msg,
+            len(value),
+        )
```
andreatnvidia, line 369: suggestion: could be nice to add `exc_info=True` here so the full traceback (including line/column position from `JSONDecodeError`) is available when debugging. No need to address here, we'll pick this up in the logging work.

andreatnvidia (approved):
LGTM just added some nits!

src/anonymizer/engine/detection/postprocess.py
lipikaramaswamy: Actually intentionally omitted that to ensure we don't output anything unsafe.

tests/engine/test_detection_postprocess.py
lipikaramaswamy: done

PR #21, opened 2026-03-02
feat: improve validation id alignment with template-guided decisions

### LINKED ISSUE #8: feat: include entity value in validation decision output
### Priority Level

Medium (Nice to have)

Validation decisions currently only include the entity ID, decision, and proposed label — not the entity value. Adding the value alongside decisions makes the output easier to inspect and debug.

Rather than adding `value` to the `ValidationDecision` LLM schema (which risks hallucinated/echoed values), enrich decisions post-hoc via a `CustomColumnConfig` step that joins `value` from `COL_VALIDATION_CANDIDATES` by ID. Keeps the LLM schema lean. See PR #3 comment on `detection_workflow.py:53`.

_No response_

### Additional context

_No response_

lipikaramaswamy:
## Summary
Issue #8 asked for validation decision traces to include entity values. While implementing that, I found a deeper reliability issue: validator IDs were not consistently aligning with candidate IDs, so some validation decisions could be silently ignored.

To make traces both readable and trustworthy, this PR goes beyond value enrichment and hardens the full validation path.

### What changed
- Added post-validation enrichment so `_validated_entities.decisions[*]` include `value` (Issue #8 objective).
- Introduced template-guided validation input via `_validation_skeleton` so the model fills predefined decision rows instead of inventing structure. I think this still aligns better with a `LLMStructuredColumn` vs. `LLMTextColumn`. 
- Added `_validation_decisions` as explicit intermediate output; `_validated_entities` remains canonical post-processed output.
- Switched entity IDs from opaque hash fragments to deterministic readable span IDs (`{label}_{start}_{end}`) to improve copy fidelity and debugging.
- Added a filter on validator decisions to limit to known candidate IDs before finalization.
- Updated tests and prompt assertions to lock in these behaviors.
- Added two minor prompt improvements (one in replacement map generation and one when validating entities) for context-fit quality.

### Why this is broader than #8
A value-only trace enhancement is not sufficient if decision IDs do not reliably map to candidates. Without ID alignment, decisions can be dropped or ignored, which reduces correctness.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [x] Breaking change
- [ ] Documentation update
- [x] Refactoring

## Testing
- [x] Tests pass locally

## Related Issues
Closes #8

### DIFF
--- a/src/anonymizer/engine/detection/custom_columns.py
+++ b/src/anonymizer/engine/detection/custom_columns.py
@@ -45,6 +45,15 @@ from anonymizer.engine.detection.postprocess import (
     group_entities_by_value,
     parse_raw_entities,
 )
+from anonymizer.engine.schemas import (
+    EntitiesSchema,
+    RawValidationDecisionsSchema,
+    ValidatedDecisionSchema,
+    ValidatedDecisionsSchema,
+    ValidationCandidatesSchema,
+    ValidationSkeletonDecisionSchema,
+    ValidationSkeletonSchema,
+)
 
 
 @custom_column_generator(
@@ -118,21 +127,13 @@ def prepare_validation_inputs(row: dict[str, Any]) -> dict[str, Any]:
 @custom_column_generator(required_columns=[COL_VALIDATION_CANDIDATES])
 def build_validation_skeleton(row: dict[str, Any]) -> dict[str, Any]:
     """Pre-populate the decisions template with candidate IDs so the LLM only fills in decision/reason."""
-    candidates = row.get(COL_VALIDATION_CANDIDATES, [])
-    row[COL_VALIDATION_SKELETON] = {
-        "decisions": [
-            {
-                "id": c["id"],
-                "value": c["value"],
-                "label": c["label"],
-                "decision": None,
-                "proposed_label": None,
-                "reason": None,
-            }
-            for c in candidates
-            if isinstance(c, dict)
+    candidates = ValidationCandidatesSchema.from_raw(row.get(COL_VALIDATION_CANDIDATES, []))
+    skeleton = ValidationSkeletonSchema(
+        decisions=[
+            ValidationSkeletonDecisionSchema(id=c.id, value=c.value, label=c.label) for c in candidates.candidates
         ]
-    }
+    )
+    row[COL_VALIDATION_SKELETON] = skeleton.model_dump(mode="json")
     return row
 
 
@@ -141,26 +142,26 @@ def build_validation_skeleton(row: dict[str, Any]) -> dict[str, Any]:
 )
 def enrich_validation_decisions(row: dict[str, Any]) -> dict[str, Any]:
     """Enrich validation decisions with entity value and filter to known candidate IDs only."""
-    validated = row.get(COL_VALIDATION_DECISIONS, {})
-    # Both columns come from upstream structured/custom DD steps in the same cell-by-cell
-    # workflow, so they are expected as native Python objects (dict/list), not JSON strings.
-    candidates = row.get(COL_VALIDATION_CANDIDATES, [])
+    raw_decisions = RawValidationDecisionsSchema.from_raw(row.get(COL_VALIDATION_DECISIONS, {}))
+    candidates = ValidationCandidatesSchema.from_raw(row.get(COL_VALIDATION_CANDIDATES, []))
 
-    if not isinstance(validated, dict):
-        return row
+    candidate_lookup = {c.id: c for c in candidates.candidates}
+    valid_ids = set(candidate_lookup)
 
-    value_by_id = {c["id"]: c["value"] for c in candidates if isinstance(c, dict) and "id" in c and "value" in c}
-    valid_ids = set(value_by_id)
+    enriched = [
+        ValidatedDecisionSchema(
+            id=d.id,
+            decision=d.decision,
+            proposed_label=d.proposed_label,
+            reason=d.reason,
+            value=candidate_lookup[d.id].value,
+            label=candidate_lookup[d.id].label,
+        )
+        for d in raw_decisions.decisions
+        if d.id in valid_ids
+    ]
 
-    decisions = validated.get("decisions", [])
-    filtered: list[dict[str, str]] = []
-    if isinstance(decisions, list):
-        for d in decisions:
-            if not isinstance(d, dict) or d.get("id") not in valid_ids:
-                continue
-            filtered.append({**d, "value": value_by_id.get(d["id"], "")})
-
-    row[COL_VALIDATED_ENTITIES] = {"decisions": filtered}
+    row[COL_VALIDATED_ENTITIES] = ValidatedDecisionsSchema(decisions=enriched).model_dump(mode="json")
     return row
 
 
@@ -184,17 +185,16 @@ def apply_validation_and_finalize(row: dict[str, Any]) -> dict[str, Any]:
 
 
 def _from_dicts(entities: list[dict[str, str | int | float]]) -> list[EntitySpan]:
-    result: list[EntitySpan] = []
-    for entity in entities:
-        result.append(
-            EntitySpan(
-                entity_id=str(entity.get("id", "")),
-                value=str(entity.get("value", "")),
-                label=str(entity.get("label", "")),
-                start_position=int(entity.get("start_position", 0)),
-                end_position=int(entity.get("end_position", 0)),
-                score=float(entity.get("score", 0.0)),
-                source=str(entity.get("source", "detector")),
-            )
+    parsed = EntitiesSchema.from_raw(entities)
+    return [
+        EntitySpan(
+            entity_id=e.id,
+            value=e.value,
+            label=e.label,
+            start_position=e.start_position,
+            end_position=e.end_position,
+            score=e.score,
+            source=e.source,
         )
-    return result
+        for e in parsed.entities
+    ]
--- a/src/anonymizer/engine/detection/custom_columns.py
+++ b/src/anonymizer/engine/detection/custom_columns.py
@@ -31,6 +31,8 @@ from anonymizer.engine.constants import (
     COL_TEXT,
     COL_VALIDATED_ENTITIES,
     COL_VALIDATION_CANDIDATES,
+    COL_VALIDATION_DECISIONS,
+    COL_VALIDATION_SKELETON,
 )
 from anonymizer.engine.detection.postprocess import (
     EntitySpan,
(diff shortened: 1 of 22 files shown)


asteier2026 (approved):

src/anonymizer/engine/detection/custom_columns.py
```diff
@@ -113,6 +115,55 @@ def prepare_validation_inputs(row: dict[str, Any]) -> dict[str, Any]:
...
+)
+def enrich_validation_decisions(row: dict[str, Any]) -> dict[str, Any]:
+    """Enrich validation decisions with entity value and filter to known candidate IDs only."""
+    validated = row.get(COL_VALIDATION_DECISIONS, {})
+    # Both columns come from upstream structured/custom DD steps in the same cell-by-cell
+    # workflow, so they are expected as native Python objects (dict/list), not JSON strings.
+    candidates = row.get(COL_VALIDATION_CANDIDATES, [])
+
+    if not isinstance(validated, dict):
+        return row
```
andreatnvidia: suggestion: when `validated` is not a dict (e.g. failed LLM call), this returns without setting `COL_VALIDATED_ENTITIES`. Downstream `apply_validation_and_finalize` falls back to `{}` which keeps all entities, which is the right behavior, but it's implicit. Could make it explicit:

```python
if not isinstance(validated, dict):
    row[COL_VALIDATED_ENTITIES] = {"decisions": []}
    return row
```

tests/engine/test_detection_custom_columns.py
```diff
@@ -63,3 +67,42 @@ def test_parse_produces_valid_tagged_text_and_notation() -> None:
...
+
+def test_enrich_validation_decisions_ignores_non_dict_validation_payload() -> None:
+    row = {
+        COL_VALIDATION_DECISIONS: "unexpected-string-payload",
+        COL_VALIDATION_CANDIDATES: [
+            {"id": "id1", "value": "Alice", "label": "first_name", "context_before": "", "context_after": ""}
+        ],
+    }
+    result = enrich_validation_decisions(row)
+    assert result == row
```
andreatnvidia: nit: `enrich_validation_decisions` has three tests but `build_validation_skeleton` doesn't have any. could be worth adding a happy-path test and one for candidates with missing keys.

src/anonymizer/engine/detection/detection_workflow.py
```diff
@@ -308,7 +320,11 @@ def run(
             final_failures = detected_result.failed_records
 
         if COL_DETECTED_ENTITIES in final_df.columns:
-            final_df[COL_FINAL_ENTITIES] = final_df[COL_DETECTED_ENTITIES]
+            final_df[COL_FINAL_ENTITIES] = final_df[COL_DETECTED_ENTITIES].apply(
```
andreatnvidia, line 255: This changes final_entities from a list to {"entities": [...]}. The replace code still expects a plain list, so this breaks detection -> replace handoff.

andreatnvidia (commented):
**Re: `ValidationDecision` Pydantic model (detection_workflow.py:59-68)**

The prompt example output includes `value` and `label` per decision, but `ValidationDecision` doesn't have those fields. Since DataDesigner injects the Pydantic schema into the prompt, the LLM gets contradictory signals: the example shows value/label in the output, but the schema says they don't exist. On top of that, DataDesigner's schema validation prunes extra properties, so even if the LLM outputs them they get silently stripped.

Adding `value` and `label` to `ValidationDecision` would fix both the prompt contradiction and the pruning. The enrichment step still works as a safety net for ID filtering and value canonicalization.

src/anonymizer/engine/detection/detection_workflow.py
```diff
@@ -308,7 +320,11 @@ def run(
...
+                lambda x: (
+                    {"entities": x.tolist() if hasattr(x, "tolist") else list(x)} if x is not None else {"entities": []}
```
andreatnvidia: This blows up on NaN/float values with "float is not iterable". Can we guard with pandas.isna() or a small normalizer so failed rows do not crash run()?

tests/engine/test_detection_workflow.py
```diff
@@ -130,7 +130,9 @@ def test_run_without_latent_detection_materializes_final_entities(
 
     assert adapter.run_workflow.call_count == 1
     assert COL_FINAL_ENTITIES in result.dataframe.columns
-    assert result.dataframe[COL_FINAL_ENTITIES].tolist() == result.dataframe[COL_DETECTED_ENTITIES].tolist()
+    detected = result.dataframe[COL_DETECTED_ENTITIES].tolist()
+    final = result.dataframe[COL_FINAL_ENTITIES].tolist()
+    assert final == [{"entities": row} for row in detected]
```
andreatnvidia: Can we add one smoke test that runs detection.run() output directly through ReplacementWorkflow(Redact)? That would catch this shape mismatch immediately.

andreatnvidia:
A few more comments related to the entities list issue:

- `replace_runner._filter_entities_by_label` assumes `entities_column` is a list of dicts. After the detection change it can be `{"entities": [...]}`, so filtering can run against the wrong shape (or end up empty).
- `strategies._build_local_replacement_map` has the same list-of-dicts assumption. If a dict wrapper is passed, local replace can silently no-op.
- `strategies._apply_replacement_map_to_text` has the same assumption. If entities are wrapped, replacements can be skipped; supporting both formats would keep backward compatibility.

andreatnvidia (changes requested):
The template-guided validation approach is a nice improvement, and the ID simplification + enrichment pipeline make sense.

The main blocker is the `COL_FINAL_ENTITIES` shape change from list to `{"entities": [...]}` (detection_workflow.py:323). This breaks the detection -> replace handoff since the replace pipeline expects a plain list of dicts. See inline comments for specifics on where it breaks.

Also:
- `ValidationDecision` Pydantic model is missing `value`/`label` fields, which contradicts the prompt example and gets silently pruned by DataDesigner's schema validation
- `enrich_validation_decisions` early return doesn't set `COL_VALIDATED_ENTITIES`, which could fail DataDesigner's custom column contract

Would be great to get an integration smoke test covering detection output -> replacement as well.

src/anonymizer/engine/detection/custom_columns.py
lipikaramaswamy: Thanks — agreed. I made this explicit. `enrich_validation_decisions` now always materializes `COL_VALIDATED_ENTITIES` through schema normalization, so invalid/non-dict validator payloads produce `{"decisions": []}` instead of relying on implicit downstream behavior.

tests/engine/test_detection_custom_columns.py
lipikaramaswamy: Added tests for `build_validation_skeleton` - happy path and missing keys

src/anonymizer/engine/detection/detection_workflow.py
lipikaramaswamy: Addressed with schema normalization in `run()`: NaN, float and non-iterable values now fail safe to an empty wrapper instead of raising e.g. `"float is not iterable"`.

tests/engine/test_detection_workflow.py
lipikaramaswamy: Added handoff smoke coverage in replace tests.

src/anonymizer/engine/detection/detection_workflow.py
lipikaramaswamy: Fixed by standardizing on a single canonical shape. Entity payloads are now wrapper-only (`{"entities": [...]}`) across detection and replace. I also removed bare-list backward compatibility so this contract is explicit and consistent end-to-end

lipikaramaswamy:
@andreatgretel thanks for the feedback — it helped drive a broader hardening pass: formalized workflow column contracts with Pydantic schema wrappers, tightened typing, and removed shape ambiguity in detection→replace handoff.

Specifically:

- Made validation fallback explicit: invalid/non-dict validation outputs now materialize `{"decisions": []}` via schema normalization instead of relying on implicit downstream behavior.
- Added `build_validation_skeleton` coverage for both happy-path and missing-key candidate cases.
- Fixed detection -> replace handoff shape issues by standardizing entity payload contracts and removing list-of-dicts assumptions in replace paths.
- Added handoff regression tests for detection-shaped output through `ReplacementWorkflow(Redact)`, including dict-wrapped numpy-array payloads.
- Fixed NaN/float/non-iterable robustness in entity materialization via schema-based normalization (fail-safe to empty wrapper instead of crashing).
- Tightened type contracts end-to-end:
  - canonical wrappers for workflow columns (`{"entities": [...]}`, `{"candidates": [...]}`, `{"entities_by_value": [...]}`)
  - intentionally dropped bare-list backward compatibility in favor of a single explicit contract.
- Added schema regression tests for malformed payloads and non-canonical bare-list inputs.

src/anonymizer/engine/schemas.py
```diff
@@ -0,0 +1,182 @@
...
+) -> T:
+    """Parse raw DataFrame cell value into a wrapper schema."""
+
+    def _safe_validate(candidate_list: list[object]) -> T:
+        try:
+            return model_cls.model_validate({key: candidate_list})
+        except ValidationError:
+            return model_cls()
+
+    if isinstance(raw, dict):
```
andreatnvidia: `from_raw` currently only handles dict wrappers. If this cell contains an `EntitiesSchema` object, we silently fall back to empty and lose entities. Can we also accept model instances (`model_dump`) or fail explicitly instead of silently dropping data?

src/anonymizer/engine/replace/strategies.py
```diff
@@ -60,38 +59,35 @@ def apply_replacement_map(
     output_df[COL_REPLACED_TEXT] = output_df.apply(
         lambda row: _apply_replacement_map_to_text(
             text=str(row.get(text_column, "")),
-            entities=row.get(entities_column, []),
-            replacement_map=row.get(replacement_map_column, {"replacements": []}),
+            entities=EntitiesSchema.from_raw(row.get(entities_column, {})),
```
andreatnvidia, line 62: `apply_replacement_map` depends on `EntitiesSchema.from_raw`, so this can silently no-op when `final_entities` is an `EntitiesSchema` object. I hit this in smoke tests (`Alice` stayed `Alice` instead of `Maya`). Can we harden parsing so valid object payloads still replace?

src/anonymizer/interface/display.py
```diff
@@ -68,7 +68,7 @@ def render_record_html(row: pd.Series, record_index: int | None = None, original
     text_col = original_text_column or "text"
     text = str(row.get(text_col, ""))
     replaced_text = str(row.get(f"{text_col}_replaced", ""))
-    entities = _normalize_entity_list(row.get(COL_DETECTED_ENTITIES, []))
+    entities = EntitiesSchema.from_raw(row.get(COL_DETECTED_ENTITIES, {})).entities
```
andreatnvidia, line 71: `render_record_html` has the same behavior: highlights disappear for object payloads even when entity data exists. Smoke test with `EntitiesSchema` in `_detected_entities` rendered no `first_name` highlight. Can we normalize object payloads here too?

tests/engine/test_replace_runner.py
```diff
@@ -125,7 +127,33 @@ def test_apply_replacement_map_handles_numpy_array_entities() -> None:
...
+                    ]
+                }
+            ],
+        }
+    )
+    output_df = apply_replacement_map(dataframe)
+    assert output_df[COL_REPLACED_TEXT].iloc[0] == "Maya works at NovaCorp"
+
+
+def test_apply_replacement_map_handles_dict_wrapped_entities() -> None:
```
andreatnvidia: Can we add one param test that runs replace/display with entity payloads as: dict wrapper, numpy-wrapped dict wrapper, and `EntitiesSchema` object? That would catch this silent-regression class in one place.

andreatnvidia (approved):
LGTM! Schemas look solid, tests pass, latest changes are clean.

One small thing: `ValidationDecision` in `detection_workflow.py` and `RawValidationDecisionSchema` in `schemas.py` are separate but need to stay in sync — might be worth a quick comment linking them. Not blocking on it.

PR #22, opened 2026-03-04
feat: add user-facing pipeline logging with progress signals

### LINKED ISSUE #14: feat: add user-facing logging with pipeline stage progress and entity stats

andreatnvidia:
## Summary

- Add `logging.getLogger("anonymizer")` with INFO-level messages at each stage of `_run_internal`: input loading, detection, replacement, and pipeline completion
- Emoji prefixes and `LOG_INDENT` (`  |-- `) for visual hierarchy
- `LoggingConfig` presets (`default`, `verbose`, `debug`) with `configure_logging()` entry point
- `configure_logging(enabled=False)` opt-out for callers who manage their own logging
- Auto-configure on first `Anonymizer()` construction (named logger only, `propagate=False`, no root logger changes)
- Suppress `data_designer` and other noisy loggers to WARNING by default
- `ProgressTracker` for large datasets (log-based progress at 10% intervals)
- DEBUG-level messages for detection, replacement, and NDD adapter internals
- Preview mode logs now use the effective record count, not the full dataset size

Closes #14

### Diff breakdown (12 files, +718/-16)

| | Files | Added | Removed |
|---|---|---|---|
| Source (`src/`) | 7 | +333 | -16 |
| Tests (`tests/`) | 5 | +385 | -0 |
| **Total** | **12** | **+718** | **-16** |

25 new logging-specific tests, 244 total tests passing.

## Test plan

- [x] All 244 tests pass (including 25 new logging tests)
- [x] Lint and format checks pass (pre-commit hooks installed)
- [x] Verified notebook output reflects preview count fix
- [x] Removed `logging_demo.ipynb` and `logging_demo.py` before merge


src/anonymizer/logging.py
```diff
@@ -0,0 +1,40 @@
...
+import logging
+import sys
+
+LOG_INDENT = "  |-- "
+
+_DEFAULT_NOISY_LOGGERS = ["httpx", "httpcore", "mcp"]
+
+
+def configure_logging(*, verbose: bool = False) -> None:
+    """Set up logging for Anonymizer.
```
andreatnvidia: Current implementation is intentionally minimal:

```python
configure_logging()                # anonymizer INFO, DD suppressed
configure_logging(verbose=True)    # anonymizer INFO, DD DEBUG
```

As a next step, we could follow DataDesigner's `LoggingConfig` pattern with preset factories and per-logger overrides:

```python
from anonymizer import configure_logging, LoggingConfig

configure_logging(LoggingConfig.default())    # anonymizer INFO, DD suppressed
configure_logging(LoggingConfig.verbose())    # anonymizer INFO, DD INFO
configure_logging(LoggingConfig.debug())      # everything DEBUG

# Fine-grained: start from a preset, override specific loggers
config = LoggingConfig.default()
config.set_level("data_designer", "DEBUG")
configure_logging(config)
```

Not needed for this PR, but worth considering if users ask for more control.

andreatnvidia:

logging_demo.py
andreatnvidia, line 1: Both this and the .ipynb version will be removed before merging

andreatnvidia:
**Follow-up: structured logging from Data Designer**

The debug mode added here logs input data stats, detection config, and NDD adapter internals. For DD-side signals (model usage summary, per-record generation failures), we currently rely on setting DD's logger to INFO and letting its human-readable messages propagate through — which works but gives us no programmatic access to the data.

I've filed NVIDIA-NeMo/DataDesigner#386 requesting that DD attach structured metadata to its log records via `extra={"dd_event": {...}}`. If adopted, we could:

- Collect per-LLM-call telemetry (tokens, latency, retries) live during pipeline runs
- Surface actual failure reasons instead of inferring them from missing record IDs
- Aggregate model usage stats without parsing log strings
- Write events to JSONL for offline cost analysis and profiling

This is a non-breaking change on DD's side (existing log output stays identical), and on our side we'd add a lightweight `logging.Handler` around NDD adapter calls to collect the events. No new DD dependencies required.

Alternatively, we could parse DD's log messages directly with regex to extract this information, but that's brittle — it couples us to DD's exact log format strings, which can change without notice between releases.

src/anonymizer/interface/anonymizer.py
```diff
@@ -113,7 +134,28 @@ def _run_internal(
...
+            logger.debug(
+                "detection config: threshold=%.2f, labels=%s",
+                config.detect.gliner_threshold,
+                config.detect.entity_labels or "(default)",
+            )
 
+        if preview_num_records is not None:
+            effective_preview = min(preview_num_records, num_records)
+            logger.info(LOG_INDENT + "👀 Preview mode: processing %d of %d records", effective_preview, num_records)
+        logger.info("🔍 Running entity detection on %d records", num_records)
```
lipikaramaswamy: In preview mode, I think the execution-stage logs should consistently use the number of records actually being processed rather than the full input size. Right now we log the preview subset in the banner, but then switch back to the full dataset count immediately after, which makes the log stream contradictory. I'd expect preview runs of 2 records on a 5 record dataset to read more like: `Loaded 5 records`, `Preview mode: processing 2 records`, `Running entity detection on 2 records`, and `Pipeline complete — 2 records processed`.

- Keep: `Loaded 5 records from ...`
- Change: `Preview mode: processing 2 of 5 records` -> `Preview mode: processing 2 records`
- Change: `Running entity detection on 5 records` -> `Running entity detection on 2 records`
- Keep: `Detection complete — 10 entities found across 2 records (0 failed) [6.8s]`
- Change: `Pipeline complete — 5 records processed, 0 total failures` -> `Pipeline complete — 2 records processed, 0 total failures`

src/anonymizer/interface/anonymizer.py
```diff
@@ -54,10 +69,16 @@ def __init__(
             detection_workflow: Custom detection workflow (advanced/testing).
             replace_runner: Custom replacement workflow (advanced/testing).
         """
+        _initialize_logging()
```
lipikaramaswamy, line 74: This is a bit of a head-scratcher to me. Logs currently work nicely out of the box, but having `Anonymizer()` auto-configure logging on construction feels pretty invasive for a core library object, since it can change logging behavior for callers that already have application logging set up.

What’s the intended contract here? Should `Anonymizer()` itself be taking ownership of logging setup, or would it make more sense for the notebook/convenience path to opt into that behavior explicitly?

Part of why this stood out is that `DataDesigner` seems to handle this through `data_designer.essentials` rather than in the core `DataDesigner()` constructor itself. I doubt following blindly is the right move here either though

src/anonymizer/interface/anonymizer.py
```diff
@@ -113,7 +134,28 @@ def _run_internal(
         preview_num_records: int | None,
     ) -> AnonymizerResult:
         input_df = read_input(data)
```
lipikaramaswamy, line 144: This PR got me thinking that I’d like us to be more opinionated about the `preview()` contract. Right now `preview()` still loads the full dataset and only limits downstream execution via `preview_num_records`. For large inputs that feels too heavy, especially since loading happens each time `preview()` and `run()` are called.

I think preview should probably be lightweight end to end - logged in #28

lipikaramaswamy (changes requested):

src/anonymizer/interface/anonymizer.py
andreatnvidia: good catch — fixed. detection and pipeline-complete logs now use the effective preview count instead of the full dataset size. kept "processing N of M" in the preview banner since it's useful context to see both numbers.

src/anonymizer/interface/anonymizer.py
andreatnvidia: good question. `essentials` doesn't exist anymore in DD — logging is now called at module import time in `interface/__init__.py`, clearing root logger handlers. our approach is less invasive: we only touch the named `"anonymizer"` logger with `propagate=False`, and defer setup to first `Anonymizer()` construction rather than import time.

that said there was no way to fully opt out. added `configure_logging(enabled=False)` — call it before constructing `Anonymizer()` to suppress all handler setup.

src/anonymizer/interface/anonymizer.py
andreatnvidia: agreed. considered slicing early in `_run_internal` but `preview_num_records` is threaded through 6+ files down to the adapter layer where the actual slicing happens. pulling that up would touch detection workflows, replace workflows, and the NDD adapter. better as a focused refactor than a side-change in this PR.

lipikaramaswamy (approved):

src/anonymizer/interface/anonymizer.py
lipikaramaswamy: Got it, thanks for adding `configure_logging(enabled=False)` to give callers a clean opt-out when they already have logging configured

PR #27, opened 2026-03-07
feat: add preflight config validation for active workflows

### LINKED ISSUE #5: feat: add config validation step for model alias references
### Priority Level

Medium

### Task Summary

Add an `Anonymizer.validate_config()` method (similar to DD's `DataDesigner.validate()`) that checks model alias references, required fields, and config consistency before running the pipeline. Currently alias validation only happens at runtime inside `load_workflow_config()`.

New public method on `Anonymizer` that validates:
- `selected_models` aliases exist in the model pool
- Required model aliases are present for the chosen replace strategy (e.g. LLMReplace needs a replacement_generator)
- Config fields are consistent

See TODO in `model_loader.py:156`.

### Dependencies

_No response_

lipikaramaswamy:
## Summary
This PR adds a public `Anonymizer.validate_config()` method to preflight-check active workflow configuration against the parsed model pool. We reuse the same preflight path in `run()` and `preview()` so invalid model selections fail before workflow execution.

We validate only the aliases required for the active config:
- detection aliases are always validated, as they are required for all workflows
- substitute alias is checked only when `Substitute` is selected in replace mode. 
- rewrite aliases are checked only in rewrite mode

We also add `InvalidConfigError` so public config validation failures surface consistently.

## Test Plan
- `uv run pytest tests/interface/test_anonymizer_interface.py tests/engine/test_model_loader.py`
- Optional full suite after unrelated payload work is resolved: `uv run --group dev pytest`
## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing
- [x] Extended model-loader and interface tests, and moved shared validation fixtures into `tests/conftest.py`; all tests pass locally
- [x] Notebooks run end-to-end

## Related Issues
Closes #5

### DIFF
--- a/src/anonymizer/engine/ndd/model_loader.py
+++ b/src/anonymizer/engine/ndd/model_loader.py
@@ -149,16 +149,42 @@ def _merge_selections(user_selections: dict[str, dict[str, str]] | None) -> Mode
     )
 
 
+def validate_model_alias_references(
+    model_configs: list[ModelConfig],
+    selected_models: ModelSelection,
+    *,
+    check_substitute: bool = False,
+    check_rewrite: bool = False,
+) -> None:
+    """Validate that active workflow model aliases exist in the model pool."""
+    known_aliases = {model_config.alias for model_config in model_configs}
+
+    roles_to_check: dict[str, str] = {
+        f"detection.{role}": alias for role, alias in selected_models.detection.model_dump().items()
+    }
+    if check_substitute:
+        roles_to_check.update(
+            {f"replace.{role}": alias for role, alias in selected_models.replace.model_dump().items()}
+        )
+    if check_rewrite:
+        roles_to_check.update(
+            {f"rewrite.{role}": alias for role, alias in selected_models.rewrite.model_dump().items()}
+        )
+
+    unknown = {path: alias for path, alias in roles_to_check.items() if alias not in known_aliases}
+    if unknown:
+        unknown_str = ", ".join(f"{path}={alias!r}" for path, alias in sorted(unknown.items()))
+        raise ValueError(
+            f"Selected model aliases not found in model pool: {unknown_str}. Known aliases: {sorted(known_aliases)}"
+        )
+
+
 def _validate_alias_references(
     models_config: dict[str, Any],
     selections: dict[str, str],
     workflow_name: str,
 ) -> None:
-    """Validate that workflow role selections reference aliases defined in models.yaml.
-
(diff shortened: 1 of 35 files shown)


src/anonymizer/interface/anonymizer.py
```diff
@@ -105,6 +109,10 @@ def preview(
             preview_num_records=num_records,
         )
 
+    def validate_config(self, config: AnonymizerConfig) -> None:
```
andreatnvidia, line 112: suggestion: `validate_config()` now raises `InvalidConfigError`, but that exception is not exported from the top-level `anonymizer` package. users can still import it from `anonymizer.interface.errors`, so nothing is broken, but if `validate_config()` is meant to be part of the public API it might be nicer to re-export `InvalidConfigError` alongside the other top-level symbols.

src/anonymizer/engine/ndd/model_loader.py
```diff
@@ -149,16 +149,42 @@ def _merge_selections(user_selections: dict[str, dict[str, str]] | None) -> Mode
...
+    selected_models: ModelSelection,
+    *,
+    check_substitute: bool = False,
+    check_rewrite: bool = False,
+) -> None:
+    """Validate that active workflow model aliases exist in the model pool."""
+    known_aliases = {model_config.alias for model_config in model_configs}
+
+    roles_to_check: dict[str, str] = {
+        f"detection.{role}": alias for role, alias in selected_models.detection.model_dump().items()
```
andreatnvidia: I think this is validating one detection alias a bit too eagerly. `latent_detector` only gets resolved when rewrite mode turns on latent detection, but `roles_to_check` currently includes it for every config. That means a replace-only custom model pool can fail `validate_config()` or `run()` just because it omits the latent model, even though that path would never use it. maybe start with the always-used detection roles and add `latent_detector` only when `check_rewrite=True`?

andreatnvidia (approved):
Looks good overall. I left two small nits around config validation, but nothing blocking from my side.

src/anonymizer/interface/anonymizer.py
lipikaramaswamy: Ok, exported all user-facing errors from the top-level

src/anonymizer/engine/ndd/model_loader.py
lipikaramaswamy: Done, thanks!

Issue #35, opened 2026-03-09
feat(rewrite): engine — final judge + RewriteWorkflow


lipikaramaswamy:
Part of #30.

## Scope

**`engine/rewrite/final_judge.py`** — `FinalJudgeWorkflow`:
- One NddAdapter call with `LLMJudgeColumnConfig` → `COL_JUDGE_EVALUATION` (`judge` alias)
- Three rubrics on 1-10 scale (to be revised in follow-up): privacy, quality, naturalness
- Prompt contextualized by `privacy_goal`
- **Non-critical**: failure logged as warning, not raised

**`engine/rewrite/rewrite_workflow.py`** — `RewriteWorkflow`:

Top-level orchestrator. Chains all 6 sub-workflows in order:
1. `DomainClassificationWorkflow`
2. `SensitivityDispositionWorkflow`
3. `QaGenerationWorkflow`
4. `RewriteGenerationWorkflow`
5. `EvaluateRepairWorkflow`
6. `FinalJudgeWorkflow` (non-critical — swallow failure, log warning)

Fast path: if no record has entities, populate passthrough columns and return without any LLM calls.

Accumulates `failed_records` from all steps. Propagates `df.attrs` throughout.

Returns `RewriteResult(dataframe, failed_records)`.

**Tests**: `tests/engine/test_final_judge.py`, `tests/engine/test_rewrite_workflow.py`

## Acceptance criteria
- [ ] `FinalJudgeWorkflow` uses `judge` alias
- [ ] `FinalJudgeWorkflow` failure does not propagate as exception from `RewriteWorkflow`
- [ ] `RewriteWorkflow` calls all 6 sub-workflows in correct order (verified via mocks)
- [ ] Fast path: no sub-workflows called when input has no entities
- [ ] All failed_records accumulated across all steps
- [ ] `df.attrs` propagated to final output

Issue #31, opened 2026-03-09
feat(rewrite): engine — domain classification + sensitivity disposition


lipikaramaswamy:
Part of #30.

## Scope

**`engine/rewrite/domain_classification.py`** — `DomainClassificationWorkflow`:
- Column factory: `columns(*, selected_models, data_summary=None) -> list[ColumnConfigT]`
- Returns `[LLMStructuredColumnConfig(COL_DOMAIN, DomainClassificationSchema, domain_classifier alias), CustomColumnConfig(COL_DOMAIN_SUPPLEMENT, _enrich_domain)]`
- Prompt: `_get_domain_classification_prompt(data_summary)`

**`engine/rewrite/sensitivity_disposition.py`** — `SensitivityDispositionWorkflow`:
- Column factory: `columns(*, selected_models, privacy_goal, data_summary=None) -> list[ColumnConfigT]`
- Returns `[LLMStructuredColumnConfig(COL_SENSITIVITY_DISPOSITION, SensitivityDispositionSchema, disposition_analyzer alias)]`
- Consumes: `COL_TAGGED_TEXT`, `COL_FINAL_ENTITIES`, `COL_LATENT_ENTITIES`, `COL_DOMAIN`, `COL_DOMAIN_SUPPLEMENT`
- Prompt: `_get_sensitivity_disposition_prompt(privacy_goal, data_summary)`
- Note: missing `COL_LATENT_ENTITIES` default injection is `RewriteWorkflow`'s responsibility (#35)

**`engine/rewrite/__init__.py`**

**Tests**: `tests/engine/test_domain_classification.py`, `tests/engine/test_sensitivity_disposition.py`

## Acceptance criteria
- [ ] `DomainClassificationWorkflow.columns()` returns `LLMStructuredColumnConfig` with `domain_classifier` alias and `CustomColumnConfig` for `COL_DOMAIN_SUPPLEMENT`
- [ ] `COL_DOMAIN_SUPPLEMENT` populated from lookup without LLM call
- [ ] `SensitivityDispositionWorkflow.columns()` returns `LLMStructuredColumnConfig` with `disposition_analyzer` alias
- [ ] `privacy_goal` correctly interpolated into disposition prompt
- [ ] Prompt helper functions reference required columns

Issue #30, opened 2026-03-09
feat: add rewrite mode


lipikaramaswamy:
Implements full rewrite mode in Anonymizer — a semantic anonymization pipeline that rewrites text to remove PII while preserving meaning, rather than simply replacing or redacting entities.

## Pipeline overview

Detection result (entities, tagged text, latent entities)
→ Domain classification
→ Sensitivity disposition (per-entity protection plan)
→ QA generation (meaning units + quality QA + privacy QA)
→ Rewrite generation (guided by disposition + replacement map)
→ Evaluate & repair loop (leakage mass, utility score, targeted repair)
→ Final judge (holistic privacy/quality/naturalness scores)

## Architecture

All steps run in a **single `run_workflow` call**. Each sub-workflow (`DomainClassificationWorkflow`, `SensitivityDispositionWorkflow`, etc.) is a column factory — it exposes a `columns()` method returning `list[ColumnConfigT]`. `RewriteWorkflow` concatenates all columns and calls the adapter once. Because columns are processed sequentially, each step sees the output of prior steps in the same call.

Repair uses the `fn(row, generator_params, models)` custom column signature so the loop runs inside a single column — no additional adapter calls.

Rows with no detected entities are split out before the call and re-merged after with default values (fast path: no LLM calls).

```
Detection output (COL_TAGGED_TEXT, COL_FINAL_ENTITIES, COL_LATENT_ENTITIES, ...)
        │
        ▼
[Pre-processing — pure Python]
  • Inject COL_LATENT_ENTITIES default if missing
  • Split: passthrough rows (no entities) ────────────────────────────────────┐
        │                                                                     │
        ▼                                                                     │
┌─────────────────────────────────────────────────────────────────────────┐   │
│  run_workflow  ("rewrite-pipeline")                                     │   │
│                                                                         │   │
│  DomainClassificationWorkflow.columns()                                 │   │
│    LLM    ──► COL_DOMAIN                                                │   │
│    Custom ──► COL_DOMAIN_SUPPLEMENT                                     │   │
│                                                                         │   │
│  SensitivityDispositionWorkflow.columns()                               │   │
│    LLM    ──► COL_SENSITIVITY_DISPOSITION                               │   │
│                                                                         │   │
│  QaGenerationWorkflow.columns()                                         │   │
│    Custom ──► COL_SENSITIVITY_DISPOSITION_BLOCK                         │   │
[...]

PR #40, opened 2026-03-10
feat: add rewrite foundation - schemas, constants, and model selection

### LINKED ISSUE #29: feat(rewrite): foundation — schemas, constants, model selection
## Summary

Establishes all shared vocabulary for the rewrite pipeline. No behavior change — pure additions that all subsequent engine PRs depend on.

## Scope

**Phase 1 — Schemas** (`engine/schemas/rewrite.py`):
- `Domain`, `DomainClassificationSchema`
- `EntitySource`, `EntityCategory`, `SensitivityLevel`, `ProtectionMethod`, `CombinedRiskLevel`
- `EntityDispositionSchema`, `SensitivityDispositionSchema` (with validators + helper methods)
- `MeaningUnitsSchema`, `QualityQAPairsSchema`, `PrivacyQAPairsSchema`
- `QualityAnswersSchema`, `PrivacyAnswersSchema`
- `RewriteOutputSchema`
- Export all from `engine/schemas/__init__.py`

**Phase 2 — Constants** (`engine/constants.py`):
- Rewrite pipeline column names: `COL_DOMAIN`, `COL_SENSITIVITY_DISPOSITION`, `COL_MEANING_UNITS`, `COL_QUALITY_QA`, `COL_PRIVACY_QA`, `COL_REWRITTEN_TEXT`, `COL_UTILITY_SCORE`, `COL_LEAKAGE_MASS`, `COL_ANY_HIGH_LEAKED`, `COL_NEEDS_HUMAN_REVIEW`, `COL_JUDGE_EVALUATION`, and others

**Phase 3 — Model selection expansion** (`config/models.py`, `default_model_configs/rewrite.yaml`):
- Expand `RewriteModelSelection` from 2 roles to 7: `domain_classifier`, `disposition_analyzer`, `meaning_extractor`, `qa_generator`, `rewriter`, `evaluator`, `repairer`, `judge`
- Update `rewrite.yaml` with role-to-alias mappings
- Update `test_model_loader.py` for new fields

## Acceptance criteria
- [ ] All schemas importable from `anonymizer.engine.schemas`
- [ ] All new column constants defined and named consistently
- [ ] `RewriteModelSelection` has all 7 roles
- [ ] `rewrite.yaml` aliases all resolve against `models.yaml`
- [ ] `test_model_loader.py` passes with new role assertions
- [ ] All existing tests still pass

lipikaramaswamy:
## Summary
  - add `engine/schemas/rewrite.py` with Pydantic output schemas for all 6 rewrite pipeline steps, and export from `anonymizer.engine.schemas`
  - define rewrite pipeline column constants in `engine/constants.py` for use by downstream engine PRs (#31–#35)
  - expand RewriteModelSelection from 2 to 8 roles — split evaluator (QA re-answering) from repairer (repair generation) since they have different model profiles; update `rewrite.yaml` defaults accordingly

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing
- [x] Verified tests pass locally - `uv run pytest tests/engine/test_schemas.py tests/engine/test_model_loader.py`
- [x] Verified `rewrite.yaml` aliases resolve against `models.yaml`
- [x] verified new rewrite schemas are importable from `anonymizer.engine.schemas`

## Related Issues
Closes #29


asteier2026 (approved):

src/anonymizer/config/default_model_configs/rewrite.yaml
```diff
@@ -1,3 +1,9 @@
 selected_models:
+  domain_classifier: gpt-oss-120b
+  disposition_analyzer: gpt-oss-120b
+  meaning_extractor: gpt-oss-120b
+  qa_generator: gpt-oss-120b
   rewriter: gpt-oss-120b
   evaluator: nemotron-30b-thinking
+  repairer: gpt-oss-120b
+  judge: nemotron-30b-thinking
```
lipikaramaswamy, line 9: The actual default models being used are placeholders.

src/anonymizer/engine/schemas/rewrite.py
```diff
@@ -0,0 +1,397 @@
...
+def generate_privacy_qa_from_disposition(
+    sensitivity_disposition: SensitivityDispositionSchema | dict[str, Any],
+) -> PrivacyQAPairsSchema:
+    """Generate privacy QA questions from sensitivity disposition without an LLM call.
+
+    Produces one question per entity that needs protection:
+        "Can the {entity_label} '{entity_value}' be deduced from the rewritten text?"
+    All questions expect the answer "no" — a "yes" answer indicates a privacy leak.
+    """
+    if isinstance(sensitivity_disposition, dict):
```
andreatnvidia, line 377: this helper says it accepts raw dicts, but the dict branch skips schema validation and can crash with a bare `KeyError` on malformed payloads. perhaps normalize dict input through `SensitivityDispositionSchema.model_validate(...)` first so callers get a proper validation error instead?

src/anonymizer/engine/schemas/rewrite.py
```diff
@@ -0,0 +1,397 @@
...
+
+    rewritten_text: str
+
+
+# ---------------------------------------------------------------------------
+# Evaluation
+# ---------------------------------------------------------------------------
+
+
+class QualityAnswerSchema(BaseModel):
```
andreatnvidia, line 308: I think the `id` type flips here in a way that will make the next step brittle. Step 3 emits question ids as numbers, but these answer schemas require strings, and Pydantic 2 rejects `{"id": 1}` instead of coercing it. maybe keep the same `int` id type across the question, answer, and compare schemas?

src/anonymizer/engine/schemas/rewrite.py
```diff
@@ -0,0 +1,397 @@
...
+    domain: Domain
+    domain_confidence: float = Field(ge=0.0, le=1.0)
+
+
+# ---------------------------------------------------------------------------
+# Sensitivity Disposition
+# ---------------------------------------------------------------------------
+
+
+class EntitySource(str, Enum):
```
kendrickb-nvidia, line 94: nit: Use StrEnum if we're targeting python 3.11+?

src/anonymizer/engine/schemas/rewrite.py
```diff
@@ -0,0 +1,397 @@
...
+
+
+class CombinedRiskLevel(str, Enum):
+    low = "low"
+    medium = "medium"
+    high = "high"
+
+
+class EntityDispositionSchema(BaseModel):
+    """Protection decision for a single entity."""
```
kendrickb-nvidia: question: Is this for a specific instance of any entity, ie characters 25-40 in the text. Or a general "entity" in the text, that might appear multiple times? "Single entity" is ambiguous to my reading, maybe clarify in the docstring.

src/anonymizer/engine/schemas/rewrite.py
```diff
@@ -0,0 +1,397 @@
...
+
+    model_config = ConfigDict(use_enum_values=True)
+
+    id: int = Field(ge=1)
+    source: EntitySource
+    category: EntityCategory
+    sensitivity: SensitivityLevel
+    entity_label: str = Field(min_length=1)
+    entity_value: str = Field(min_length=1)
+    does_need_protection: bool
```
kendrickb-nvidia: nit: `does_need_protection` is a mouthful (and tokenful). Unless `does_` is a pattern we want to use, consider `needs_protection`, or another name perhaps? `protection_needed` parallels the other protection fields for example.

There'a also validation between `does_need_protection, protection_reason, protection_method_suggestion` that they need to be set in specific ways. Better for typing if we can encode that in the pydantic model. But I realize that may not work as nicely with the LLMs.

src/anonymizer/engine/schemas/rewrite.py
```diff
@@ -0,0 +1,397 @@
...
+    def _validate_protection_consistency(self) -> EntityDispositionSchema:
+        if not self.does_need_protection and self.protection_method_suggestion != ProtectionMethod.left_as_is:
+            raise ValueError(
+                f"Entity {self.id}: does_need_protection=False requires protection_method_suggestion='left_as_is', "
+                f"got '{self.protection_method_suggestion}'"
+            )
+        if self.does_need_protection and self.protection_method_suggestion == ProtectionMethod.left_as_is:
+            raise ValueError(
+                f"Entity {self.id}: does_need_protection=True cannot have protection_method_suggestion='left_as_is'"
+            )
```
kendrickb-nvidia, line 156: question: This just encodes that `does_need_protection = protection_method_suggestion == ProtectionMethod.left_as_is`, right? Do we need the extra field at all? Or can we make `needs_protection()` a method (or property `needs_protection`) that is programmatically determined from `protection_method_suggestion`.

Seems like we're making extra work for the LLM to jump through to make a well formed output that conforms to this schema, does that have value?

src/anonymizer/engine/schemas/rewrite.py
```diff
@@ -0,0 +1,397 @@
...
+
+    @model_validator(mode="after")
+    def _validate_ids_sequential(self) -> SensitivityDispositionSchema:
+        ids = [e.id for e in self.sensitivity_disposition]
+        expected = list(range(1, len(ids) + 1))
+        if sorted(ids) != expected:
+            raise ValueError(f"Entity IDs must be sequential starting from 1. Got: {sorted(ids)}, expected: {expected}")
+        return self
+
+    def entities_needing_protection(self) -> list[EntityDispositionSchema]:
```
kendrickb-nvidia: super nit: Start with a verb for a method? Or make a property if you want the noun? Feel free to ignore :)

src/anonymizer/engine/schemas/rewrite.py
```diff
@@ -0,0 +1,397 @@
...
+        lines = []
+        for e in entities:
+            lines.append(
+                f'- [{e.sensitivity.upper()}] {e.entity_label}: "{e.entity_value}" → {e.protection_method_suggestion} (Reason: {e.protection_reason})'
+            )
+        return "\n".join(lines)
+
+
+# ---------------------------------------------------------------------------
+# Meaning Units
```
kendrickb-nvidia, line 214: question: Is "meaning units" standard terminology in privacy field that I should know?

Overall struggling a bit at seeing some of these names out of context. This probably all works if we have an architecture diagram or docs that talk through the rewrite process.

src/anonymizer/engine/constants.py
```diff
@@ -47,6 +47,28 @@
...
+COL_QUALITY_QA_REANSWER = "_quality_qa_reanswer"
+COL_QUALITY_QA_COMPARE = "_quality_qa_compare"
+COL_PRIVACY_QA_REANSWER = "_privacy_qa_reanswer"
+COL_JUDGE_EVALUATION = "_judge_evaluation"
+
+# Rewrite pipeline — user-facing output columns
+COL_UTILITY_SCORE = "utility_score"
+COL_LEAKAGE_MASS = "leakage_mass"
+COL_ANY_HIGH_LEAKED = "any_high_leaked"
+COL_NEEDS_HUMAN_REVIEW = "needs_human_review"
```
kendrickb-nvidia, line 70: nit: From these names I can't tell the types (and maybe interpretation) for several of these columns, not sure that's useful to convey or document here, but unless enforced via type annotations I can see this being easy to mess up. E.g., is `COL_NEEDS_HUMAN_REVIEW` a bool, an int with 1 (needs review) or 0 (no review), a string with "yes" or "no", or a string containing the snippet requiring review?

tests/engine/test_model_loader.py
```diff
@@ -21,12 +21,12 @@ def test_load_workflow_config_contains_selected_models() -> None:
     config = load_workflow_config(WorkflowName.detection)
     assert "model_configs" in config
     assert "selected_models" in config
-    assert config["selected_models"]["entity_detector"] == "gliner-pii-detector"
+    assert config["selected_models"]["entity_detector"]
```
kendrickb-nvidia: question: Is this intentional to assert not None and len > 0? Is this the recommended way to assert that?

If reading this in the future, my first guess is a typo that missed the == "blah". Consider a comment or a more explicit check like

```
entity_detector = config["selected_models"]["entity_detector"]
assert entity_detector is not none
assert len(entity_detector) > 0
```
Though I suppose linter may say to switch the last line to `assert entity_detector`, so maybe just a comment is the best.

tests/engine/test_model_loader.py
```diff
@@ -21,12 +21,12 @@ def test_load_workflow_config_contains_selected_models() -> None:
...
 
 
 def test_get_model_alias_reads_workflow_mapping() -> None:
     alias = get_model_alias(workflow_name=WorkflowName.detection, role="entity_validator")
-    assert alias == "gpt-oss-120b"
+    assert isinstance(alias, str) and alias
```
kendrickb-nvidia, line 37: comment: This is much clearer about the intention of the test and does not leave me wondering if the test has a typo or not.

kendrickb-nvidia (approved):
Some generic comments, no major concerns.

src/anonymizer/engine/schemas/rewrite.py
lipikaramaswamy: Will do in a separate cleanup PR, there are other instances of `(str, Enum)` too.

src/anonymizer/engine/schemas/rewrite.py
lipikaramaswamy: yeah that' a good point - for all other protection method suggestions, the specific recommendation of the model matters, but if an entity does not require protection then we can force the suggestion to be "left_as_is". Noted here, we will clean up in the next PR where the sensitivity disposition step/prompt is actually introduced.

src/anonymizer/engine/schemas/rewrite.py
lipikaramaswamy: Good point - it's not a standard term. I added clarification on what this means [up top](https://github.com/NVIDIA-NeMo/Anonymizer/pull/40/changes#diff-9fdcff2867e150dd0977f1132bcff38e3b33e23a839335dff69c758c143d9396R14), and an architecture diagram for rewrite to our [main issue](https://github.com/NVIDIA-NeMo/Anonymizer/issues/30#issue-4048279010) that hopefully helps!

src/anonymizer/engine/schemas/rewrite.py
lipikaramaswamy: thanks, good point - all are ints now.

src/anonymizer/engine/schemas/rewrite.py
lipikaramaswamy: added, thanks

src/anonymizer/engine/constants.py
lipikaramaswamy: I'm guessing the confusing ones here are those that correspond to custom columns and don't have pydantic schemas. `COL_NEEDS_HUMAN_REVIEW` will be bool.

tests/engine/test_model_loader.py
lipikaramaswamy: Updated the other test

src/anonymizer/engine/schemas/rewrite.py
lipikaramaswamy: it's for one entity that may have multiple spans, clarified

src/anonymizer/engine/schemas/rewrite.py
lipikaramaswamy: changed per suggestion

tests/engine/test_model_loader.py
lipikaramaswamy: updated

src/anonymizer/engine/schemas/rewrite.py
lipikaramaswamy: updated this class with noun properties and verb methods

Issue #9, opened 2026-02-25
feat: move prompts to versioned template files


lipikaramaswamy:
### Priority Level

Medium

### Task Summary

Detection and replacement prompts are currently inline Python strings in `detection_workflow.py` and `llm_replace_workflow.py`. Move them to versioned template files for easier review, diffing, and iteration without touching Python code.

### Technical Details & Implementation Plan

Create `src/anonymizer/engine/prompts/` directory with template files (e.g. `validation.txt`, `augmentation.txt`, `latent_detection.txt`, `replacement.txt`). Load at module level or lazily. Keep `<<PLACEHOLDER>>` tag convention for variable substitution. Consider whether Jinja blocks should stay inline or move to the template files.

Suggested by @andreatgretel  in PR #3  review.

PR #45, opened 2026-03-15
feat: domain classification and sensitivity disposition workflows

### LINKED ISSUE #31: feat(rewrite): engine — domain classification + sensitivity disposition

lipikaramaswamy:
## Summary

  Implements the first two rewrite pipeline steps as column factories, part of the
  broader single-workflow rewrite architecture tracked in #30.

  - `engine/rewrite/domain_classification.py` — `DomainClassificationWorkflow`
    - Classifies input text into one of 24 domains via `LLMStructuredColumnConfig`
    - `_enrich_domain` custom column looks up per-domain guidance from `DOMAIN_SUPPLEMENT_MAP`
    - `_DOMAIN_LIST` (for the prompt) and `DOMAIN_SUPPLEMENT_MAP` (for enrichment) are intentionally separate
  - `engine/rewrite/sensitivity_disposition.py` — `SensitivityDispositionWorkflow`
    - Produces a structured per-entity protection plan against `SensitivityDispositionSchema`
  - `engine/constants.py` — added `COL_DOMAIN_SUPPLEMENT`

  ## Design Decisions

  **Column factory pattern**: workflows expose `columns() -> list[ColumnConfigT]` only.
  No `run()` method — all steps will be collected and passed to a single `NddAdapter.run_workflow()` call in the top-level `RewriteWorkflow` (tracked in #30).

  **Prompt section headers**: standardized to XML tags (`<privacy_goal>`, `<input_tagged_text>`, etc.)
as XML provides the clearest semantic structure across several model families (gpt-oss, claude, nemotron)

  **Data summary label**: standardized to `Dataset description:` in prompts. Python param
  stays `data_summary` to match `AnonymizerConfig`.

  **Trust DataDesigner output types**: `_enrich_domain` accesses `.domain` directly on the
  `DomainClassificationSchema` Pydantic object — no defensive dict/fallback handling since
  `LLMStructuredColumnConfig` guarantees a valid schema instance.

  ## Follow-ups

  - [ ] Align entity detection prompts (`_get_validation_prompt`, `_get_augment_prompt`, `_get_latent_prompt`) to XML section headers and `Dataset description:` label (TODO in
  `sensitivity_disposition.py`)
## Related Issues
Closes #31


src/anonymizer/engine/rewrite/domain_classification.py
```diff
@@ -0,0 +1,449 @@
...
+
+# ---------------------------------------------------------------------------
+# Custom column generator
+# ---------------------------------------------------------------------------
+
+
+@custom_column_generator(required_columns=[COL_DOMAIN])
+def _enrich_domain(row: dict[str, Any]) -> dict[str, Any]:
+    """Look up domain-specific guidance from DOMAIN_SUPPLEMENT_MAP."""
+    domain = row[COL_DOMAIN].domain
```
andreatnvidia: Codex caught a runtime mismatch here: Data Designer hands structured outputs to custom columns as plain dicts, so `_domain.domain` blows up before `_domain_supplement` gets created. maybe read `row[COL_DOMAIN]["domain"]` or normalize through `DomainClassificationSchema.model_validate(...)` first?

tests/engine/test_domain_classification.py
```diff
@@ -0,0 +1,56 @@
...
+    cols = DomainClassificationWorkflow().columns(selected_models=stub_rewrite_model_selection)
+    assert len(cols) == 2
+    assert isinstance(cols[0], LLMStructuredColumnConfig)
+    assert isinstance(cols[1], CustomColumnConfig)
+    assert cols[0].name == COL_DOMAIN
+    assert cols[1].name == COL_DOMAIN_SUPPLEMENT
+    assert cols[0].model_alias == stub_rewrite_model_selection.domain_classifier
+
+
+def test_enrich_domain_populates_supplement_for_known_domain() -> None:
```
andreatnvidia, line 32: I think this test is using the wrong `_domain` shape. The workflow gets a dict back from `LLMStructuredColumnConfig`, so this passes even though the real column chain fails. could we switch it to the serialized dict form or add a tiny workflow-level check here?

src/anonymizer/engine/rewrite/domain_classification.py
```diff
@@ -0,0 +1,449 @@
...
+    (Domain.OTHER, "Use only if the text clearly does not fit any of the above domains."),
+]
+
+# ---------------------------------------------------------------------------
+# Domain supplement map — injected into the meaning unit extraction and
+# sensitivity disposition prompts via COL_DOMAIN_SUPPLEMENT to tell the LLM
+# what to preserve vs. drop when processing text in each domain.
+# ---------------------------------------------------------------------------
+
+DOMAIN_SUPPLEMENT_MAP: dict[Domain, str] = {
```
andreatnvidia, line 85: suggestion: this module is starting to carry a lot of prompt data along with the workflow wiring. totally fine for now, but if rewrite keeps growing it may be worth moving the domain supplement text into a dedicated prompt-data module so the workflow file stays focused on column orchestration.

src/anonymizer/engine/rewrite/sensitivity_disposition.py
```diff
@@ -0,0 +1,187 @@
...
+- Include ONE entry for EVERY provided latent entity.
+- IDs must be sequential starting from 1.
+
+QUALITY REQUIREMENTS:
+- protection_reason must be specific to this document (not generic boilerplate).
+- combined_risk_level must consider other entities being retained.
+</output_requirements>"""
+    return (
+        prompt.replace("<<PRIVACY_GOAL>>", privacy_goal_str)
+        .replace("<<DOMAIN>>", _jinja(COL_DOMAIN + "['domain']"))
```
andreatnvidia: nit: for Jinja access, `{{ _domain.domain }}` would be a bit more in line with the Data Designer examples than bracket syntax here. not a blocker, just slightly easier to read if we want to keep prompt templates stylistically consistent.

src/anonymizer/engine/rewrite/domain_classification.py
```diff
@@ -0,0 +1,449 @@
...
+    )
+
+
+# ---------------------------------------------------------------------------
+# Custom column generator
+# ---------------------------------------------------------------------------
+
+
+@custom_column_generator(required_columns=[COL_DOMAIN])
+def _enrich_domain(row: dict[str, Any]) -> dict[str, Any]:
```
andreatnvidia, line 418: one style thing from the rest of Anonymizer and Data Designer: cross-column payloads usually get normalized right at the boundary before we use them. elsewhere we do `*.from_raw(...)` / `model_validate(...)` instead of assuming the runtime shape directly. might be nice to wrap `_domain` the same way here so the custom column stays aligned with the engine contract and future changes are less brittle.

src/anonymizer/engine/rewrite/domain_classification.py
lipikaramaswamy: Ah this will recur in all subsequent additions, good call-out. Addressed.

src/anonymizer/engine/rewrite/domain_classification.py
lipikaramaswamy: For now, we’re keeping each rewrite step in its own module and focusing this PR on getting the column-factory pieces in place for the single-workflow orchestration tracked in #30. I agree the prompt data may deserve its own module if the rewrite stack keeps growing; we’ve noted that as follow-up cleanup #9, but I’d prefer to keep it out of the MVP path for now

src/anonymizer/engine/rewrite/sensitivity_disposition.py
lipikaramaswamy: Amended `_jinja()` to support keyed access since this pattern recurs throughout the rewrite workflow. Kept the shared helper form for consistency.

src/anonymizer/engine/rewrite/domain_classification.py
lipikaramaswamy: Addressed this by normalizing `COL_DOMAIN` with `DomainClassificationSchema.model_validate(...)` at the custom-column boundary before reading `domain`. So the column no longer assumes the runtime shape

tests/engine/test_domain_classification.py
lipikaramaswamy: Updated the test to use the dict-shaped `COL_DOMAIN` payload that `LLMStructuredColumnConfig` passes between columns, so it now matches the real runtime shape. I kept a schema-instance case as supplemental coverage

src/anonymizer/engine/constants.py
```diff
@@ -53,6 +53,7 @@
 
 COL_DOMAIN = "_domain"
 COL_DOMAIN_CONFIDENCE = "_domain_confidence"
```
andreatnvidia: this constant isn't referenced anywhere in the new code or tests - is it needed now, or can it wait until the column is actually produced/consumed by a step? happy to have it here if there's a near-term use, just flagging it as currently unreachable

src/anonymizer/engine/rewrite/domain_classification.py
```diff
@@ -0,0 +1,449 @@
...
+    """Look up domain-specific guidance from DOMAIN_SUPPLEMENT_MAP."""
+    parsed_domain = DomainClassificationSchema.model_validate(row.get(COL_DOMAIN, {}))
```
andreatnvidia: the PR description says we trust `LLMStructuredColumnConfig` output and skip defensive handling, but `row.get(COL_DOMAIN, {})` is a silent fallback - if `COL_DOMAIN` is absent (pipeline misconfiguration), `model_validate({})` raises a `ValidationError` about missing fields rather than a clear `KeyError` about a missing column. `row[COL_DOMAIN]` matches the stated intent and gives a better signal

src/anonymizer/engine/rewrite/domain_classification.py
```diff
@@ -0,0 +1,449 @@
...
+from anonymizer.engine.constants import COL_DOMAIN, COL_DOMAIN_SUPPLEMENT, COL_TEXT, _jinja
+from anonymizer.engine.ndd.model_loader import resolve_model_alias
+from anonymizer.engine.schemas import Domain, DomainClassificationSchema
+
+# ---------------------------------------------------------------------------
+# Domain list — short descriptions used in the domain classification prompt
+# to help the LLM identify which domain a given input text belongs to.
+# ---------------------------------------------------------------------------
+
+_DOMAIN_LIST: list[tuple[Domain, str]] = [
```
andreatnvidia, line 22: there's already a test that `DOMAIN_SUPPLEMENT_MAP` covers all `Domain` values - worth a matching one for `_DOMAIN_LIST` so a new domain can't slip through one without the other. alternatively, a single `(Domain, short_desc, supplement)` tuple list would make divergence impossible by construction. not blocking, just something to consider before the rest of the rewrite steps land

andreatnvidia:
> `src/anonymizer/engine/schemas/__init__.py`
>
> ```python
> __all__ = [
>     ...
>     "SensitivityLevel",
> ]
> ```

`generate_privacy_qa_from_disposition` is defined in `schemas/rewrite.py` and already tested, but it's missing from `__all__` here - every other rewrite type is exported so this looks like an oversight. downstream orchestration code will need this function and shouldn't have to go through the submodule directly

andreatnvidia (approved):
clean PR overall - the column-factory pattern is well-suited for the single-`NddAdapter.run_workflow()` call, and the Pydantic validation on LLM outputs is thorough. left a few inline nits: one missing export in `schemas/__all__`, an unused constant, a minor defensive-handling inconsistency in `_enrich_domain`, and a suggestion around exhaustiveness coverage for `_DOMAIN_LIST`. nothing blocking - approve.

asteier2026 (approved):

lipikaramaswamy:
> > `src/anonymizer/engine/schemas/__init__.py`
> > ```python
> > __all__ = [
> >     ...
> >     "SensitivityLevel",
> > ]
> > ```
> 
> `generate_privacy_qa_from_disposition` is defined in `schemas/rewrite.py` and already tested, but it's missing from `__all__` here - every other rewrite type is exported so this looks like an oversight. downstream orchestration code will need this function and shouldn't have to go through the submodule directly

PR #48 moves this function to its final resting place in the pipeline, so we handle exports there rather than adding another interim export from `schemas.__init__` on this branch.

src/anonymizer/engine/constants.py
lipikaramaswamy: Yeah originally I thought this was consumed downstream, but I checked the research implementation -- `domain_confidence` exists in the domain-classification schema there as well, but I don’t see any downstream consumer in the pipeline logic. I'll drop it.

src/anonymizer/engine/rewrite/domain_classification.py
lipikaramaswamy: We have a matching exhaustiveness [test](https://github.com/NVIDIA-NeMo/Anonymizer/pull/45/changes#diff-a73a042bcd74b27487bd3936e83ddf56a429b293ad0e52c671161b94fbe30cb4R69) for `_DOMAIN_LIST` so it stays aligned with `Domain`, alongside the existing `DOMAIN_SUPPLEMENT_MAP` coverage - `test_domain_list_covers_all_domains`. 

That said, I agree a single combined structure would make divergence impossible by construction, but tracking as a follow up #55

andreatnvidia (approved):
looks good, ship it!

PR #48, opened 2026-03-16
feat: add QA generation workflow for rewrite pipeline

### LINKED ISSUE #32: feat(rewrite): engine — QA generation
Part of #30.

## Scope

**`engine/rewrite/qa_generation.py`** — `QaGenerationWorkflow`:

Three stages:

**Stage 1 — Meaning unit extraction (LLM)**
- Pre-step (pure Python): format `COL_SENSITIVITY_DISPOSITION` → `COL_SENSITIVITY_DISPOSITION_BLOCK`
- NddAdapter call → `COL_MEANING_UNITS` (MeaningUnitsSchema, `meaning_extractor` alias)
- Prompt uses `COL_SENSITIVITY_DISPOSITION_BLOCK` and `COL_DOMAIN_SUPPLEMENT` as Jinja2 vars

**Stage 2 — Quality QA generation (LLM)**
- Pre-step (pure Python): serialize `COL_MEANING_UNITS` → `COL_MEANING_UNITS_SERIALIZED`
- NddAdapter call → `COL_QUALITY_QA` (QualityQAPairsSchema, `qa_generator` alias)
- Prompt: "Does the rewritten text still convey X?" questions from meaning units

**Stage 3 — Privacy QA generation (pure Python, no LLM)**
- `_generate_privacy_qa_from_disposition(disposition)`: for each entity where `does_need_protection=True`, generate "Can [entity_value] be deduced from the rewritten text?"
- Applied row-by-row → `COL_PRIVACY_QA` (PrivacyQAPairsSchema shape)
- Failed records accumulated from stages 1 and 2 only

**Tests**: `tests/engine/test_qa_generation.py`

## Acceptance criteria
- [ ] Stage 1 NddAdapter call uses `meaning_extractor` alias
- [ ] Stage 2 NddAdapter call uses `qa_generator` alias
- [ ] Stage 3 produces correct number of privacy QA items from stub disposition
- [ ] Failed records from stages 1+2 accumulated; stage 3 never adds failures
- [ ] `df.attrs` propagated

lipikaramaswamy:
## Summary
Implements `QAGenerationWorkflow` for the rewrite pipeline. This stage runs after sensitivity disposition and produces the QA artifacts used downstream to evaluate rewrite quality and privacy leakage.

  - Adds `src/anonymizer/engine/rewrite/qa_generation.py`
  - `QAGenerationWorkflow.columns()`
  - `_format_disposition_block()` for prompt-safe disposition serialization
  - `_serialize_meaning_units()` for quality-QA prompt injection
  - `generate_privacy_qa_from_disposition()` and `_generate_privacy_qa_column()` for template-based privacy QA
  - Meaning unit extraction runs through `LLMStructuredColumnConfig` using the `meaning_extractor` alias
  - Quality QA generation runs through `LLMStructuredColumnConfig` using the `qa_generator` alias
  - Privacy QA is generated without an LLM call from protected entities in `SensitivityDispositionSchema`

  ## Design Decisions

Structured cross-column payloads are normalized at the custom-column boundary with `model_validate(...)` rather than assuming live schema instances. Prompt sections use the same XML-style layout introduced in #45 

  ## Dependencies

Depends on #45 for domain classification and sensitivity disposition.

  ## Testing

Tests pass locally and cover column ordering, model alias wiring, disposition/meaning-unit serialization, privacy-QA generation, and prompt column references.

  ## Related Issues

Closes #32

### DIFF
--- a/src/anonymizer/engine/constants.py
+++ b/src/anonymizer/engine/constants.py
@@ -145,6 +145,11 @@ DEFAULT_ENTITY_LABELS: list[str] = list(ENTITY_LABEL_EXAMPLES.keys())
 # ---------------------------------------------------------------------------
 
 
-def _jinja(col: str) -> str:
-    """Wrap a column name in Jinja2 template syntax for use in NDD prompts."""
-    return "{{ " + col + " }}"
+def _jinja(col: str, *, key: str | None = None) -> str:
+    """Wrap a column name in Jinja2 template syntax for use in NDD prompts.
+
+    When *key* is given the expression becomes ``{{ col['key'] }}``,
+    providing a single grep-able call site for nested schema access.
+    """
+    expr = col if key is None else f"{col}['{key}']"
+    return "{{ " + expr + " }}"
--- a/src/anonymizer/engine/constants.py
+++ b/src/anonymizer/engine/constants.py
@@ -53,8 +53,11 @@ COL_FINAL_ENTITIES = "final_entities"
 
 COL_DOMAIN = "_domain"
 COL_DOMAIN_CONFIDENCE = "_domain_confidence"
+COL_DOMAIN_SUPPLEMENT = "_domain_supplement"
 COL_SENSITIVITY_DISPOSITION = "_sensitivity_disposition"
+COL_SENSITIVITY_DISPOSITION_BLOCK = "_sensitivity_disposition_block"
 COL_MEANING_UNITS = "_meaning_units"
+COL_MEANING_UNITS_SERIALIZED = "_meaning_units_serialized"
 COL_QUALITY_QA = "_quality_qa"
 COL_PRIVACY_QA = "_privacy_qa"
 COL_REWRITTEN_TEXT = "_rewritten_text"  # pre-repair intermediate; renamed to {text_col}_rewritten in user output
--- a/src/anonymizer/engine/rewrite/qa_generation.py
+++ b/src/anonymizer/engine/rewrite/qa_generation.py
@@ -1,0 +1,299 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+from __future__ import annotations
+
+import json
+from typing import Any
+
+from data_designer.config import custom_column_generator
+from data_designer.config.column_configs import CustomColumnConfig, LLMStructuredColumnConfig
+from data_designer.config.column_types import ColumnConfigT
+
+from anonymizer.config.models import RewriteModelSelection
+from anonymizer.engine.constants import (
+    COL_DOMAIN,
+    COL_DOMAIN_SUPPLEMENT,
+    COL_MEANING_UNITS,
+    COL_MEANING_UNITS_SERIALIZED,
+    COL_PRIVACY_QA,
+    COL_QUALITY_QA,
+    COL_SENSITIVITY_DISPOSITION,
+    COL_SENSITIVITY_DISPOSITION_BLOCK,
+    COL_TEXT,
+    _jinja,
+)
+from anonymizer.engine.ndd.model_loader import resolve_model_alias
+from anonymizer.engine.schemas import (
+    EntityCategory,
+    MeaningUnitsSchema,
+    PrivacyQAPairsSchema,
+    PrivacyQuestionSchema,
+    QualityQAPairsSchema,
+    SensitivityDispositionSchema,
+    SensitivityLevel,
+)
+
+# ---------------------------------------------------------------------------
+# Stage 1 pre-step: format disposition → disposition block
+# ---------------------------------------------------------------------------
+
+
+@custom_column_generator(required_columns=[COL_SENSITIVITY_DISPOSITION])
+def _format_disposition_block(row: dict[str, Any]) -> dict[str, Any]:
+    """Serialize sensitivity disposition into a JSON block for the meaning unit extraction prompt."""
+    disposition: SensitivityDispositionSchema = row[COL_SENSITIVITY_DISPOSITION]
+    block = [
+        {
+            "entity_value": e.entity_value,
+            "needs_protection": e.needs_protection,
+            "protection_method": e.protection_method_suggestion,
+            "category": e.category,
+        }
+        for e in disposition.sensitivity_disposition
+    ]
+    row[COL_SENSITIVITY_DISPOSITION_BLOCK] = json.dumps(block, ensure_ascii=False)
+    return row
+
+
+# ---------------------------------------------------------------------------
+# Stage 1 prompt: meaning unit extraction
+# ---------------------------------------------------------------------------
+
+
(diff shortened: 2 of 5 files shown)


src/anonymizer/engine/rewrite/qa_generation.py
```diff
@@ -0,0 +1,304 @@
...
+    MeaningUnitsSchema,
+    PrivacyQAPairsSchema,
+    PrivacyQuestionSchema,
+    QualityQAPairsSchema,
+    SensitivityDispositionSchema,
+    SensitivityLevel,
+)
+
+# Derived from the schema so the Jinja key stays in sync with the field name.
+_DOMAIN_KEY = next(name for name, info in DomainClassificationSchema.model_fields.items() if info.annotation is Domain)
```
andreatnvidia: the `_DOMAIN_KEY` derivation is clever and stays in sync with the schema. one small thing: `next()` on an empty iterator raises `StopIteration`, which python 3.7+ re-raises as `RuntimeError` if it escapes a generator frame - and since this runs at module load time the error message would be confusing. suggest `next(..., None)` plus an `assert _DOMAIN_KEY is not None`. low priority since the schema is stable, but it's a cheap fix.

src/anonymizer/engine/constants.py
```diff
@@ -53,8 +53,11 @@
 
 COL_DOMAIN = "_domain"
 COL_DOMAIN_CONFIDENCE = "_domain_confidence"
+COL_DOMAIN_SUPPLEMENT = "_domain_supplement"
```
andreatnvidia, line 55: `_domain_supplement` is referenced in the meaning unit extraction prompt but i don't see where it gets populated - is this set by the domain classification step? if it can be empty/missing at runtime, worth a comment on the constant so future readers know where to look.

andreatnvidia:
> `src/anonymizer/engine/schemas/rewrite.py:167`
>
> ```python
> sensitivity_disposition: list[EntityDispositionSchema] = Field(min_length=1)
> ```

worth a docstring note that this field enforces `min_length=1` by design - the expectation is that the orchestrator short-circuits before this step if no entities were detected. it's a non-obvious contract; a future caller constructing this schema directly (e.g. in tests) will get a confusing `ValidationError` on an empty list with no hint as to why

andreatnvidia:
one thing to keep in mind when you get to the artifact storage layer: `trace_dataframe` contains `_privacy_qa`, and each question in there embeds the raw entity value verbatim (e.g. `"Can the first_name 'Alice' be deduced from the rewritten text?"`). so if `trace_dataframe` gets written to `artifact_path` as parquet/csv, those files would contain PII in the question strings. probably worth a scrub step or a separate privacy-safe artifact before persisting.

src/anonymizer/engine/rewrite/qa_generation.py
```diff
@@ -0,0 +1,304 @@
...
+    prompt = """You are extracting IMPORTANT MEANING UNITS from a text to evaluate anonymization quality.
+The original text may contain sensitive entities; a separate system will rewrite it.
+Your job is to capture the core informational content that MUST be preserved for the text to remain
+useful, while avoiding unsafe disclosure of identifying details.
+
+<entity_protection_rules>
+You are given an ENTITY PROTECTION BLOCK — a JSON array where each entity has an "entity_value" string
+(e.g., a name, address, occupation, city, diagnosis). Treat every entity_value where needs_protection=true
+as a BANNED SURFACE FORM:
+
```
asteier2026: We really want to do a Hard ban on "does_need_protection = True
  AND protection_method_suggestion is "replace" OR "remove"", if it's generalize then we create a meaning unit with entity generalized.

src/anonymizer/engine/rewrite/qa_generation.py
```diff
@@ -0,0 +1,304 @@
...
+<<TEXT>>
+---
+</input>"""
+    return (
+        prompt.replace("<<ENTITY_PROTECTION_BLOCK>>", _jinja(COL_SENSITIVITY_DISPOSITION_BLOCK))
+        .replace("<<DOMAIN>>", _jinja(COL_DOMAIN, key=_DOMAIN_KEY))
+        .replace("<<DOMAIN_SUPPLEMENT>>", _jinja(COL_DOMAIN_SUPPLEMENT))
+        .replace("<<TEXT>>", _jinja(COL_TEXT))
+    )
+
```
asteier2026, line 194: Lipika, this prompt changed dramatically from GitLab. I spent an enormous amount of time tweaking it to do the right thing. If we want to try alternative prompts, I think that should be post P0. It will be very time-consuming to test.

asteier2026 (commented):
The meaning unit prompt was changed too dramatically from what it was is GitLab. It was a lot of work to tweak it to do the right thing. If we want to do something different I think it should be post P0 as it will require exhaustive testing.

src/anonymizer/engine/rewrite/qa_generation.py
lipikaramaswamy: Unintentional. Should be back to matching gitlab now, please review.

src/anonymizer/engine/constants.py
lipikaramaswamy: It's populated by the domain classification step (`_enrich_domain` in `domain_classification.py`), which was merged in #45. I've rebased now so it should be clearer.

src/anonymizer/engine/rewrite/qa_generation.py
lipikaramaswamy: Added a robust `RuntimeError` below.

src/anonymizer/engine/rewrite/qa_generation.py
lipikaramaswamy: The A/B/C branching added now implements the hard-ban replace/remove and allow generalized meaning units for generalize/paraphrase. The GitLab meaning unit prompt didn't differentiate by protection method — it banned all entities uniformly. 

The rest of the prompt (importance criteria, segmentation rules, domain context) is structurally the same as GitLab, just reformatted to XML tags per #45 conventions.

lipikaramaswamy:
> one thing to keep in mind when you get to the artifact storage layer: `trace_dataframe` contains `_privacy_qa`, and each question in there embeds the raw entity value verbatim (e.g. `"Can the first_name 'Alice' be deduced from the rewritten text?"`). so if `trace_dataframe` gets written to `artifact_path` as parquet/csv, those files would contain PII in the question strings. probably worth a scrub step or a separate privacy-safe artifact before persisting.

The trace dataframe already contains the raw text, tagged text, entity values, and sensitivity disposition — all of which carry the same PII. The privacy QA question strings don't introduce new exposure. When we build an artifact storage layer, we'll probably need to handle the entire trace dataframe as sensitive, not just this column. But that's probably a service consideration, Anonymizer is just a library for now...

andreatnvidia (approved):
re-reviewed the latest diff and the earlier comments look addressed. focused checks passed too. approving.

asteier2026 (approved):

PR #49, opened 2026-03-17
feat: add rewrite generation workflow

### LINKED ISSUE #33: feat(rewrite): engine — rewrite generation
## Scope

  **`engine/rewrite/rewrite_generation.py`** — `RewriteGenerationWorkflow`:

  **Fast path**: rows with no entities get `COL_REWRITTEN_TEXT = COL_TEXT` immediately, skipping all LLM calls.

  **Step 1 — Replacement map (LLM)**
  - Reuse existing `LlmReplaceWorkflow.generate_map_only()` on rows with entities
  - Requires `COL_ENTITIES_BY_VALUE` (already computed by detection for all modes)
  - Produces `COL_REPLACEMENT_MAP` — consistent synthetic values the rewriter should use for `replace`-method entities
  - Note: whether to always generate the replacement map or skip it when disposition suggests only generalize/remove is left as a `# TODO` for future optimization

  **Step 2a — Disposition block (pure Python)**
  - Build `COL_REWRITE_DISPOSITION_BLOCK` from `COL_SENSITIVITY_DISPOSITION`
  - Include only `needs_protection=True` entities
  - Serialize per-entity: label, value, sensitivity, protection method, reason

  **Step 2b — Prompt replacement map filter (pure Python)**
  - Build `COL_REPLACEMENT_MAP_FOR_PROMPT` from `COL_REPLACEMENT_MAP`
  - Keep only replacements for entities whose `protection_method_suggestion == "replace"`

  **Step 3 — Rewrite (LLM)**
  - `NddAdapter` call produces `COL_FULL_REWRITE` (`RewriteOutputSchema`, `rewriter` alias)
  - Prompt uses:
    - `COL_TAGGED_TEXT`
    - `COL_REWRITE_DISPOSITION_BLOCK`
    - `COL_REPLACEMENT_MAP_FOR_PROMPT`
  - The prompt follows disposition decisions directly rather than using generic entity-type context

  **Step 4 — Extract text (pure Python)**
  - Read `rewritten_text` from `COL_FULL_REWRITE`
  - Accept either structured dict payloads or `RewriteOutputSchema` instances
  - Fallback to original text on failure

  Re-merge rewritten and passthrough rows while preserving original record order.

  **Tests**: `tests/engine/test_rewrite_generation.py`

  ## Acceptance criteria
  - [ ] Rows with no entities skip LLM calls and get passthrough text
  - [ ] `LlmReplaceWorkflow.generate_map_only()` runs before the rewrite adapter call
  - [ ] Rewrite adapter call uses the `rewriter` alias
  - [ ] Mixed passthrough/entity outputs preserve original input row order
  - [ ] `COL_FULL_REWRITE` extraction works for both dict payloads and schema instances
  - [ ] Only `replace`-method entities appear in `COL_REPLACEMENT_MAP_FOR_PROMPT`
  - [ ] Output length equals input length
  - [ ] `df.attrs` are propagated

lipikaramaswamy:
## Summary

Implements `RewriteGenerationWorkflow` for the rewrite pipeline. This stage runs after sensitivity disposition and handles records with detected entities by generating replacement maps, preparing rewrite prompt inputs, invoking the `rewriter` model, and extracting rewritten text for downstream evaluation.

  - Reuses `LlmReplaceWorkflow.generate_map_only()` to produce `COL_REPLACEMENT_MAP`
  - Builds `COL_REWRITE_DISPOSITION_BLOCK` from `COL_SENSITIVITY_DISPOSITION`
  - Filters `COL_REPLACEMENT_MAP_FOR_PROMPT` to `replace`-method entities only
  - Produces `COL_FULL_REWRITE` through `LLMStructuredColumnConfig` and extracts `COL_REWRITTEN_TEXT`
  - Preserves passthrough behavior, `df.attrs`, failed-record accumulation, and original row order

## Design Decisions

Structured cross-column payloads are normalized at the custom-column boundary rather than assuming live schema instances. The rewrite prompt uses the disposition block directly instead of a generic entity-context layer.

## Testing

Tests pass locally and cover passthrough behavior, replacement-map filtering, prompt construction, schema/dict output extraction, failed-record propagation, attribute propagation, and mixed-row order preservation.

## Related Issues

Closes #33


src/anonymizer/engine/constants.py
```diff
@@ -68,6 +71,7 @@
 COL_LEAKAGE_MASS = "leakage_mass"
 COL_ANY_HIGH_LEAKED = "any_high_leaked"
 COL_NEEDS_HUMAN_REVIEW = "needs_human_review"
+COL_REWRITE_GENERATION_ROW_ORDER = "_rewrite_generation_row_order"
```
andreatnvidia: minor: `COL_REWRITE_GENERATION_ROW_ORDER` ends up in the "user-facing output columns" section but its `_` prefix and usage make it an internal column. would be cleaner to move it up into the internal-columns block alongside `COL_REWRITE_DISPOSITION_BLOCK` etc.

src/anonymizer/engine/rewrite/rewrite_generation.py
```diff
@@ -0,0 +1,351 @@
...
+def _extract_rewritten_text(row: dict[str, Any]) -> dict[str, Any]:
+    """Extract rewritten_text from the LLM structured output, falling back to original on failure."""
+    try:
+        full_rewrite = row[COL_FULL_REWRITE]
+        if hasattr(full_rewrite, "model_dump"):
+            full_rewrite = full_rewrite.model_dump(mode="python")
+        if isinstance(full_rewrite, dict):
+            row[COL_REWRITTEN_TEXT] = str(full_rewrite["rewritten_text"])
+        else:
+            row[COL_REWRITTEN_TEXT] = str(full_rewrite.rewritten_text)
```
andreatnvidia: the else-branch (`str(full_rewrite.rewritten_text)`) is unreachable in practice - `LLMStructuredColumnConfig` returns either a pydantic model (caught by `hasattr(model_dump)`) or a dict. the bare except below already handles any failure. worth removing the else or documenting what third type it is intended to handle.

tests/engine/test_rewrite_generation.py
```diff
@@ -0,0 +1,686 @@
...
+            COL_ENTITIES_BY_VALUE: [
+                stub_entities_by_value_with_entities,
+                {"entities_by_value": []},
+            ],
+            COL_SENSITIVITY_DISPOSITION: [stub_sensitivity_disposition, {"sensitivity_disposition": []}],
+            COL_REPLACEMENT_MAP: [stub_replacement_map, {"replacements": []}],
+        }
+    )
+
+    replace_result_df = pd.DataFrame(
```
andreatnvidia, line 528: `replace_result_df` here (and in `test_run_collects_failed_records_from_replace_and_rewrite` around line 653) doesn't include `COL_REWRITE_GENERATION_ROW_ORDER`. in production `generate_map_only` preserves it as a seed column, so the mocks silently diverge. adding that column to those fixtures would improve mock fidelity - tests pass as-is because `rewrite_result_df` injects the column back, but it hides a potential future breakage.

src/anonymizer/engine/rewrite/rewrite_generation.py
```diff
@@ -0,0 +1,351 @@
...
+    except Exception:
```
andreatnvidia, line 202: if the LLM returns an empty or whitespace-only `rewritten_text`, this stores `""` silently - no fallback to the original text, no warning logged. the except handles exceptions but not blank outputs. a model malfunction that returns empty rewrites would produce silent data loss in the output dataset. worth at least a `logger.warning` when the result is blank.

src/anonymizer/engine/rewrite/rewrite_generation.py
```diff
@@ -0,0 +1,351 @@
...
+def _filter_replacement_map_for_prompt(row: dict[str, Any]) -> dict[str, Any]:
+    """Keep only replacement entries for entities with protection_method_suggestion='replace'."""
+    disposition_block: list[dict] = row.get(COL_REWRITE_DISPOSITION_BLOCK, [])
+    replace_values = {
+        e["entity_value"] for e in disposition_block if e.get("protection_method_suggestion") == "replace"
+    }
+    raw_map = row.get(COL_REPLACEMENT_MAP, {})
+    if hasattr(raw_map, "model_dump"):
+        raw_map = raw_map.model_dump(mode="python")
+    parsed_map = EntityReplacementMapSchema.model_validate(raw_map if isinstance(raw_map, dict) else {})
```
andreatnvidia: if `COL_REPLACEMENT_MAP` is explicitly `None` (not absent), `model_validate({})` runs silently and produces an empty replacement list - the rewrite prompt gets no entity substitution context with no log message. combining this with the blank-output case above, a row could silently get rewritten with no guidance and come back empty. a `logger.debug` or `logger.warning` here would make this much easier to diagnose.

asteier2026 (commented):
The wording is not all that different from the GItLab prompt, but recall in GitLab even though we said only use the Replacement Map for entities tagged with Replace, and for tags that specify generalize, then generalize, nonetheless, often times the Replacement map was still used for entities tagged with generalize. Remember the plan to only feed into the rewrite prompt the portion of the Replacement map that contained direct identifiers.

lipikaramaswamy:
> The wording is not all that different from the GItLab prompt, but recall in GitLab even though we said only use the Replacement Map for entities tagged with Replace, and for tags that specify generalize, then generalize, nonetheless, often times the Replacement map was still used for entities tagged with generalize. Remember the plan to only feed into the rewrite prompt the portion of the Replacement map that contained direct identifiers.

@asteier2026 We already handle it in this PR: `_filter_replacement_map_for_prompt` [filters](https://github.com/NVIDIA-NeMo/Anonymizer/pull/49/changes#diff-47f79aba450dc60926670cd0e490974c424c663335f429e13d642838c0960e34R157) `COL_REPLACEMENT_MAP` down to only entities where the disposition says `protection_method_suggestion="replace"`, and the prompt references this filtered version (`COL_REPLACEMENT_MAP_FOR_PROMPT`) rather than the raw map. The <replacement_map> section is also wrapped in a Jinja conditional so it's omitted entirely when there are no "replace"-method entities — the LLM never sees replacement values for generalize/remove/paraphrase entities.

src/anonymizer/engine/constants.py
lipikaramaswamy: actually removed from constants as it is immediately dropped from the dataframe after rewrite is completed.

src/anonymizer/engine/rewrite/rewrite_generation.py
lipikaramaswamy: cleaned up, no third type it handles

tests/engine/test_rewrite_generation.py
lipikaramaswamy: added the temp col to the fixture

src/anonymizer/engine/rewrite/rewrite_generation.py
lipikaramaswamy: added a warning, and stored as None type - also added a followup TODO in the docstring

src/anonymizer/engine/rewrite/rewrite_generation.py
lipikaramaswamy: Makes sense. I added an early-return for `None` with a conditional warning — it only logs when there are entities that actually need replacement (`protection_method_suggestion="replace"`), so it won't noise up logs when the disposition is all generalize/remove/paraphrase.

asteier2026 (approved):

andreatnvidia (approved):
Approving - non-blocking stuff to track:

- The orchestrator sets `compute_grouped_entities=True` only when `config.replace is not None`. This workflow needs `COL_ENTITIES_BY_VALUE`, so when rewrite gets wired in that condition needs to cover `config.rewrite` too or you'll get a `KeyError`. Worth a TODO somewhere.
- This is the third workflow doing split-on-has-entities / reorder / recombine, and the row-order column name already diverged (`_anonymizer_row_order` vs `_row_order`). Might be time to pull that into a shared helper before the next one copies it again.
- `_has_entities` swallows all parse errors silently (`except Exception: return False`) - a `logger.debug` would help when rows unexpectedly skip rewriting. Also does full schema validation just to check emptiness when a dict `.get()` would do.

PR #51, opened 2026-03-17
feat: port developer tooling and build from Safe-Synthesizer repo

binaryaaron:
## Summary

Foundation infrastructure PR that ports developer tooling from Safe-Synthesizer and adapts it for Anonymizer.

## Key files created/changed

- `tools/binaries/` -- bootstrap scripts for standalone tools (uv, ty, yq, gh)
- `tools/codestyle/` -- format, lint, typecheck, copyright enforcement scripts
- `ruff.toml` -- extracted from pyproject.toml; stricter rules commented out for gradual adoption
- `pyproject.toml` -- docs dep group, ty in dev deps, uv-dynamic-versioning, pytest markers
- `.pre-commit-config.yaml` -- local hooks (format, typecheck, copyright-fix, uv-lock, DCO signoff)
- `Makefile` -- new targets: bootstrap-tools, bootstrap, format, format-check, typecheck, check, lock-check, test-slow, test-e2e, build-wheel, publish-internal, publish-pypi, test-container
- `.python-version`, `.copyrightignore`, `.vscode/settings.json`
- `tools/binaries/test_tool_install.sh` -- container smoke test

## Key decisions

- **Ruff stays as a dev dependency** (unlike Safe-Synthesizer which manages it as a standalone binary). Removed from tools.yaml; available via `uv run` in the venv.
- **uv-dynamic-versioning** replaces hardcoded version -- version derived from git tags
- **Pre-commit uses local hooks** via `uv run` -- ensures consistent ruff/ty versions from the venv
- **`.vscode/settings.json` is force-added** despite `.vscode/` being in .gitignore (same approach as Safe-Synthesizer)
- **copyright_fixer.py uses PEP 723 inline script deps** -- runs via `uv run --script`

## Testing

- `uv sync --group dev` succeeds
- No Safe-Synthesizer references remain in ported files

Made with [Cursor](https://cursor.com)


tools/binaries/bootstrap_tools.sh
```diff
@@ -0,0 +1,245 @@
+#!/usr/bin/env bash
```
lipikaramaswamy: I'm trying to understand why we need a repo-owned bootstrap toolchain here. The development workflow in Anonymizer is pretty straightforward today, so I’m trying to understand which setup/infra pieces are value-adds.  Is there a contributor or CI workflow that improves here beyond our current uv based install flow?

More generally, how are you thinking about the tradeoff between managing installation logic for tools like uv, yq, gh, ty, etc. in this repo versus relying on documented prerequisites plus normal project dependency setup?

Makefile
```diff
@@ -3,25 +3,46 @@ help:
...
 	@echo "  install-pre-commit     - Install pre-commit hooks (run once after cloning)"
 	@echo ""
+	@echo "  format                 - Format and fix code"
+	@echo "  format-check           - Check format and lint (read-only)"
+	@echo "  typecheck              - Run type checks (advisory, non-blocking)"
+	@echo "  check                  - Run all read-only checks"
+	@echo "  lock-check             - Check uv.lock is up to date"
+	@echo ""
 	@echo "  test                   - Run all unit tests"
+	@echo "  test-slow              - Run slow tests"
```
lipikaramaswamy: don't think we have any tests marked as slow right now, the only ones that are slow that may come in the future are e2e tests

Makefile
```diff
@@ -3,25 +3,46 @@ help:
...
+	@echo "  bootstrap-tools-ci     - Install CI tools (non-interactive)"
+	@echo "  bootstrap              - Install Python dependencies (dev group)"
 	@echo "  install                - Install project dependencies with uv"
 	@echo "  install-dev            - Install project with dev dependencies"
 	@echo "  install-dev-notebooks  - Install dev + notebook dependencies"
 	@echo "  install-pre-commit     - Install pre-commit hooks (run once after cloning)"
 	@echo ""
+	@echo "  format                 - Format and fix code"
+	@echo "  format-check           - Check format and lint (read-only)"
+	@echo "  typecheck              - Run type checks (advisory, non-blocking)"
```
lipikaramaswamy, line 14: I’m a bit overwhelmed by typechecking, especially given #53. My initial read is that `ty` is currently flagging a lot of issues here, and some of them are due to the gap between static analysis and the way we use Pydantic throughout the codebase. That'll likely be a recurring source of noise for Anonymizer -- does it belong in the repo right now? Or from your experience do you expect it to pay for itself by helping us catch bugs early?

Makefile
```diff
@@ -3,25 +3,46 @@ help:
...
 	@echo "  coverage               - Run tests with coverage report"
 	@echo ""
-	@echo "  format                 - Format code with ruff"
-	@echo "  format-check           - Check code formatting without changes"
-	@echo "  lint                   - Lint with ruff"
-	@echo "  lint-fix               - Lint with ruff and apply fixes"
-	@echo "  check-all              - Run all checks (format-check + lint)"
-	@echo "  check-all-fix          - Run all checks with autofix (format + lint-fix)"
+	@echo "  build-wheel            - Build wheel (version from git tag)"
+	@echo "  publish-internal       - Publish wheel to NVIDIA Artifactory"
```
lipikaramaswamy: Is this something that is required for `nmp`? I’m not currently seeing the Anonymizer use case for publishing to Artifactory

Makefile
```diff
@@ -3,25 +3,46 @@ help:
...
+	@echo "  publish-pypi           - Publish wheel to PyPI"
+	@echo "  test-container         - Run container smoke test"
```
lipikaramaswamy: why do we need a container smoke test?

Makefile
```diff
@@ -42,34 +63,70 @@ install-pre-commit:
 	uv run pre-commit install
 	@echo "Done! Hooks will run on git commit."
 
+format:
+	@echo "Formatting and fixing code..."
+	uv run tools/codestyle/format.sh
+
+format-check:
```
lipikaramaswamy, line 65: So this target calls `ruff_check.sh`, which does the lint check - but isn't it also meant to cover `ruff format --check`?

tools/codestyle/copyright_fixer.py
```diff
@@ -0,0 +1,302 @@
+#!/usr/bin/env -S uv run --script
```
lipikaramaswamy, line 1: IIUC, this file is intended to ensure we consistently add copyright headers to source files. Is that right?

tools/codestyle/format.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
```
lipikaramaswamy, line 1: Could you say a bit more about what this wrapper is intended to buy us in Anonymizer? Since our Ruff workflow is currently pretty direct, I’m trying to understand what problem this extra codestyle layer is solving for. Same for `ruff_check.sh` 🤔

tools/diff-lockfile.py
```diff
@@ -0,0 +1,369 @@
+#!/usr/bin/env -S uv run
```
lipikaramaswamy: We have a fairly small dependency surface in Anonymizer, so just wondering how often would this tool be useful in practice

.pre-commit-config.yaml
```diff
@@ -13,10 +14,41 @@ repos:
...
+        files: pyproject.toml
+        stages: [pre-commit]
+        pass_filenames: false
+
+      - id: check-signoff
+        name: Check commit is signed off (DCO)
+        entry: "grep -qE '^Signed-off-by: .+ <.+>'"
+        language: system
+        stages: [commit-msg]
+        always_run: true
```
lipikaramaswamy, line 54: When I was getting the DCO assistant check passing in Anonymizer, I ended up following the Data Designer style setup with the repo secret plus a signatures branch. How does that play with this local DCO enforcement in the pre commit hook?

ruff.toml
```diff
@@ -0,0 +1,51 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
lipikaramaswamy, line 1: Curious - is it common practice  to separate the Ruff config into `ruff.toml`? This is roughly the configuration we had in `pyproject.toml` so just wondering if this is a structural preference

tools/binaries/bootstrap_tools.sh
binaryaaron: You're right — for Anonymizer's current scale, a repo-owned bootstrap toolchain is overhead. Everything a developer needs (ruff, ty, pre-commit, pytest) is available via `uv sync --group dev`. The bootstrap infrastructure was carried over from Safe-Synthesizer where the tool surface and team size justify it, but Anonymizer doesn't need it today.

**Change**: Removed `tools/binaries/` entirely. `make bootstrap-tools` and `bootstrap-tools-ci` are gone. The getting-started flow is now: install uv (documented prerequisite) → `uv sync --group dev` → `uv run pre-commit install`. CI uses `astral-sh/setup-uv` directly.

Makefile
binaryaaron: Agreed. Removed the `test-slow` target and `slow` pytest marker. Keeping `test-e2e` since that's the realistic future need. We can reintroduce test tiers when we actually have tests that need them.

Makefile
binaryaaron: Reasonable concern. The 41 diagnostics in #53 are real — not false positives — but most stem from Pydantic's runtime behavior being invisible to static analysis (dict-vs-model patterns, dynamic validators). That's a known gap in `ty` today.

Why keep it:
- It's **strictly advisory** — zero friction for developers. The pre-commit hook uses `|| true`, the Makefile target ignores failures, CI uses `continue-on-error: true`. Nobody is blocked by it.
- The diagnostics that *aren't* Pydantic-related (wrong arg count, missing imports, unreachable code) are genuine catches.
- As `ty` matures its Pydantic plugin story (tracked upstream), the signal-to-noise ratio will improve without us changing anything.

If the noise is distracting even as advisory, happy to defer it entirely — remove from pre-commit and CI, keep only as a `make typecheck` target that developers can opt into manually. #53 tracks the resolution either way.

Makefile
binaryaaron: Not immediately needed — this was ported from Safe-Synthesizer's NMP distribution workflow. Anonymizer doesn't ship via Artifactory today.

**Change**: Removed `publish-internal` Makefile target and `dev-wheel.yml` workflow. Simplified `internal-release.yml` to PyPI-only publishing. We can add Artifactory back when there's an actual distribution target.

Makefile
binaryaaron: The original purpose was to validate that the bootstrap + install flow works in a clean environment (catches "works on my machine" issues). With the bootstrap toolchain removed, this no longer has a clear purpose — CI's own `uv sync` step serves as the implicit smoke test.

**Change**: Removed `test-container` Makefile target, `test_tool_install.sh`, and `container-test.yml` workflow.

Makefile
binaryaaron: Good catch — this was a bug. `format-check` was only calling `ruff_check.sh` (lint rules) but not `ruff format --check` (formatting verification).

**Change**: Fixed to run both:
```makefile
format-check:
	uv run tools/codestyle/format.sh --check
	uv run tools/codestyle/ruff_check.sh
```
The first line checks formatting, the second checks lint rules.

tools/codestyle/copyright_fixer.py
binaryaaron: Correct. It scans source files for SPDX copyright headers and adds them where missing. It respects `.copyrightignore` for files that shouldn't have headers (third-party, generated, etc.). Runs as a pre-commit hook so headers are added automatically on commit — developers don't need to think about it. Adapted from Safe-Synthesizer's copyright enforcement.

tools/codestyle/format.sh
binaryaaron: The scripts are the single interface between three invocation contexts:

1. **Makefile**: `uv run tools/codestyle/format.sh`
2. **Pre-commit**: `uv run tools/codestyle/format.sh` (same script, with filenames passed)
3. **CI**: same scripts called from workflow steps

The value is consistency — if we change how ruff is invoked (add flags, change file targeting, handle `--check` vs fix), we change one script, not three places. The `_lib.sh` layer provides `--check` mode toggle, file collection from args or git, and tool-availability checking.

That said, I've simplified `_lib.sh` in this update — removed the version-checking machinery that depended on yq/tools.yaml (now gone). What remains is the core: `require_tool` (PATH check) and `collect_py_files` (arg parsing + git file discovery).

tools/diff-lockfile.py
binaryaaron: Fair — this was included by mistake. With Anonymizer's small dependency surface, `git diff uv.lock` is sufficient.

**Change**: Removed `tools/diff-lockfile.py`.

.pre-commit-config.yaml
binaryaaron: They're complementary — different layers of the same enforcement:

- **Pre-commit hook** (`check-signoff`): catches missing `Signed-off-by:` *before push*. Saves a round-trip to CI. Only runs for developers who have pre-commit installed.
- **DCO Assistant bot**: CI-side enforcement that validates signatures on every PR. Catches external contributors, web-UI commits, and anyone who doesn't run pre-commit.

The hook's grep pattern matches the same `Signed-off-by: Name <email>` format the bot validates. No conflict — the hook just makes the feedback loop faster for local development.

ruff.toml
binaryaaron: Ruff's docs recommend standalone `ruff.toml` for projects that want automatic editor discovery — VS Code / Cursor with the Ruff extension find `ruff.toml` at the repo root without needing `pyproject.toml` parsing. It's the same configuration, just in its own file.

That said, it's a mild structural preference, not a hard requirement. Functionally identical to `[tool.ruff]` in `pyproject.toml`. If you'd prefer to keep everything in `pyproject.toml` for fewer files, happy to move it back. Low stakes either way.

pyproject.toml
```diff
@@ -17,6 +17,17 @@ dev = [
     "pytest>=8.0,<9",
     "pytest-cov>=7.0,<8",
     "ruff>=0.12.3,<1",
+    "ty>=0.0.23,<0.1",
+]
+docs = [
```
mckornfield, line 22: is the docs stuff configured in gh pages too?

mckornfield (approved):

tools/codestyle/format.sh
lipikaramaswamy: Should CI also call the same scripts? Right now CI still invokes Ruff directly, so we still have multiple enforcement paths

.pre-commit-config.yaml
```diff
@@ -13,10 +14,41 @@ repos:
...
-        args: [--fix]
-      - id: ruff-format
+      - id: format
+        name: Format
+        language: system
+        entry: uv run tools/codestyle/format.sh
+        types: [python]
+        pass_filenames: true
+
+      - id: typecheck
```
lipikaramaswamy: Since ty is still advisory/non-blocking, I think advisory CI is easier to justify than adding it to the default local pre-commit path right now. Should we drop the pre-commit hook for now and keep `make typecheck` plus the non-blocking CI job until we are ready to enforce?

Makefile
```diff
@@ -3,25 +3,41 @@ help:
...
-	@echo "  lint-fix               - Lint with ruff and apply fixes"
-	@echo "  check-all              - Run all checks (format-check + lint)"
-	@echo "  check-all-fix          - Run all checks with autofix (format + lint-fix)"
+	@echo "  build-wheel            - Build wheel (version from git tag)"
+	@echo "  publish-pypi           - Publish wheel to PyPI"
+	@echo ""
+	@echo "  install-dev-docs       - Install dev + docs dependencies"
+	@echo "  docs-serve             - Start docs dev server (live-reload)"
+	@echo "  docs-build             - Build docs site (strict mode)"
+	@echo ""
```
lipikaramaswamy, line 28:
```suggestion
```
it's duplicated!

.pre-commit-config.yaml
mckornfield: I'll comment it out. I think because of the `|| true` you'll never see it in pre-commit anyhow

Makefile
mckornfield: yep removin'

Makefile
```diff
@@ -42,34 +58,77 @@ install-pre-commit:
...
+docs-build:
+	@echo "Building docs site..."
+	uv run --group docs mkdocs build --strict
 
-lint-fix:
-	@echo "Fixing linting issues with ruff..."
-	uv run ruff check --fix .
+install-dev-docs:
+	@echo "Installing dev + docs dependencies..."
+	uv sync --group dev --group docs
```
lipikaramaswamy, line 123:
```suggestion
```

Makefile
```diff
@@ -42,34 +58,77 @@ install-pre-commit:
...
 
-check-all: format-check lint
+docs-serve:
+	@echo "Starting docs dev server (live-reload)..."
+	uv run --group docs mkdocs serve
```
lipikaramaswamy, line 127:
```suggestion
```

Makefile
```diff
@@ -42,34 +58,77 @@ install-pre-commit:
...
 
-check-all-fix: format lint-fix
+docs-build:
+	@echo "Building docs site..."
+	uv run --group docs mkdocs build --strict
```
lipikaramaswamy, line 131:
```suggestion
```

lipikaramaswamy:
re docs: temporary scaffolding looks fine, and I'll follow up with the actual Anonymizer-specific content. We do have some small dups cleanup to do tho

pyproject.toml
lipikaramaswamy: I don’t think this PR adds gh pages deployment wiring. `.github/workflows/ci.yml` has test/lint/typecheck jobs, but I’m not seeing a pages deploy job. I guess we can add that once the docs are actually ready

mckornfield: it's in another PR that uses this branch as a target, but I think https://github.com/NVIDIA-NeMo/Anonymizer/pull/54 has it. I was just going to try to merge this with what it has, then rebase that one

mckornfield:
believe I pushed, want to take a final look?

lipikaramaswamy (approved):

PR #57, opened 2026-03-17
fix: enforce entity_labels scope on augmented entities in detection output

### LINKED ISSUE #56: fix: enforce entity_labels scope on augmented entities in detection output
### Priority Level

Medium

## Problem

When a user provides an explicit `entity_labels` list via `Detect(entity_labels=[...])`, the LLM augmentation step can still introduce entities with labels outside that list. These leak into `COL_DETECTED_ENTITIES`, `COL_TAGGED_TEXT`, `COL_ENTITIES_BY_VALUE`, and downstream into replacement.

This is a regression from #44, which removed the replace-side `_filter_entities_by_label` safety net under the assumption that scoping detection via `_resolve_detection_labels` was sufficient. However, the augment prompt explicitly allows the LLM to create novel labels ("If no known label fits, create a concise snake_case label"), and neither the `AugmentedEntitySchema` nor `apply_augmented_entities` constrains the output.

## Proposed Fix

Two layers of enforcement:

1. **Prompt (soft):** Make `_get_augment_prompt` strict when `entity_labels` is explicitly provided ("Use ONLY labels from this list"), permissive when using defaults.
2. **Post-processing (hard):** Add `_filter_to_allowed_labels` in `detect_and_validate_entities` that drops entities with out-of-scope labels and rebuilds derived columns. Only applied when `entity_labels is not None`. (OR find the best place to put this) 

## Acceptance Criteria

- Augmented entities with labels not in the user's `entity_labels` list are excluded from detection output
- `entity_labels=None` (defaults) preserves current behavior — augmenter can still create novel labels
- Regression tests cover both paths

_No response_

### Dependencies

_No response_

lipikaramaswamy:
## Summary
This PR addresses a regression in behavior that was inadvertently introduced by #44, and fortifies the implementation.

When a user provides an explicit `Detect(entity_labels=[...])`, the LLM augmentation step should not introduce entities with labels outside that list. To enforce this, we add two layers:
1. **Prompt (soft enforcement):** `_get_augment_prompt` accepts a `strict_labels` flag — strict ("Use ONLY labels from this list") when `entity_labels` is explicitly provided, permissive when using defaults.
2. **Post-processing (hard enforcement):** `_materialize_final_entities` filters `COL_FINAL_ENTITIES` to only include entities whose label is in the user's list. This runs during the existing `.apply()` pass in `run()` — no extra loop. `COL_DETECTED_ENTITIES` retains the full unfiltered trace for debugging. Only applied when `entity_labels is not None`; defaults pass through unfiltered.

**Note:** When `entity_labels` is explicitly provided — even if it matches `DEFAULT_ENTITY_LABELS` — strict mode and filtering apply. `entity_labels=None` is the only way to get permissive augmentation. This is tracked for documentation in a follow-up (`TODO(docs)` in code).

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing
- [x] Tests pass locally
- [x] Notebooks run end-to-end with expected behavior
image
image
image

## Related Issues 
Closes #56

### DIFF
--- a/src/anonymizer/engine/detection/detection_workflow.py
+++ b/src/anonymizer/engine/detection/detection_workflow.py
@@ -45,10 +45,12 @@ from anonymizer.engine.detection.custom_columns import (
     parse_detected_entities,
     prepare_validation_inputs,
 )
+from anonymizer.engine.detection.postprocess import EntitySpan, group_entities_by_value
 from anonymizer.engine.ndd.adapter import FailedRecord, NddAdapter
 from anonymizer.engine.ndd.model_loader import resolve_model_alias
 from anonymizer.engine.schemas import (
     AugmentedEntitiesSchema,
+    EntitiesByValueSchema,
     EntitiesSchema,
     LatentEntitiesSchema,
     ValidationDecisionsSchema,
@@ -78,7 +80,6 @@ class EntityDetectionWorkflow:
         gliner_detection_threshold: float,
         entity_labels: list[str] | None = None,
         data_summary: str | None = None,
-        compute_grouped: bool = True,
         preview_num_records: int | None = None,
     ) -> EntityDetectionResult:
         """Run the core detection pipeline: GLiNER NER, LLM validation, LLM augmentation, and finalization.
@@ -162,8 +163,6 @@ class EntityDetectionWorkflow:
             preview_num_records=preview_num_records,
         )
         detected_df = detection_result.dataframe.copy()
-        if not compute_grouped and COL_ENTITIES_BY_VALUE in detected_df.columns:
-            detected_df = detected_df.drop(columns=[COL_ENTITIES_BY_VALUE])
         return EntityDetectionResult(dataframe=detected_df, failed_records=detection_result.failed_records)
 
     def identify_latent_entities(
@@ -241,7 +240,6 @@ class EntityDetectionWorkflow:
             gliner_detection_threshold=gliner_detection_threshold,
             entity_labels=entity_labels,
             data_summary=data_summary,
-            compute_grouped=compute_grouped,
             preview_num_records=preview_num_records,
         )
 
@@ -271,6 +269,8 @@ class EntityDetectionWorkflow:
             final_df[COL_FINAL_ENTITIES] = final_df.apply(
                 lambda row: _materialize_final_entities(row, allowed_labels=allowed), axis=1
             )
+            if compute_grouped:
+                final_df[COL_ENTITIES_BY_VALUE] = final_df[COL_FINAL_ENTITIES].apply(_build_entities_by_value)
         if "original_text_column" in dataframe.attrs:
             final_df.attrs["original_text_column"] = dataframe.attrs["original_text_column"]
         return EntityDetectionResult(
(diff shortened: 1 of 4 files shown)


asteier2026 (approved):

src/anonymizer/engine/detection/detection_workflow.py
```diff
@@ -260,9 +262,14 @@ def run(
...
 
+        # When entity_labels is explicitly provided (even if it matches DEFAULT_ENTITY_LABELS),
+        # the augmenter is strict and out-of-scope labels are filtered.
+        # entity_labels=None is the only way to get permissive augmentation.
+        # TODO(docs): document this None-vs-explicit contract in user-facing docs.
         if COL_DETECTED_ENTITIES in final_df.columns:
-            final_df[COL_FINAL_ENTITIES] = final_df[COL_DETECTED_ENTITIES].apply(
-                lambda x: EntitiesSchema.from_raw(x).model_dump()
+            allowed = set(entity_labels) if entity_labels is not None else None
+            final_df[COL_FINAL_ENTITIES] = final_df.apply(
```
andreatnvidia: I think this still leaves one leak in the `Substitute` path. `final_entities` gets filtered here, but `_entities_by_value` is still built from the unfiltered detection trace, and `LlmReplaceWorkflow.generate_map_only()` reads that grouped column later. I repro'd it with a `server_name` example: the replaced text was scoped correctly, but `_replacement_map` still contained `server_name`. maybe worth filtering or rebuilding `_entities_by_value` from the same allowed set here?

src/anonymizer/engine/detection/detection_workflow.py
```diff
@@ -409,18 +425,28 @@ def _get_validation_prompt(*, data_summary: str | None, labels: list[str]) -> st
     return prompt.replace("<<DATA_SUMMARY>>", context_section)
 
 
-def _get_augment_prompt(*, data_summary: str | None, labels: list[str]) -> str:
+def _get_augment_prompt(*, data_summary: str | None, labels: list[str], strict_labels: bool = False) -> str:
+    if strict_labels:
+        label_block = (
+            "Here are the allowed entity classes. Use ONLY labels from this list:\n"
```
andreatnvidia, line 457: tiny prompt nit: in strict mode we now say "use only labels from this list", but the example lower in this prompt still invents `first_name` / `last_name`. examples tend to steer the model as much as the prose does, so maybe switch to a strict-specific example here so the soft enforcement stays aligned with the hard filter?

src/anonymizer/engine/detection/detection_workflow.py
lipikaramaswamy: Valid leak! Instead of patching `_entities_by_value` with a separate filter, I moved its construction out of `apply_validation_and_finalize` entirely. `COL_ENTITIES_BY_VALUE` is now always derived from `COL_FINAL_ENTITIES` in `run()` via `_build_entities_by_value` — so it's guaranteed to match the filtered set. The custom column no longer produces it. This also means the `compute_grouped` flag moved from `detect_and_validate_entities` to `run()`, since that's where the column is now built.

src/anonymizer/engine/detection/detection_workflow.py
lipikaramaswamy: Ok, I extended this if else block to also diverge on how we present the example. @asteier2026 could you review?

asteier2026 (approved):

andreatnvidia (approved):
Approving - couple of small follow-ups:

- `_materialize_final_entities` only reads `row[COL_DETECTED_ENTITIES]`, so the `axis=1` apply could just be a column-wise apply on that column directly - faster and simpler.
- Same function: `EntitiesSchema(entities=[e.model_dump() for e in kept]).model_dump()` round-trips through dicts unnecessarily. `kept` already holds pydantic instances so `EntitiesSchema(entities=kept).model_dump()` works.

lipikaramaswamy:
> Approving - couple of small follow-ups:
> 
> * `_materialize_final_entities` only reads `row[COL_DETECTED_ENTITIES]`, so the `axis=1` apply could just be a column-wise apply on that column directly - faster and simpler.
> * Same function: `EntitiesSchema(entities=[e.model_dump() for e in kept]).model_dump()` round-trips through dicts unnecessarily. `kept` already holds pydantic instances so `EntitiesSchema(entities=kept).model_dump()` works.

Addressed, thanks!

PR #61, opened 2026-03-20
feat: add evaluate and repair workflows

### LINKED ISSUE #34: feat(rewrite): engine — evaluate & repair
Part of #30. Most complex engine step.

## Scope

**`engine/rewrite/evaluate_repair.py`** — `EvaluateRepairWorkflow`:

**Repair loop** (up to `EvaluationCriteria.max_repair_iterations`):

1. `_evaluate(df)` — 2 NddAdapter calls:
   - Quality answering: score each `COL_QUALITY_QA` item on `COL_REWRITTEN_TEXT` → `COL_QUALITY_ANSWERS` (`evaluator` alias)
   - Privacy answering: answer each `COL_PRIVACY_QA` item ("Can X be deduced?") → `COL_PRIVACY_ANSWERS` (`evaluator` alias)

2. `_maybe_reanswer(df)` — if >50% of privacy answers for a row are `unknown`, re-run with `disposition_analyzer` alias → `COL_PRIVACY_ANSWERS_REANSWER`

3. `_compute_metrics(df)` — pure Python:
   - `COL_UTILITY_SCORE`: mean of quality answer scores
   - `COL_LEAKAGE_MASS`: weighted sum of leaked entities using `EvaluationCriteria.sensitivity_weights`
   - `COL_ANY_HIGH_LEAKED`: True if any HIGH-sensitivity entity leaked

4. `_determine_repair_needs(df, evaluation)` — pure Python boolean mask:
   - `COL_ANY_HIGH_LEAKED=True` AND `evaluation.repair_any_high_leak=True`, OR
   - `COL_LEAKAGE_MASS >= evaluation.get_effective_threshold(domain)`

5. If any rows need repair: `_repair(failing_rows)` — 1 NddAdapter call with targeted repair prompt listing specific leaked entities → new `COL_REWRITTEN_TEXT` (`evaluator` alias)

6. Merge repaired + passed; increment iteration

**After loop** — `_flag_for_review(df, evaluation)`:
- `COL_NEEDS_HUMAN_REVIEW=True` if utility low, leakage high, or any HIGH leak remains
- `COL_REPAIR_ITERATIONS` set

**Tests**: `tests/engine/test_evaluate_repair.py`

## Acceptance criteria
- [ ] `_compute_metrics` correctly computes leakage_mass with sensitivity weights
- [ ] `_determine_repair_needs` respects all EvaluationCriteria combinations
- [ ] Repair loop runs at most `max_repair_iterations` times
- [ ] Reanswer NddAdapter call made only when >50% unknowns
- [ ] `needs_human_review=True` when any flag condition is met
- [ ] Failed records accumulated across all NddAdapter calls in loop

lipikaramaswamy:
## Summary

Implements Step 5 of the rewrite pipeline: evaluate rewritten text for quality and privacy, then repair if privacy leaks are detected.

**`EvaluateWorkflow`** (`engine/rewrite/evaluate.py`) — 6-column NDD pipeline: LLM quality re-answer, LLM privacy re-answer, QA pairing, LLM quality comparison, metric computation (utility_score, leakage_mass, any_high_leaked), and repair-needs determination.

**`RepairWorkflow`** (`engine/rewrite/repair.py`) — 2-column NDD pipeline: leaked-items formatting and LLM repair via `PydanticResponseRecipe` inside a custom column generator with dynamically-resolved model alias.

**Schema changes**: `PrivacyAnswer` enum enforces yes/no only — privacy questions are binary ("can entity X be deduced?"), so "unknown" is a hedge that doesn't serve the pipeline; `format_for_rewrite_context()` now uses `protected_entities` instead of `medium_and_high_sensitivity_entities` — the rewriter and evaluator must operate on the same entity set, otherwise LOW-sensitivity entities flagged for protection (due to combinatorial re-identification risk) would be evaluated but never shown to the rewriter.

**Design decision** - separate workflows instead of one combined pipeline, so the orchestrator can run evaluate on all rows, filter to failing rows, run repair on just those, and re-evaluate only repaired rows (no wasted LLM calls on passing rows)

Prompts ported from the research repo with XML section tags, `<<PLACEHOLDER>>` + `.replace()` convention, and schema field names validated at import time via `_field()`.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing
- [x] Tests pass locally

## Related Issues
Closes #34

### DIFF
--- a/src/anonymizer/engine/rewrite/evaluate.py
+++ b/src/anonymizer/engine/rewrite/evaluate.py
@@ -3,11 +3,13 @@
 
 from __future__ import annotations
 
+import json
 from typing import Any
 
 from data_designer.config import custom_column_generator
-from data_designer.config.column_configs import CustomColumnConfig, LLMStructuredColumnConfig
+from data_designer.config.column_configs import CustomColumnConfig
 from data_designer.config.column_types import ColumnConfigT
+from data_designer.engine.models.recipes.response_recipes import PydanticResponseRecipe
 from pydantic import BaseModel
 
 from anonymizer.config.models import RewriteModelSelection
@@ -15,6 +17,7 @@ from anonymizer.config.rewrite import EvaluationCriteria
 from anonymizer.engine.constants import (
     COL_ANY_HIGH_LEAKED,
     COL_LEAKAGE_MASS,
+    COL_NEEDS_REPAIR,
     COL_PRIVACY_QA,
     COL_PRIVACY_QA_REANSWER,
     COL_QUALITY_QA,
@@ -25,37 +28,22 @@ from anonymizer.engine.constants import (
 )
 from anonymizer.engine.ndd.adapter import NddAdapter
 from anonymizer.engine.ndd.model_loader import resolve_model_alias
+from anonymizer.engine.rewrite.parsers import (
+    parse_privacy_answers,
+    parse_privacy_qa,
+    parse_quality_answers,
+    parse_quality_compare,
+    parse_quality_qa,
+)
 from anonymizer.engine.schemas.rewrite import (
     PrivacyAnswer,
     PrivacyAnswerItemSchema,
     PrivacyAnswersSchema,
     PrivacyQAPairsSchema,
     QACompareResultsSchema,
-    QualityAnswerSchema,
     QualityAnswersSchema,
-    QualityQAItemSchema,
-    QualityQAPairsSchema,
     SensitivityLevel,
 )
-
-_COL_NEEDS_REPAIR = "_needs_repair"
-_COL_QA_COMPARISON_BLOCK = "_qa_comparison_block"
-
-
-# Schema field names validated at import time.
-def _field(model: type, name: str) -> str:
-    if name not in model.model_fields:
-        raise KeyError(f"{model.__name__} has no field '{name}'")
-    return name
-
-
-_F_ITEMS = _field(QualityQAPairsSchema, "items")
-_F_ANSWERS = _field(QualityAnswersSchema, "answers")
-_F_ID = _field(QualityQAItemSchema, "id")
-_F_QUESTION = _field(QualityQAItemSchema, "question")
-_F_ANSWER = _field(QualityAnswerSchema, "answer")
-_F_REFERENCE_ANSWER = _field(QualityQAItemSchema, "reference_answer")
-
 
 # ---------------------------------------------------------------------------
 # Generator params
@@ -73,63 +61,85 @@ class RepairNeedsParams(BaseModel):
 
 
 # ---------------------------------------------------------------------------
-# Prompts (Jinja2 templates resolved by NDD)
+# Python-rendered prompts
 # ---------------------------------------------------------------------------
 
 
-def _quality_reanswer_prompt() -> str:
+def _render_quality_reanswer_prompt(row: dict[str, Any]) -> str:
+    qa = parse_quality_qa(row.get(COL_QUALITY_QA, {}))
+    skeleton = [{"id": item.id, "question": item.question, "answer": ""} for item in qa.items]
+
     prompt = """You are taking a reading comprehension exam. You will answer each question about the text.
 
 <rules>
 - If the text does not state the answer, use "unknown"
 - Keep answers concise and factual
 - Do not invent details
+- You MUST provide an answer for EVERY item in the template below
 </rules>
 
 <text>
-{{ <<COL_REWRITTEN_TEXT>> }}
+<<REWRITTEN_TEXT>>
 </text>
 
-<questions>
-{% for it in <<COL_QUALITY_QA>>['<<F_ITEMS>>'] %}- {{ it['<<F_ID>>'] }}: {{ it['<<F_QUESTION>>'] }}
-{% endfor %}</questions>
+<task>
+Fill in the "answer" field for each item. Do not add or remove items.
+</task>
+<answer_template>
+<<SKELETON>>
+</answer_template>
 """
-    return (
-        prompt.replace("<<COL_REWRITTEN_TEXT>>", COL_REWRITTEN_TEXT)
-        .replace("<<COL_QUALITY_QA>>", COL_QUALITY_QA)
-        .replace("<<F_ITEMS>>", _F_ITEMS)
-        .replace("<<F_ID>>", _F_ID)
-        .replace("<<F_QUESTION>>", _F_QUESTION)
+    return prompt.replace("<<REWRITTEN_TEXT>>", str(row.get(COL_REWRITTEN_TEXT, ""))).replace(
+        "<<SKELETON>>", json.dumps({"answers": skeleton}, indent=2)
     )
 
 
-def _privacy_reanswer_prompt() -> str:
+def _render_privacy_reanswer_prompt(row: dict[str, Any]) -> str:
+    qa = parse_privacy_qa(row.get(COL_PRIVACY_QA))
+    skeleton = [{"id": item.id, "question": item.question, "answer": ""} for item in qa.items]
+
     prompt = """You are a privacy auditor. Read the text and answer each question with "yes" or "no".
 
 <rules>
-- Answer "yes" if the specific entity value can be identified or reasonably inferred from the text.
-- Answer "no" if it cannot be answered from the text.
+- Answer "yes" if the specific entity value can be identified or reasonably inferred from the text
+- Answer "no" if it cannot
 - You must commit to "yes" or "no". Do not hedge.
+- You MUST provide an answer for EVERY item in the template below
 </rules>
 
 <text>
-{{ <<COL_REWRITTEN_TEXT>> }}
+<<REWRITTEN_TEXT>>
 </text>
 
-<questions>
-{% for it in <<COL_PRIVACY_QA>>['<<F_ITEMS>>'] %}- {{ it['<<F_ID>>'] }}: {{ it['<<F_QUESTION>>'] }}
-{% endfor %}</questions>
+<task>
+Fill in the "answer" field for each item with "yes" or "no". Do not add or remove items.
+</task>
+<answer_template>
+<<SKELETON>>
+</answer_template>
 """
-    return (
-        prompt.replace("<<COL_REWRITTEN_TEXT>>", COL_REWRITTEN_TEXT)
-        .replace("<<COL_PRIVACY_QA>>", COL_PRIVACY_QA)
-        .replace("<<F_ITEMS>>", _F_ITEMS)
-        .replace("<<F_ID>>", _F_ID)
-        .replace("<<F_QUESTION>>", _F_QUESTION)
+    return prompt.replace("<<REWRITTEN_TEXT>>", str(row.get(COL_REWRITTEN_TEXT, ""))).replace(
+        "<<SKELETON>>", json.dumps({"answers": skeleton}, indent=2)
     )
 
 
-def _quality_compare_prompt() -> str:
+def _render_quality_compare_prompt(row: dict[str, Any]) -> str:
+    qa = parse_quality_qa(row.get(COL_QUALITY_QA, {}))
+    reanswer = parse_quality_answers(row.get(COL_QUALITY_QA_REANSWER))
+    answer_lookup: dict[int, str] = {a.id: a.answer for a in reanswer}
+
+    skeleton = [
+        {
+            "id": item.id,
+            "question": item.question,
+            "expected_answer": item.reference_answer,
+            "student_answer": answer_lookup.get(item.id, "(no answer)"),
+            "score": "",
+            "reason": "",
+        }
+        for item in qa.items
+    ]
+
     prompt = """A student took a reading comprehension exam.
 
 <scoring_rubric>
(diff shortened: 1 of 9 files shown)


src/anonymizer/engine/rewrite/evaluate.py
```diff
@@ -0,0 +1,387 @@
...
+# Prompts (Jinja2 templates resolved by NDD)
+# ---------------------------------------------------------------------------
+
+
+def _quality_reanswer_prompt() -> str:
+    prompt = """You are taking a reading comprehension exam. You will answer each question about the text.
+
+<rules>
+- If the text does not state the answer, use "unknown"
+- Keep answers concise and factual
```
asteier2026, line 77: I thought we didn't want to hedge with "unknown" only yes or no?

asteier2026 (commented):
Only one small comment in evaluate, in the re-answer prompt line 84

asteier2026 (approved):

src/anonymizer/engine/rewrite/evaluate.py
```diff
@@ -0,0 +1,387 @@
...
+    """Weighted sum of leaked entities.
+
+    An entity is "leaked" when the privacy re-answer is ``yes``.
+    The weight comes from ``sensitivity_weights[entity.sensitivity]``.
+    """
+    qa_lookup: dict[int, str] = {item.id: item.sensitivity for item in privacy_qa.items}
+    total = 0.0
+    for answer in privacy_answers:
+        if answer.answer == PrivacyAnswer.yes:
+            sensitivity = qa_lookup.get(answer.id, SensitivityLevel.low)
```
andreatnvidia: when an answer ID isn't found in qa_lookup, this falls back to SensitivityLevel.low - which means a hallucinated answer gets silently counted with 0.3 weight. the flip side is worse though: if the LLM *skips* a HIGH-sensitivity answer entirely, it just doesn't show up in the sum at all, so a leaking row can pass evaluation. might be better to validate that the returned answer IDs match the QA IDs before computing metrics, and treat mismatches as a failed evaluation that forces repair?

src/anonymizer/engine/rewrite/evaluate.py
```diff
@@ -0,0 +1,387 @@
...
+        return PrivacyQAPairsSchema.model_validate(raw)
+    raise TypeError(f"Expected PrivacyQAPairsSchema or dict, got {type(raw).__name__}")
+
+
+# ---------------------------------------------------------------------------
+# Pure-Python metric helpers (public, testable)
+# ---------------------------------------------------------------------------
+
+
+def compute_utility_score(compare_scores: list[float]) -> float:
```
andreatnvidia, line 237: suggestion: `compute_utility_score` averages whatever scores it gets, but if the compare model drops one of the QA items, the missing item just disappears from the denominator and the average can go up. could we pad missing IDs with 0.0, or validate coverage against the original quality QA IDs?

src/anonymizer/engine/rewrite/evaluate.py
```diff
@@ -0,0 +1,387 @@
...
+    PrivacyQAPairsSchema,
+    QACompareResultsSchema,
+    QualityAnswerSchema,
+    QualityAnswersSchema,
+    QualityQAItemSchema,
+    QualityQAPairsSchema,
+    SensitivityLevel,
+)
+
+_COL_NEEDS_REPAIR = "_needs_repair"
```
andreatnvidia: this will need to be readable by the orchestrator for row filtering - might want to move it to constants.py as COL_NEEDS_REPAIR alongside COL_REPAIR_ITERATIONS?

tests/engine/test_repair.py
```diff
@@ -0,0 +1,36 @@
...
+from anonymizer.engine.constants import COL_REWRITTEN_TEXT
+from anonymizer.engine.rewrite.repair import RepairWorkflow
+
+_STUB_PRIVACY_GOAL = PrivacyGoal(
+    protect="Direct identifiers, quasi-identifier combinations, and latent inferences",
+    preserve="General utility, content quality, and semantic meaning of the original text",
+)
+
+
+def test_repair_columns_pipeline(
```
andreatnvidia, line 190: nit: the repair module has some non-trivial logic in `_leaked_items_text`, `_format_protection_block`, and `_render_repair_prompt` that would benefit from unit tests - the evaluate side has great coverage for its equivalent functions, wdyt about matching that here?

src/anonymizer/engine/rewrite/repair.py
```diff
@@ -0,0 +1,269 @@
...
+    RewriteOutputSchema,
+    SensitivityDispositionSchema,
+)
+
+logger = logging.getLogger("anonymizer.rewrite.repair")
+
+_COL_LEAKED_PRIVACY_ITEMS = "_leaked_privacy_items"
+
+# Schema field names validated at import time.
+_F_NEEDS_PROTECTION = EntityDispositionSchema.model_fields["needs_protection"] and "needs_protection"
```
andreatnvidia: nit: the `and` trick for field validation works but the error message is less clear than evaluate.py's `_field()` helper - might be nice to use the same pattern in both modules for consistency

andreatnvidia (changes requested):
The workflow structure is solid and the metric helpers are clean and well-tested. Two concerns on the evaluation logic:

1. `compute_leakage_mass` and `compute_any_high_leaked` don't validate that the returned answer IDs match the QA IDs. If the evaluator drops a HIGH-sensitivity answer, the leak isn't counted and the row can skip repair. For a privacy gate, this should fail closed.

2. `compute_utility_score` averages whatever compare scores it gets without checking coverage. A partial compare response inflates the score.

Also flagged: `_COL_NEEDS_REPAIR` should probably move to constants.py before the orchestrator PR, and the repair module could use more test coverage.

andreatnvidia:
> `repair.py:69-74` and `evaluate.py:163-168`
>
> ```python
> def _parse_privacy_answers(raw: Any) -> list[PrivacyAnswerItemSchema]:
>     if isinstance(raw, PrivacyAnswersSchema):
>         return raw.answers
>     if isinstance(raw, dict):
>         return PrivacyAnswersSchema.model_validate(raw).answers
>     raise TypeError(...)
> ```

`_parse_privacy_answers` and `_parse_privacy_qa` are identical between evaluate and repair - maybe extract to a shared module or add classmethods on the schemas?

src/anonymizer/engine/rewrite/evaluate.py
lipikaramaswamy: This is a good point and raises a bigger question of - why accept a missing id?

So I changed the approach to use custom columns to introduce row-specific context-validated parsers that enforce complete ID coverage per row via Pydantic `model_validate(..., context={"expected_ids": [...]})`. DD retries if LLM skips IDs.

src/anonymizer/engine/rewrite/evaluate.py
lipikaramaswamy: Same as `compute_leakage_mass`, introduced a custom column to prevent missing IDs.

tests/engine/test_repair.py
lipikaramaswamy: Added unit tests for `_leaked_items_text`, `_format_protection_block`, and `_render_repair_prompt`.

src/anonymizer/engine/rewrite/evaluate.py
lipikaramaswamy: all moved to `constants.py`

src/anonymizer/engine/rewrite/repair.py
lipikaramaswamy: using a shared function now.

lipikaramaswamy:
> > `repair.py:69-74` and `evaluate.py:163-168`
> > ```python
> > def _parse_privacy_answers(raw: Any) -> list[PrivacyAnswerItemSchema]:
> >     if isinstance(raw, PrivacyAnswersSchema):
> >         return raw.answers
> >     if isinstance(raw, dict):
> >         return PrivacyAnswersSchema.model_validate(raw).answers
> >     raise TypeError(...)
> > ```
> 
> `_parse_privacy_answers` and `_parse_privacy_qa` are identical between evaluate and repair - maybe extract to a shared module or add classmethods on the schemas?

Extracted to `rewrite/parsers.py`. The shared parse helpers (`parse_privacy_answers`, `parse_privacy_qa`, `parse_quality_qa`, `parse_quality_answers`, `parse_quality_compare`, `parse_sensitivity_disposition`) and the `field()` validator live there. Both `evaluate.py` and `repair.py` import from it.

src/anonymizer/engine/rewrite/evaluate.py
```diff
@@ -0,0 +1,444 @@
...
+    auto_repair_privacy: bool,
+    repair_any_high_leak: bool,
+    effective_threshold: float,
+) -> bool:
+    """Decide whether a single row needs repair."""
+    if not auto_repair_privacy:
+        return False
+    if repair_any_high_leak and any_high_leaked:
+        return True
+    return leakage_mass >= effective_threshold
```
andreatnvidia: I think `max_leakage_mass=0.0` is broken with the current `>=` check. a row with `leakage_mass == 0.0` still gets marked for repair, so the strictest valid setting can never pass.

src/anonymizer/engine/schemas/rewrite.py
```diff
@@ -312,21 +312,49 @@ class QualityAnswerSchema(BaseModel):
...
-    """LLM output schema for quality QA re-answer step (on rewritten text)."""
+    """LLM output schema for quality QA re-answer step (on rewritten text).
+
+    When validated with ``context={"expected_ids": [1, 2, ...]}``,
+    raises if any expected ID is missing from the answers.
+    """
 
     answers: list[QualityAnswerSchema]
 
+    @model_validator(mode="after")
```
andreatnvidia, line 341: the new context validator fixes missing IDs, but it still lets duplicate and extra IDs through. that means something like `[1, 1, 2]` passes validation and `compute_leakage_mass()` counts the duplicated HIGH leak twice. might be worth rejecting duplicates and extras here too, then the metric helpers can keep assuming one answer per question.

src/anonymizer/engine/rewrite/evaluate.py
```diff
@@ -0,0 +1,444 @@
...
+</text>
+
+<task>
+Fill in the "answer" field for each item. Do not add or remove items.
+</task>
+<answer_template>
+<<SKELETON>>
+</answer_template>
+"""
+    return prompt.replace("<<REWRITTEN_TEXT>>", str(row.get(COL_REWRITTEN_TEXT, ""))).replace(
```
andreatnvidia: nit: if the rewritten text ever contains the literal string `<<SKELETON>>`, the second `.replace()` would also substitute inside the text block. same risk in the repair prompt with 10 chained replacements. a single-pass approach or reversed replacement order would be safer.

src/anonymizer/engine/rewrite/repair.py
```diff
@@ -0,0 +1,241 @@
...
+    privacy_qa = parse_privacy_qa(row.get(COL_PRIVACY_QA))
+    row[_COL_LEAKED_PRIVACY_ITEMS] = _leaked_items_text(privacy_answers, privacy_qa)
+    return row
+
+
+def _make_repair_column(repairer_alias: str) -> Any:
+    """Factory that creates a repair column generator bound to a resolved model alias."""
+
+    @custom_column_generator(
+        required_columns=[_COL_LEAKED_PRIVACY_ITEMS, COL_REWRITTEN_TEXT],
```
andreatnvidia: `required_columns` only lists `_COL_LEAKED_PRIVACY_ITEMS` and `COL_REWRITTEN_TEXT`, but `_render_repair_prompt` reads from `COL_SENSITIVITY_DISPOSITION`, `COL_TEXT`, `COL_REPLACEMENT_MAP_FOR_PROMPT`, `COL_LEAKAGE_MASS`, `COL_ANY_HIGH_LEAKED`, and `COL_UTILITY_SCORE` too. if DD uses this for dependency ordering, the generator could run before those columns are populated.

andreatnvidia:
nice progress on this PR. one thing i’d consider as follow-up, borrowing a bit from DD’s patterns, is tightening the workflow contract around the custom columns:

- make `required_columns` exhaustive for anything the generator actually reads, since DD uses that metadata for DAG ordering and missing-input checks
- keep pushing validation to the parser/schema boundary, not the metric helpers. missing IDs are fixed now, but duplicate / extra IDs would be good to reject there too
- add one tiny workflow-level preview/smoke test that runs the evaluate/repair columns together on a row, so metadata issues show up before orchestrator wiring

none of that needs to block this PR beyond the concrete bugs already called out, but it feels like the next step that would make these workflows more DD-native and harder to misuse later.

lipikaramaswamy:
> nice progress on this PR. one thing i’d consider as follow-up, borrowing a bit from DD’s patterns, is tightening the workflow contract around the custom columns:
> 
> * make `required_columns` exhaustive for anything the generator actually reads, since DD uses that metadata for DAG ordering and missing-input checks
> * keep pushing validation to the parser/schema boundary, not the metric helpers. missing IDs are fixed now, but duplicate / extra IDs would be good to reject there too
> * add one tiny workflow-level preview/smoke test that runs the evaluate/repair columns together on a row, so metadata issues show up before orchestrator wiring
> 
> none of that needs to block this PR beyond the concrete bugs already called out, but it feels like the next step that would make these workflows more DD-native and harder to misuse later.

First two done: `required_columns` is now exhaustive across all 7 generators (audited), and duplicate/extra IDs are rejected at the schema boundary via `_validate_id_coverage()`. The workflow-level smoke test we'll pick up in the orchestrator PR.

src/anonymizer/engine/schemas/rewrite.py
lipikaramaswamy: Added `_validate_id_coverage()` shared helper in the schema module — all three answer schemas now reject missing, duplicate, and extra IDs when validated with context. Tests added for duplicate and extra cases.

src/anonymizer/engine/rewrite/evaluate.py
lipikaramaswamy: Fixed

src/anonymizer/engine/rewrite/evaluate.py
lipikaramaswamy: Reversed the replacement order so skeleton (safe JSON) is substituted first, then user text last.

src/anonymizer/engine/rewrite/repair.py
lipikaramaswamy: Fixed — `_repair_column` now declares all 8 columns that `_render_repair_prompt` reads in `required_columns`. Also checked all other custom columns to be sure.

andreatnvidia (approved):
Approving. The blocking issues I called out are addressed on the current head.

One follow-up to keep in mind: the placeholder-collision fix looks complete in the evaluate prompts, but `_render_repair_prompt()` still appears to have the same sequential `.replace()` risk if user text contains one of the later placeholder tokens. I do not think that needs to hold this PR, but it would be good to clean up next.

lipikaramaswamy:
> Approving. The blocking issues I called out are addressed on the current head.
> 
> One follow-up to keep in mind: the placeholder-collision fix looks complete in the evaluate prompts, but `_render_repair_prompt()` still appears to have the same sequential `.replace()` risk if user text contains one of the later placeholder tokens. I do not think that needs to hold this PR, but it would be good to clean up next.

Decided to fix here. Added `render_template()` to `parsers.py` — single-pass regex substitution that replaces all placeholders simultaneously. Both evaluate and repair prompts now use it, so no `.replace()` chain ordering risk anywhere. Also added `test_parsers.py` with a dedicated cross-contamination test that verifies `<<SCORE>>` inside user text doesn't get substituted. (and tracking #66) 

Other changes I've made since your review:
- `.model_dump()` on all three evaluate custom column outputs (DD's `sanitize_record()` needs JSON-serializable values)
- Repair column writes to `COL_REWRITTEN_TEXT_NEXT` instead of overwriting the seed column (DD rejects generated columns with the same name as seed columns)

Issue #60, opened 2026-03-19
refactor: extract shared entity-row split/reorder/recombine helper


lipikaramaswamy:
## Context

Three workflows now implement the same pattern:
1. Add a `_anonymizer_row_order` column to preserve original ordering.
2. Partition DataFrame into entity rows vs passthrough rows.
3. Run LLM/column pipeline on entity rows only.
4. Set a default value on passthrough rows (e.g., empty replacement map, original text).
5. `pd.concat` + `sort_values("_anonymizer_row_order")` + `drop(columns=...)` + `reset_index`.
6. Propagate `df.attrs`.

**Locations**:
- `engine/replace/llm_replace_workflow.py` — `generate_map_only()`.
- `engine/rewrite/rewrite_generation.py` — `RewriteGenerationWorkflow.run()`.
- `engine/detection/detection_workflow.py` — similar split pattern.

## Proposal

Extract a shared helper, e.g.:

```python
def run_on_entity_rows(
    df: pd.DataFrame,
    *,
    has_entities_fn: Callable[[Any], bool],
    passthrefaults: dict[str, Any],
    run_fn: Callable[[pd.DataFrame], WorkflowRunResult],
) -> WorkflowRunResult:
```

Handles partitioning, ordering, recombination, and `attrs` propagation in one place.

## Acceptance criteria
- [ ] Shared helper in `engine/utils.py` or similar.
- [ ] `generate_map_only` and `RewriteGenerationWorkflow.run()` refactored to use it.
- [ ] No behavior change — existing tests pass as-is.

Issue #37, opened 2026-03-09
feat(rewrite): judge rubric refinement


lipikaramaswamy:
Follow-up to #30. To be opened after the rewrite pipeline is live and we can observe real judge output.

## Context

The initial implementation uses 3 rubrics on a 1-10 scale (privacy, quality, naturalness), ported directly from the research repo. This issue is to revisit those choices with real data.

## Scope

- Run the rewrite pipeline on representative samples and observe judge score distributions
- Evaluate whether 1-10 gives useful spread or clusters
- Decide on 3 vs 4 rubrics — the case for adding **faithfulness** (does the rewrite accurately represent the original's non-PII content, i.e. does it hallucinate?) as a fourth dimension
- Rewrite anchor descriptions to be more distinct and actionable
- Update `FinalJudgeWorkflow` prompt and rubric definitions
- Update tests

## Decision to make

| Option | Rubrics | Scale |
|---|---|---|
| A (current) | privacy, quality, naturalness | 1-10 |
| B | privacy, utility, naturalness | 1-10 with better anchors |
| C | privacy, utility, faithfulness, fluency | 1-5 |

## Acceptance criteria
- [ ] Rubric choice documented and justified with observed score distributions
- [ ] Anchor descriptions are clearly distinct at each level
- [ ] `FinalJudgeWorkflow` updated
- [ ] Tests updated

lipikaramaswamy:
PR #86 improved the judge prompt wording and iterated on scoring behavior, addressing this in the interim. 

However, the formal rubric decision (Options A/B/C), score distribution analysis, and anchor point documentation were not completed. Closing in favor of a focused follow-up: see #106. The LLM-as-a-judge extension for REPLACE mode is tracked separately in #98.

PR #64, opened 2026-03-23
feat: add final judge and RewriteWorkflow orchestrator

### LINKED ISSUE #35: feat(rewrite): engine — final judge + RewriteWorkflow

lipikaramaswamy:
> **Note:** This PR is stacked on #61 (evaluate & repair workflows). The base branch is `lipikaramaswamy/feat/rewrite-engine-eval-repair`. Please review only the diff against that branch. Once #61 merges, this will be rebased onto `main`.

## Summary

- **`FinalJudgeWorkflow`** (`engine/rewrite/final_judge.py`) -- holistic LLM evaluation using `LLMJudgeColumnConfig` with three rubrics (privacy, quality, naturalness) on a 1-10 scale, ported verbatim from the research repo. `needs_human_review` flagging based on objective metrics only (failed rewrite, utility below threshold, leakage above threshold, any HIGH-sensitivity leak); judge scores are informational -- not used for automated decisions (deferred to #37).

- **`RewriteWorkflow`** (`engine/rewrite/rewrite_workflow.py`) -- top-level orchestrator chaining all 6 sub-workflows: domain classification, sensitivity disposition, QA generation, rewrite generation, evaluate-repair loop, and final judge. Evaluate-repair loop runs up to `max_repair_iterations` (from `EvaluationCriteria`), exits early when all rows pass, and only sends failing rows to repair. Final judge is non-critical (failure logged, defaults applied). Fast path skips all LLM calls when no entities detected.

- **Schema cleanup** -- removed `JudgeScoreSchema` / `JudgeEvaluationSchema` (redundant with `LLMJudgeColumnConfig` + `Score` rubrics).

- **Row split/merge helpers** extracted as free functions (`_split_by_entities`, `_merge_and_reorder`, `_apply_passthrough_defaults`) to prepare for #60.

## Config gaps closed
- `evaluation.max_repair_iterations` -- consumed by evaluate-repair loop
- `evaluation.flag_utility_below` -- consumed by `needs_human_review`
- `evaluation.flag_leakage_mass_above` -- consumed by `needs_human_review`

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing
- [x] Tests pass locally

## Test plan
- [x] 15 tests for `FinalJudgeWorkflow` (prompt, rubrics, human review flagging)
- [x] 12 tests for `RewriteWorkflow` (fast path, call order, failed records, attrs, judge failure tolerance, repair loop)
- [x] Full suite: 391 passed

## Related Issues
Closes #35


asteier2026 (approved):

asteier2026 (approved):

src/anonymizer/engine/schemas/__init__.py
```diff
@@ -34,8 +34,6 @@
     EntityCategory,
     EntityDispositionSchema,
     EntitySource,
-    JudgeEvaluationSchema,
```
andreatnvidia, line 37: moving this onto `LLMJudgeColumnConfig` seems like a nice simplification. one thing to consider though: removing `JudgeEvaluationSchema` and `JudgeScoreSchema` is an import break for anyone using those names directly, so maybe a short compat window would help if that python surface matters.

tests/engine/test_rewrite_workflow.py
```diff
@@ -0,0 +1,769 @@
...
+                replace_model_selection=stub_replace_model_selection,
+                privacy_goal=_PRIVACY_GOAL,
+                evaluation=EvaluationCriteria(max_repair_iterations=max_iters),
+            )
+
+        workflow_names = [call.kwargs["workflow_name"] for call in adapter.run_workflow.call_args_list]
+        repair_calls = [n for n in workflow_names if "repair" in n]
+        assert len(repair_calls) == max_iters
+
+    def test_only_failing_rows_sent_to_repair(
```
andreatnvidia: suggestion: could we add one degraded-case test here? something like evaluate or repair returns fewer rows, failed rows are recorded, and the rest of the batch still completes. that would make the intended behavior a lot clearer.

src/anonymizer/engine/rewrite/rewrite_workflow.py
```diff
@@ -0,0 +1,422 @@
...
+                eval_seed,
+                model_configs=model_configs,
+                columns=eval_columns,
+                workflow_name=f"rewrite-evaluate-{iteration}",
+                preview_num_records=preview_num_records,
+            )
+            df = _join_new_columns(df, eval_result.dataframe, overwrite=True)
+            all_failed.extend(eval_result.failed_records)
+
+            needs_repair_mask = df[COL_NEEDS_REPAIR].apply(bool)
```
andreatnvidia, line 376: I think there is still a rough edge if evaluate returns fewer rows than it received. `_join_new_columns()` can skip the short join, but the loop then still reads `_needs_repair`, so partial row loss may still end up crashing the batch. maybe reattaching by stable row identity would make this path safer.

andreatnvidia (changes requested):
with `#68` handling the public wiring, the main thing I am still wondering about in this PR is the short-output case in `RewriteWorkflow`. right now we can warn on a row-count mismatch and keep going, but the loop still seems to assume the joined columns are present afterwards. maybe this wants either stricter failure semantics here, or a more explicit way of carrying partial-row failures through the loop.

src/anonymizer/engine/schemas/__init__.py
lipikaramaswamy: These were internal schemas backing `_judge_evaluation`. They were never part of the public API surface and the replacement (`LLMJudgeColumnConfig` w `Score` rubrics) handles its own output format, so no consumers exist outside this repo.

tests/engine/test_rewrite_workflow.py
lipikaramaswamy: Added `test_evaluate_dropping_rows_degrades_gracefully` where evaluate drops 1 of 2 rows, surviving row completes through judge, failed record tracked in result.

src/anonymizer/engine/rewrite/rewrite_workflow.py
lipikaramaswamy: Ok, `_join_new_columns` now aligns on `RECORD_ID_COLUMN` when lengths differ: filters target to surviving rows, sorts both sides by record ID, then does the column copy. The dropped rows are removed from the working dataframe (not just warned about), so the loop only operates on rows that have all required columns. `RECORD_ID_COLUMN` is included in every seed list so the adapter preserves the stable ID from detection across all calls.

lipikaramaswamy:
> with `#68` handling the public wiring, the main thing I am still wondering about in this PR is the short-output case in `RewriteWorkflow`. right now we can warn on a row-count mismatch and keep going, but the loop still seems to assume the joined columns are present afterwards. maybe this wants either stricter failure semantics here, or a more explicit way of carrying partial-row failures through the loop.

Addressed in the latest commit. `_join_new_columns` no longer skips the join on row-count mismatch -- it filters the working dataframe to surviving rows (via `RECORD_ID_COLUMN`), joins the new columns onto those rows, and lets the dropped rows propagate through `failed_records`. Added a degraded-case test to cover this path.

src/anonymizer/engine/rewrite/rewrite_workflow.py
```diff
@@ -0,0 +1,433 @@
...
+                model_configs=model_configs,
+                columns=self._judge_wf.columns(
+                    selected_models=selected_models,
+                    privacy_goal=privacy_goal,
+                    evaluation=evaluation,
+                ),
+                workflow_name="rewrite-final-judge",
+                preview_num_records=preview_num_records,
+            )
+            df = _join_new_columns(df, judge_result.dataframe)
```
andreatnvidia: follow-up to the record-id join fix: this now also applies to `rewrite-final-judge`, so if judge returns fewer rows we drop those records entirely even though this step is supposed to be non-critical. I reproduced that locally with 2 input rows and 1 judge row out, and the final dataframe shrank to 1 row. I think this path should probably preserve the row and default the judge outputs for the failed record instead of removing it from the result.

andreatnvidia: quick thought on a minimal fix here: maybe special-case the join for `rewrite-final-judge` instead of using the generic row-dropping helper. since judge is supposed to be non-critical, I think it would be safer to preserve the full dataframe, merge returned judge outputs by `RECORD_ID_COLUMN`, and default missing judge rows to something like `_judge_evaluation=None` and `needs_human_review=True` rather than removing the record entirely.

something along these lines:

```python
def _join_judge_columns(target: pd.DataFrame, source: pd.DataFrame) -> pd.DataFrame:
    if len(source) == len(target):
        return _join_new_columns(target, source)

    result = target.copy()
    source_by_id = source.set_index(RECORD_ID_COLUMN)

    if COL_JUDGE_EVALUATION not in result.columns:
        result[COL_JUDGE_EVALUATION] = None
    if COL_NEEDS_HUMAN_REVIEW not in result.columns:
        result[COL_NEEDS_HUMAN_REVIEW] = True

    for idx, record_id in result[RECORD_ID_COLUMN].astype(str).items():
        if record_id in source_by_id.index.astype(str):
            row = source_by_id.loc[record_id]
            result.at[idx, COL_JUDGE_EVALUATION] = row.get(COL_JUDGE_EVALUATION)
            result.at[idx, COL_NEEDS_HUMAN_REVIEW] = row.get(COL_NEEDS_HUMAN_REVIEW, True)

    return result
```

and then in `_run_final_judge()`:

```python
df = _join_judge_columns(df, judge_result.dataframe)
```

that seems closer to the "judge is informational / non-critical" contract to me.

lipikaramaswamy: Fair, eventually and probably as part of #37, we'll have judge be critical - but we can deal with that later.

andreatnvidia: still a bit worried that it's silently dropping rows though, even if it's supposed to be critical eventually. if judge is non-critical in the current implementation, I think partial judge output should preserve rows and default the judge fields. if judge is meant to be critical already, I would expect an explicit hard failure rather than a partial-result path that changes output cardinality without surfacing it. not blocking further on my side, but I do not think `#37` quite resolves this particular behavior.

andreatnvidia (approved):

src/anonymizer/engine/rewrite/rewrite_workflow.py
lipikaramaswamy: That's right, and #37 doesn't call this out, but it is essentially an issue capturing a revamp of the judge step. It's not explicit there, didn't mean to make it sound like that. anyway, the suggestion is in.

src/anonymizer/engine/rewrite/final_judge.py
```diff
@@ -0,0 +1,197 @@
...
+        "10": "Perfect privacy - complete anonymization, impossible to re-identify",
+        "9": "Excellent privacy - virtually impossible to re-identify",
+        "8": "Strong privacy - very difficult to re-identify",
+        "7": "Good privacy - only low-sensitivity context remains",
+        "6": "Fair privacy - minor quasi-identifiers remain",
+        "5": "Acceptable - some low-sensitivity leaks",
+        "4": "Moderate leaks - mostly medium/low sensitivity information",
+        "3": "Significant leaks - some high-sensitivity information remains",
+        "2": "Major privacy issues - multiple high-sensitivity entities exposed",
+        "1": "Severe privacy leaks - direct identifiers easily recoverable",
```
johnnygreco: nit: these scores can be `int` now.

src/anonymizer/engine/rewrite/rewrite_workflow.py
```diff
@@ -0,0 +1,468 @@
...
+                columns=repair_columns,
+                workflow_name=f"rewrite-repair-{iteration}",
+                preview_num_records=preview_num_records,
+            )
+            all_failed.extend(repair_result.failed_records)
+
+            repaired = repair_result.dataframe
+            if COL_REWRITTEN_TEXT_NEXT in repaired.columns:
+                failing_rows[COL_REWRITTEN_TEXT] = repaired[COL_REWRITTEN_TEXT_NEXT].values
+            failing_rows[COL_REPAIR_ITERATIONS] = failing_rows[COL_REPAIR_ITERATIONS].apply(lambda x: int(x) + 1)
```
johnnygreco, line 406: Bug risk (medium-high): If the repair adapter drops rows (tracked via `failed_records`), `repaired` will have fewer rows than `failing_rows`. The `.values` assignment will either raise a `ValueError` on length mismatch or silently misalign data — unlike every other step, this bypasses `_join_new_columns` for alignment.

Suggested fix:
```python
repaired = repair_result.dataframe
failing_rows = _join_new_columns(failing_rows, repaired)
if COL_REWRITTEN_TEXT_NEXT in failing_rows.columns:
    failing_rows[COL_REWRITTEN_TEXT] = failing_rows[COL_REWRITTEN_TEXT_NEXT]
```

– Claude

src/anonymizer/engine/rewrite/rewrite_workflow.py
```diff
@@ -0,0 +1,468 @@
...
+            target[target[RECORD_ID_COLUMN].astype(str).isin(surviving_ids)]
+            .sort_values(RECORD_ID_COLUMN)
+            .reset_index(drop=True)
+        )
+        source = source.sort_values(RECORD_ID_COLUMN).reset_index(drop=True)
+
+    for col in source.columns:
+        if col not in target.columns or overwrite:
+            target[col] = source[col].values
+    return target
```
johnnygreco, line 163: Bug risk (medium-high): With `overwrite=True` (used for re-evaluate after repair on line 564), this iterates **all** `source.columns` — including seed columns that were passed through the adapter. If the adapter transforms, truncates, or re-serializes seed column values, this silently overwrites the canonical copy in `target`.

Consider filtering to only the columns that the workflow step actually produced:
```python
new_cols = [c for c in source.columns if c not in target.columns or overwrite]
```
or passing an explicit list of columns to overwrite rather than relying on `source.columns`.

– Claude

lipikaramaswamy: Fixed by skipping seed/pass-through columns when `overwrite=True` on evaluate merges, so we only overwrite workflow-produced outputs rather than adapter-round-tripped inputs

src/anonymizer/engine/rewrite/rewrite_workflow.py
lipikaramaswamy: Agreed. Fixed by rejoining repair output via `_join_new_columns()` before updating `COL_REWRITTEN_TEXT`, so partial repair row loss stays aligned and doesn’t crash.

andreatnvidia (approved):

lipikaramaswamy:
Thanks all for your review, this was a hairy one. Merging now :)

PR #68, opened 2026-03-25
feat: interface wiring + display

### LINKED ISSUE #36: feat(rewrite): interface wiring + display
Part of #30. Makes rewrite mode visible to users.

## Scope

**`interface/anonymizer.py`**:
- Add `rewrite_runner: RewriteWorkflow | None = None` to `__init__`
- Default-construct `RewriteWorkflow` with all 6 sub-workflows when not injected
- `_run_internal`: call `self._rewrite_runner.run()` when `config.rewrite is not None`
- `_rename_output_columns`: rename `COL_REWRITTEN_TEXT` → `{text_col}_rewritten`
- `_build_user_dataframe`: expose `{text_col}_rewritten`, `utility_score`, `leakage_mass`, `any_high_leaked`, `needs_human_review`
- Detection call: explicitly pass `compute_grouped_entities=True` when rewrite mode (ensures `COL_ENTITIES_BY_VALUE` available for replacement map step)

**`interface/display.py`**:
- `render_record_html`: detect rewrite mode by presence of `{text_col}_rewritten` column, dispatch to `_render_rewrite_html`
- `_render_rewrite_html`: four sections — Original (with entity highlights), Rewritten, Scores (`utility_score`, `leakage_mass`, `needs_human_review` badge, judge scores if present), Entity disposition table
- Existing replace-mode rendering unchanged

**Tests**:
- `tests/interface/test_anonymizer_interface.py`: rewrite path calls `_rewrite_runner.run()`, output includes `{text_col}_rewritten` + score columns, failed records merged
- `tests/interface/test_display.py`: rewrite mode shows rewritten section + scores, replace mode unchanged

## Acceptance criteria
- [ ] `AnonymizerConfig(rewrite=Rewrite())` produces `{text_col}_rewritten` in `result.dataframe`
- [ ] `utility_score`, `leakage_mass`, `any_high_leaked`, `needs_human_review` in `result.dataframe`
- [ ] Internal `_`-prefixed rewrite columns only in `result.trace_dataframe`
- [ ] `display_record()` shows rewrite-mode layout for rewrite results
- [ ] `display_record()` shows existing replace-mode layout for replace results
- [ ] `COL_ENTITIES_BY_VALUE` computed in detection when rewrite mode
- [ ] All existing tests pass

lipikaramaswamy:
> **Note:** This PR is stacked on #64 (final judge + RewriteWorkflow orchestrator). The base branch is `lipikaramaswamy/feat/rewrite-engine-final-judge`. Please review only the diff against that branch. Once #64 merges, this will be rebased onto `main`.

## Summary

- Wires `RewriteWorkflow` into the public `Anonymizer` facade so users can run rewrite mode via `AnonymizerConfig(rewrite=Rewrite())`
- Adds rewrite-mode HTML rendering in `display_record()` with four sections: Original (highlighted entities), Rewritten text, Scores (utility, leakage, needs_human_review, judge), and Entity Disposition table
- Existing replace-mode rendering unchanged
- `compute_grouped_entities` now also `True` for rewrite mode (required by replacement map step)
- Internal `_`-prefixed rewrite columns only appear in `trace_dataframe`; user-facing `dataframe` exposes `{text_col}_rewritten`, `utility_score`, `leakage_mass`, `any_high_leaked`, `needs_human_review`

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing
- [x] Tests pass locally

## Test plan
- [x] 5 new tests for rewrite wiring in `test_anonymizer_interface.py` (runner called, output columns, trace-only internals, merged failures, compute_grouped_entities)
- [x] 4 new tests for rewrite display in `test_display.py` (rewrite layout, judge scores, disposition table, replace mode regression)
- [x] Updated logging test mock for rewrite path
- [x] Full suite: 435 passed

## Related Issues
Closes #36


asteier2026 (approved):

src/anonymizer/interface/anonymizer.py
```diff
@@ -218,15 +223,38 @@ def _run_internal(
...
+                replace_elapsed,
+            )
+        elif config.rewrite is not None:
+            logger.info("✏️ Running rewrite pipeline")
+            t0 = time.perf_counter()
+            rewrite_result = self._rewrite_runner.run(
+                detection_result.dataframe,
+                model_configs=self._model_configs,
+                selected_models=self._selected_models.rewrite,
+                replace_model_selection=self._selected_models.replace,
```
andreatnvidia, line 244: one thing I think changed with this PR: rewrite mode now also depends on the replace model selection, because `RewriteWorkflow` is called with `replace_model_selection=self._selected_models.replace`, but `_validate_preflight_config()` still only checks replace aliases for `Substitute`. that seems to let `validate_config(rewrite=Rewrite())` succeed even if `replace.replacement_generator` is invalid, and defer the failure to runtime. would it make sense to validate the replace aliases whenever rewrite mode is enabled too?

src/anonymizer/interface/display.py
```diff
@@ -338,7 +371,122 @@ def _normalize_replacement_map(raw: str | dict | object) -> list[dict[str, str]]
...
+    if hasattr(raw, "model_dump"):
+        raw = raw.model_dump()
+    if isinstance(raw, str):
+        try:
+            raw = json.loads(raw)
+        except (json.JSONDecodeError, ValueError):
+            return []
+    if not isinstance(raw, dict):
+        return []
+    scores = raw.get("scores", [])
```
andreatnvidia: I think `_extract_judge_scores()` may be assuming the wrong shape here. `LLMJudgeColumnConfig` output is usually keyed by rubric name, with each value carrying `{score, reasoning}`, rather than `{"scores": [...]}`. in that shape the new judge section stays empty even when `_judge_evaluation` is present.

tests/interface/test_display.py
```diff
@@ -449,3 +449,105 @@ def test_display_record_out_of_bounds_raises() -> None:
...
+def test_render_record_html_rewrite_mode_with_judge_scores() -> None:
+    row = pd.Series(
+        {
+            "text": "Alice works at Acme",
+            "text_rewritten": "Beth works at Globex",
+            COL_DETECTED_ENTITIES: {"entities": []},
+            "utility_score": 0.9,
+            "leakage_mass": 0.1,
+            "needs_human_review": False,
+            "_judge_evaluation": {
```
andreatnvidia, line 495: suggestion: could this test use the real judge payload shape from `LLMJudgeColumnConfig`? right now it builds `_judge_evaluation` as `{"scores": [...]}`, which matches the parser here but not the runtime structure, so it may be locking in the wrong contract.

andreatnvidia (changes requested):
I think there are 2 blocking issues in the interface layer here. First, rewrite mode now relies on replace-model aliases without validating them up front, so a bad replace alias can still slip through `validate_config()` and fail later at runtime. Second, the new rewrite display appears to parse judge output in a shape that does not match the actual `LLMJudgeColumnConfig` result, so judge scores can disappear from the preview even when they are present. The rest of the wiring looked fine in my pass, and the changed interface tests passed, but both of these cases reproduced in a small smoke test.

One broader thought from the DD side: this might get simpler if the rewrite interface treats trace-only outputs a bit more explicitly and keeps structured columns in their native shape all the way to display. DD tends to declare side-effect outputs up front, then consume nested structured values directly rather than reshaping them first. That seems especially relevant here for `_judge_evaluation` and the other rewrite trace columns, and it would probably make the display tests easier to keep aligned with runtime output too.

src/anonymizer/interface/anonymizer.py
lipikaramaswamy: Good catch. `check_substitute` now also fires when `config.rewrite is not None`, so `validate_config(rewrite=Rewrite())` will reject a bad `replace.replacement_generator` up front. Added a test for it.

src/anonymizer/interface/display.py
lipikaramaswamy: You're right. Rewrote `_extract_judge_scores` to iterate over the dict keys.

tests/interface/test_display.py
lipikaramaswamy: Fixed alongside the parser so the test now uses the real DD output shape with rubric-keyed dicts and reasoning fields.

lipikaramaswamy:
> I think there are 2 blocking issues in the interface layer here. First, rewrite mode now relies on replace-model aliases without validating them up front, so a bad replace alias can still slip through `validate_config()` and fail later at runtime. Second, the new rewrite display appears to parse judge output in a shape that does not match the actual `LLMJudgeColumnConfig` result, so judge scores can disappear from the preview even when they are present. The rest of the wiring looked fine in my pass, and the changed interface tests passed, but both of these cases reproduced in a small smoke test.
> 
> One broader thought from the DD side: this might get simpler if the rewrite interface treats trace-only outputs a bit more explicitly and keeps structured columns in their native shape all the way to display. DD tends to declare side-effect outputs up front, then consume nested structured values directly rather than reshaping them first. That seems especially relevant here for `_judge_evaluation` and the other rewrite trace columns, and it would probably make the display tests easier to keep aligned with runtime output too.

Agreed on the direction. For this PR, the display layer is a thin consumer that normalizes at the boundary (json.loads for strings, model_validate for dicts). The trace columns flow through as-is so display only reads the ones it needs. Happy to tighten the structured-column contract in a purely display focused follow-up once the rewrite pipeline stabilizes. Tracking in #70

andreatnvidia (approved):
Looks good! Just a few things to keep in mind, maybe for a follow-up:

- the new public-dataframe whitelist in `_build_user_dataframe()` now drops ordinary passthrough columns like `row_id` / `source` from `result.dataframe` in both replace and rewrite mode. that seems like a meaningful user-facing contract change, since those columns were previously preserved and are often what let callers join results back to their source data.
- the switch from `attrs.get("original_text_column", "text")` to strict `attrs["original_text_column"]` is a bit less forgiving for custom injected workflows/runners. probably fine for the built-in path, but it does make that metadata a harder implicit contract at the facade boundary.

asteier2026:
image
[problem_rewrite_trace_dataframe.csv](https://github.com/user-attachments/files/26286288/problem_rewrite_trace_dataframe.csv)

asteier2026:
image

PR #69, opened 2026-03-25
chore: address a handful of cve versions with overrides/updates

mckornfield:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing
- [x] Tests pass locally

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,7 @@ packages = ["src/anonymizer"]
 package = true
 
 override-dependencies = [
-  "grpcio>=1.68.0",  # forcing to a higher version since litellm is constraining it
+  "grpcio>=1.68.0",  # forcing to a higher version since litellm is constraining it, to avoid CVE-2024-11407
 ]
 
 [tool.pytest.ini_options]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -36,6 +36,10 @@ packages = ["src/anonymizer"]
 [tool.uv]
 package = true
 
+override-dependencies = [
+  "grpcio>=1.68.0",  # forcing to a higher version since litellm is constraining it
```
lipikaramaswamy: Is it helpful to add the CVE ID or scanner reference to this comment so we know when the override can be removed? (e.g., when litellm (via data designer) relaxes its grpcio constraint)

lipikaramaswamy (approved):

pyproject.toml
mckornfield:
```suggestion
  "grpcio>=1.68.0",  # forcing to a higher version since litellm is constraining it, to avoid CVE-2024-11407
```

Issue #4, opened 2026-02-25
refactor: replace df.attrs with typed pipeline context dataclass


lipikaramaswamy:
### Priority Level

Medium

### Task Summary

Replace the use of df.attrs (experimental pandas feature) for threading pipeline metadata through workflow stages. Currently original_text_column is carried via df.attrs, which is not preserved through merge/concat/groupby operations.

### Technical Details & Implementation Plan

Create a `PipelineContext` (or similar name) dataclass wrapping a DataFrame + metadata dict. Update `read_input`, `_run_internal`, `LlmReplaceWorkflow`, `_rename_output_columns`, and `_build_user_dataframe` to pass/return this container instead of relying on `df.attrs`. Remove the .attrs comment/TODO in [llm_replace_workflow.py:87](https://github.com/NVIDIA-NeMo/Anonymizer/blob/main/src/anonymizer/engine/replace/llm_replace_workflow.py#L87).

PR #79, opened 2026-03-31
refactor: extract shared row-partitioning helpers

### LINKED ISSUE #60: refactor: extract shared entity-row split/reorder/recombine helper

andreatnvidia:
## Summary

- Extract the duplicated split/reorder/recombine DataFrame pattern into a shared `engine/row_partitioning.py` module with `split_rows()` and `merge_and_reorder()`
- Refactor `rewrite_workflow.py` and `llm_replace_workflow.py` to use the shared helpers instead of inline implementations
- Net reduction of ~46 lines of duplicated logic

Closes #60

## Test plan

- [x] 10 new unit tests for `row_partitioning.py` (partitioning, edge cases, roundtrip)
- [x] All 495 existing tests pass with no changes
- [x] Lint clean (`ruff check`)

### DIFF
--- a/src/anonymizer/engine/replace/llm_replace_workflow.py
+++ b/src/anonymizer/engine/replace/llm_replace_workflow.py
@@ -54,9 +54,7 @@ class LlmReplaceWorkflow:
         working_df["_entities_for_replace_json"] = working_df["_entities_for_replace"].apply(json.dumps)
 
         # Partition: rows with an empty entity list bypass replacement-map generation.
-        entity_rows, passthrough_rows = split_rows(
-            working_df, column="_entities_for_replace", predicate=bool
-        )
+        entity_rows, passthrough_rows = split_rows(working_df, column="_entities_for_replace", predicate=bool)
         passthrough_rows[COL_REPLACEMENT_MAP] = [{"replacements": []} for _ in range(len(passthrough_rows))]
 
         if entity_rows.empty:
@@ -97,7 +95,9 @@ class LlmReplaceWorkflow:
             axis=1,
         )
 
-        combined = merge_and_reorder(output_df, passthrough_rows, attrs={**run_result.dataframe.attrs, **dataframe.attrs})
+        combined = merge_and_reorder(
+            output_df, passthrough_rows, attrs={**run_result.dataframe.attrs, **dataframe.attrs}
+        )
         return LlmReplaceResult(dataframe=combined, failed_records=run_result.failed_records)
 
 
--- a/src/anonymizer/engine/replace/llm_replace_workflow.py
+++ b/src/anonymizer/engine/replace/llm_replace_workflow.py
@@ -15,6 +15,7 @@ from anonymizer.config.models import ReplaceModelSelection
 from anonymizer.engine.constants import COL_ENTITIES_BY_VALUE, COL_REPLACEMENT_MAP, ENTITY_LABEL_EXAMPLES
 from anonymizer.engine.ndd.adapter import FailedRecord, NddAdapter
 from anonymizer.engine.ndd.model_loader import resolve_model_alias
+from anonymizer.engine.row_partitioning import merge_and_reorder, split_rows
 from anonymizer.engine.schemas import EntitiesByValueSchema, EntityReplacementMapSchema
 
 logger = logging.getLogger("anonymizer.replace.llm_workflow")
@@ -45,7 +46,6 @@ class LlmReplaceWorkflow:
         replace_alias = resolve_model_alias("replacement_generator", selected_models)
 
         working_df = dataframe.copy()
-        working_df["_anonymizer_row_order"] = range(len(working_df))
 
         # Parse the per-row entity payload once, then reuse it for prompt inputs.
         parsed_entities = working_df[entities_column].apply(EntitiesByValueSchema.from_raw)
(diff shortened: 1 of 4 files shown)


src/anonymizer/engine/row_partitioning.py
```diff
@@ -0,0 +1,46 @@
...
+"""Shared helpers for partitioning DataFrames by a row predicate and recombining them."""
+
+from __future__ import annotations
+
+from collections.abc import Callable
+from typing import Any
+
+import pandas as pd
+
+_ROW_ORDER_COL = "_anonymizer_row_order"
```
lipikaramaswamy: nit: Now that this is the single source of truth, consider dropping the leading underscore so tests can import it:

```
ROW_ORDER_COL = "_anonymizer_row_order"
```
`test_rewrite_workflow.py` (8 places) and `test_llm_replace_workflow.py` (3 places) hard-code "_anonymizer_row_order" in mock fixtures. Having them import the constant prevents silent drift if the value ever changes.

src/anonymizer/engine/row_partitioning.py
```diff
@@ -0,0 +1,46 @@
...
+    can restore the original ordering after the two partitions are processed
+    independently.
+    """
+    working = df.copy()
+    working[_ROW_ORDER_COL] = range(len(working))
+    mask = working[column].apply(predicate)
+    return working[mask].copy(), working[~mask].copy()
+
+
+def merge_and_reorder(
```
lipikaramaswamy, line 34: nit: `pd.concat([])` raises ValueError if no positional args are passed. Current callers always pass at least one part, but a guard like below would make the contract explicit. 

```
if not parts:
    raise ValueError("merge_and_reorder requires at least one partition")
```

src/anonymizer/engine/replace/llm_replace_workflow.py
```diff
@@ -103,19 +95,10 @@ def generate_map_only(
...
-            .sort_values("_anonymizer_row_order")
-            .drop(columns="_anonymizer_row_order")
-            .reset_index(drop=True)
+        combined = merge_and_reorder(
+            output_df, passthrough_rows, attrs={**run_result.dataframe.attrs, **dataframe.attrs}
         )
-        # Carry pipeline metadata (e.g. original_text_column) through to downstream steps.
-        # pandas .attrs is experimental and not preserved through merge/concat/groupby,
-        # but is preserved through copy which is all we do here.
-        # TODO: consider wrapping df + metadata in a typed container.
```
lipikaramaswamy, line 116: Just noting it's ok to drop this TODO even though it's not addressed because we have issue https://github.com/NVIDIA-NeMo/Anonymizer/issues/4

tests/engine/test_row_partitioning.py
```diff
@@ -0,0 +1,95 @@
...
+    assert combined.attrs == attrs
+
+
+def test_merge_and_reorder_single_partition() -> None:
+    df = pd.DataFrame({"val": [1, 2, 3]})
+    matching, _ = split_rows(df, column="val", predicate=bool)
+    combined = merge_and_reorder(matching, attrs={})
+    assert list(combined["val"]) == [1, 2, 3]
+
+
```
lipikaramaswamy, line 91:
```suggestion

def test_merge_and_reorder_raises_on_empty_parts() -> None:
    with pytest.raises(ValueError, match="at least one partition"):
        merge_and_reorder(attrs={})
```

consider adding to check the guard I suggested above

tests/engine/test_row_partitioning.py
```diff
@@ -0,0 +1,95 @@
...
+def test_roundtrip_preserves_columns_added_after_split() -> None:
+    df = pd.DataFrame({"val": [1, 0, 3]})
+    matching, non_matching = split_rows(df, column="val", predicate=bool)
+
+    matching["extra"] = "processed"
+    non_matching["extra"] = "skipped"
+
+    combined = merge_and_reorder(matching, non_matching, attrs={})
+    assert list(combined["extra"]) == ["processed", "skipped", "processed"]
```
lipikaramaswamy, line 101: Worth also asserting on the column the split was based on.
```
assert list(combined["val"]) == [1, 0, 3]
```

Currently if `merge_and_reorder` scrambled the base data but got `"extra"` right by coincidence, this test would still pass.

lipikaramaswamy (approved):
Looks good, just a few nits and suggestions. e2e test ran without issues.

src/anonymizer/engine/row_partitioning.py
andreatnvidia: Done - tests import the constant now.

src/anonymizer/engine/row_partitioning.py
andreatnvidia: Added. Better than a cryptic `pd.concat` error.

src/anonymizer/engine/replace/llm_replace_workflow.py
andreatnvidia: Yep, leaving as-is since #4 tracks it.

tests/engine/test_row_partitioning.py
andreatnvidia: Added.

tests/engine/test_row_partitioning.py
andreatnvidia: Good catch, added.

Issue #83, opened 2026-04-01
feat: support URL data loading for Colab notebooks


lipikaramaswamy:
## Problem

Tutorial notebooks use local paths for `AnonymizerInput.source` (e.g. `../data/NVIDIA_synthetic_biographies.csv`). This works for local execution and `make convert-notebooks`, but won't work in Google Colab because the data files aren't available.

## Proposed fix

Two changes needed:

1. **`AnonymizerInput`**: Support URL sources -- detect if `source` is a URL, download to a temp file, and pass that to the pipeline. Currently `source` validates with `Path(value).exists()` which rejects URLs.

2. **Notebooks**: Once URL support is added, update all tutorial notebooks to use raw GitHub URLs:
   - `https://raw.githubusercontent.com/NVIDIA-NeMo/Anonymizer/main/docs/data/NVIDIA_synthetic_biographies.csv`
   - `https://raw.githubusercontent.com/NVIDIA-NeMo/Anonymizer/main/docs/data/TAB_legal_sample25.csv`

## Prerequisite

The repo must be public for raw GitHub URLs to work without auth tokens.

## Files

- `src/anonymizer/config/anonymizer_config.py` (URL support in `AnonymizerInput`)
- `docs/notebook_source/*.py` (update data paths)
- `docs/overrides/main.html` (Colab badge already in place)

lipikaramaswamy:
done in #96

PR #84, opened 2026-04-01
docs: documentation site content

### LINKED ISSUE #80: fix: remove {text} from allowed template placeholders in Redact and Hash
## Problem

`Redact` and `Hash` both allow `{text}` in their `format_template` validators, but using it defeats the purpose of these strategies:

- **Redact** with `{text}`: exposes the original PII that was supposed to be redacted
- **Hash** with `{text}`: exposes the original PII alongside the hash

This is inherited from the shared `_validate_template` call in `ReplaceMethodBase` which passes `allowed_fields={"text", "label"}` uniformly.

## Proposed fix

Remove `"text"` from `allowed_fields` in the `validate_format_template` validators for both `Redact` and `Hash` in `src/anonymizer/config/replace_strategies.py`.

Consider also adding a `{mask}` placeholder for Redact that generates `*` characters matching the entity length (separate feature).

## Files

- `src/anonymizer/config/replace_strategies.py`

lipikaramaswamy:
## Summary

Complete documentation site for NeMo Anonymizer, built with MkDocs Material.

### Content
- **Home page**: Landing with tabbed strategy examples, install, quickstart code
- **Concepts**: Models, Detect, Replace, Rewrite -- config tables, examples, admonitions
- **Tutorials**: 5 notebooks with executed outputs (biographies + legal datasets)
- **API Reference**: Auto-generated from docstrings, empty pages excluded

### Code changes
- Reorder Substitute first in `replace_strategies.py` (API ref priority)
- Remove `{text}` from Redact/Hash allowed template fields (fixes #80)
- Reduce `max_parallel_requests` to 2 for rate limit safety
- Add `make convert-notebooks` target
- Move e2e script to `tests_e2e/`
- Polished Tier 1 docstrings (result types, replace methods, logging)

### Styling
- NVIDIA green headers, incognito logo/favicon, card grid styling
- Sidebar nav (no top tabs), collapsed sections, DD-consistent layout
- Light mode code block backgrounds

### Known follow-ups
- Colab notebook support (blocked on public repo, #83)

## Related Issues
Closes #80


docs/notebook_source/01_your_first_anonymization.py
```diff
@@ -0,0 +1,77 @@
...
+# ---
+
+# %% [markdown]
+# # Your First Anonymization
+#
+# Detect sensitive entities and replace them with LLM-generated substitutes.
+# This is the simplest end-to-end example of Anonymizer.
+
+# %% [markdown]
+# ## Setup
```
nina-xu: what is an effective way to set `NVIDIA_API_KEY` in the notebook? I had to add a cell that sets `os.environ`. If there is something to do to set it beforehand, might be good to mention it here

docs/notebook_source/03_choosing_a_replacement_strategy.py
nina-xu: it's not clear to me here where to put the `format_template`. could we do something like 
```suggestion
# Default: `[REDACTED_FIRST_NAME]`. Customize with `Redact(format_template=...)`.
```

nina-xu:
```suggestion
# Default: `<Alice, first_name>`. Customize with `format_template` — must include `{text}` and `{label}`, eg. `Annotate(format_template="<{text}--{label}>")`
```
something like that... there isn't a cell that actually does custom template. same for the other strategies

nina-xu: nit: missing "with"

docs/notebook_source/04_rewriting_biographies.py
```diff
@@ -12,7 +12,7 @@
 # ---
 
 # %% [markdown]
-# # Rewrite Mode
+# # Rewriting Biographies
 #
 # Instead of replacing entities with tokens, rewrite mode generates a
```
nina-xu, line 20: suggestion: point to some replacement notebooks/docs here in case the user does not go through the tutorials in order and is not familiar with replacement yet.

docs/notebook_source/01_your_first_anonymization.py
```diff
@@ -0,0 +1,77 @@
+# ---
```
nina-xu, line 4: before running the notebooks, i had to install ipykernel and register the venv as a kernel. are we assuming these are the steps a user already knows how to do?

docs/notebook_source/05_rewriting_legal_documents.py
```diff
@@ -0,0 +1,123 @@
...
+# and custom entity labels tailored for legal proceedings.
+
+# %% [markdown]
+# ## Setup
+
+# %%
+from anonymizer import Anonymizer, AnonymizerConfig, AnonymizerInput, Detect, LoggingConfig, Rewrite, configure_logging
+from anonymizer.config.rewrite import EvaluationCriteria, PrivacyGoal, RiskTolerance
+
+configure_logging(LoggingConfig.debug())
```
nina-xu: why are we using debug logging here? could we add a comment about that?

nina-xu (commented):
awesome stuff. thanks!

saw that anonymizer replaced SpaceX with Blue Origin. Savage

docs/concepts/detection.md
```diff
@@ -0,0 +1,85 @@
+# Detect
+
+Entity detection is the first stage of every Anonymizer pipeline. Both replace and rewrite modes depend on it.
+
+---
+
+## How it works
+
+Detection combines a lightweight NER model (GLiNER-PII) with LLM-based refinement. GLiNER produces an initial set of entity spans, then an LLM augments it with entities the NER missed and validates each detection -- keeping, reclassifying, or dropping entities based on context.
```
mckornfield:
```suggestion
Detection combines a lightweight NER model (GLiNER-PII) with LLM-based refinement. GLiNER produces an initial set of entity spans, then an LLM augments it with entities that NER missed and validates each detection -- keeping, reclassifying, or dropping entities based on context. 
```

docs/concepts/detection.md
```diff
@@ -0,0 +1,85 @@
...
+
+Detection is configured via the `Detect` object on `AnonymizerConfig`:
+
+```python
+from anonymizer import AnonymizerConfig, Detect, Redact
+
+config = AnonymizerConfig(
+    detect=Detect(
+        entity_labels=["first_name", "last_name", "email", "phone_number"],
+        gliner_threshold=0.3,
```
mckornfield: based on the below, do we want to always be explicit about parameters in the docs? or just leave off defaults?

docs/notebook_source/01_your_first_anonymization.py
alexahaushalter: I would like the notebooks to be totally standalone and include these steps if possible, or at least a comment about prerequisites

docs/concepts/detection.md
```diff
@@ -0,0 +1,85 @@
...
+```python
+# Strict: only detect these 3 labels
+Detect(entity_labels=["first_name", "last_name", "email"])
+
+# Permissive: detect all defaults + LLM can infer new label types
+Detect()  # entity_labels=None
+```
+## Tuning the threshold
+
+For gliner_threshold, start with the default `0.3`. If you're seeing too many false positives, raise it to `0.5`. If entities are being missed, try lowering to `0.2`. The LLM validation step catches many false positives, so erring on the side of lower thresholds is usually safe.
```
mckornfield: might be good to give examples on a sample dataset of "how many false positives -> how many an LLM of a certain size will detect".

docs/concepts/models.md
```diff
@@ -0,0 +1,128 @@
+# Models
+
+Anonymizer uses LLMs for entity detection, replacement, and rewriting. Models are configured via YAML and mapped to workflow roles.
+
+---
+
+## Defaults
+
+Set your API key and Anonymizer uses models hosted on [build.nvidia.com](https://build.nvidia.com).
```
mckornfield:
```suggestion
Set your API key so that Anonymizer will use models hosted on [build.nvidia.com](https://build.nvidia.com).
```

docs/concepts/models.md
```diff
@@ -0,0 +1,128 @@
+# Models
+
+Anonymizer uses LLMs for entity detection, replacement, and rewriting. Models are configured via YAML and mapped to workflow roles.
```
mckornfield, line 6: I wonder if we want to use a note etc. on the fact that we'll be sending the data/snippets of it to these LLMs. It would make sense to use a local/trusted LLM then if they don't/can't send the data to build.nvidia.com

docs/concepts/models.md
```diff
@@ -0,0 +1,128 @@
...
+| `gpt-oss-120b` | `openai/gpt-oss-120b` | Validation, augmentation, replacement, rewriting |
+| `nemotron-30b-thinking` | `nvidia/nemotron-3-nano-30b-a3b` | Latent detection, evaluation, final judge |
+
+Each pipeline stage has a **role** mapped to one of these aliases. See the full role list in the default configs: [`detection.yaml`](https://github.com/NVIDIA-NeMo/Anonymizer/blob/main/src/anonymizer/config/default_model_configs/detection.yaml), [`replace.yaml`](https://github.com/NVIDIA-NeMo/Anonymizer/blob/main/src/anonymizer/config/default_model_configs/replace.yaml), [`rewrite.yaml`](https://github.com/NVIDIA-NeMo/Anonymizer/blob/main/src/anonymizer/config/default_model_configs/rewrite.yaml).
+
+---
+
+## Custom providers
+
+To use models from a non-default provider (OpenAI, OpenRouter, etc.), define a providers YAML:
```
mckornfield:
```suggestion
To use models from a different provider (OpenAI, OpenRouter, etc.), define a providers YAML:
```

docs/concepts/models.md
```diff
@@ -0,0 +1,128 @@
...
+
+```python
+anonymizer = Anonymizer(model_providers="my_providers.yaml")
+```
+
+---
+
+## Custom models
+
+Override specific roles by passing a unified YAML to `Anonymizer(model_configs=...)`. The `provider` field references a provider by name -- use `nvidia` for build.nvidia.com, or a custom provider defined above.
```
mckornfield:
```suggestion
Override specific roles by passing a unified YAML path to `Anonymizer(model_configs=...)`. The `provider` field references a provider by name -- use `nvidia` for build.nvidia.com, or a custom provider defined above.
```

docs/concepts/models.md
```diff
@@ -0,0 +1,128 @@
...
+    model: anthropic/claude-sonnet-4
+    provider: openrouter
+    inference_parameters:
+      max_tokens: 8192
+      temperature: 0.3
+```
+
+```python
+anonymizer = Anonymizer(
+    model_configs="my_models.yaml",
```
mckornfield, line 98: out of curiosity, can it also be a dict/string of YAML? or only a path for these

docs/concepts/models.md
```diff
@@ -0,0 +1,128 @@
...
+
+    Use [`anonymizer.validate_config(config)`](../reference/anonymizer/interface/anonymizer.md) (or [`anonymizer validate`](../reference/anonymizer/interface/cli/main.md) from the CLI) after changing model configs to catch alias mismatches before processing data.
+
+---
+
+## Choosing custom models
+
+For Anonymizer, the best overall leaderboard model is not always the best default for every role.
+Some roles are simple classification or constrained JSON generation tasks, while others require deeper
+reasoning about privacy risk, long-context rewriting, and leakage repair.
```
mckornfield: might be worth defining "leakage repair" unless you did and I missed it

docs/concepts/models.md
```diff
@@ -0,0 +1,128 @@
...
+
+Use benchmarks as signals for role fit, not as a single global ranking.
+
+### Most useful benchmark signals
+
+| Benchmark / metric | What it predicts well in Anonymizer |
```
mckornfield: all of these are benchmarks except the last one yeah? should we just put those 3 on their own and make this `Benchmark` ?

docs/concepts/models.md
```diff
@@ -0,0 +1,128 @@
...
+| `IFBench` | Following detailed instructions, producing constrained outputs, and obeying prompt rules. |
+| `AA-Omniscience Accuracy` | Recovering the right facts without dropping important information. |
+| `AA-Omniscience Non-Hallucination` | Avoiding invented entities, facts, or unsupported claims. |
+| `AA-LCR` | Handling long prompts with tagged text, domain guidance, replacement maps, and evaluation context. |
+| `Humanity's Last Exam` / `GPQA Diamond` | General reasoning depth for privacy-sensitive planning and rewriting. |
+| Latency / output speed / verbosity | Notebook UX, evaluation-loop cost, and practical throughput. |
+
+### Practical guidance
+
+- Use your strongest models for `latent_detector`, `disposition_analyzer`, `rewriter`, and often `repairer`.
```
mckornfield: The `often` is weird to me here and on 127, should we just nix it?

tests_e2e/test_e2e.py
```diff
@@ -0,0 +1,32 @@
...
+preview = anonymizer.preview(
+    config=config,
+    data=input_data,
+    num_records=3,
+)
+
+result = anonymizer.run(config=config, data=input_data)
+
+print(result)
+result.dataframe.head()
```
mckornfield, line 30: should this be written as a test/ have some asserts?

docs/index.md
```diff
@@ -1,3 +1,113 @@
...
+
+```bash
+# Get an API key from build.nvidia.com
+export NVIDIA_API_KEY="your-nvidia-api-key"
+```
+By default, Anonymizer uses NVIDIA-hosted models for detection and LLM-based anonymization. You can also [bring your own models](concepts/models.md).
+
+## Anonymize
+
+```python
```
mckornfield: worth having a tab for how to do it with the CLI as well?

docs/notebooks/04_rewriting_biographies.ipynb
mckornfield, line 1: These notebooks look cool, but the logging snippets are pretty verbose... wonder if they can be run with lower verbosity or something so they aren't as big? might not matter if the sections are scrollable

docs/notebook_source/01_your_first_anonymization.py
mckornfield: that is relatively standard yeah?

docs/notebook_source/05_rewriting_legal_documents.py
mckornfield: mentioned elsewhere, think some of the notebooks are really chonky because of that

docs/notebook_source/02_inspecting_detected_entities.py
nina-xu, line 1: I feel that the motivation of this notebook needs some clarification. Who's the target audience here and how will this help them? is it
- beginner users (i doubt it)
- advanced users (does this help them set a better config to improve performance? does this help them understand the inner workings of anonymizer so that they can be assured what needs to be done is done?)
- developers (to understand which step of the pipeline is falling short so they can fix it?)

docs/concepts/detection.md
lipikaramaswamy: Good point. If there are good defaults, we can just leave them off and describe in the fields section.

docs/notebook_source/02_inspecting_detected_entities.py
lipikaramaswamy: Added, helps investigate and iterate on configs.

docs/notebooks/04_rewriting_biographies.ipynb
lipikaramaswamy: They now have default verbosity

docs/index.md
lipikaramaswamy: added

tests_e2e/test_e2e.py
lipikaramaswamy: The print is unnecessary, and head() gives a glimpse into what one can expect in the output

docs/concepts/models.md
lipikaramaswamy: added a link

docs/concepts/models.md
lipikaramaswamy: good point, moved last one to it's own sentence

docs/concepts/models.md
lipikaramaswamy: gone

docs/concepts/models.md
lipikaramaswamy: yup string only, clarified for both model configs and model providers

docs/concepts/models.md
lipikaramaswamy: yes, 100% thanks for flagging, added.

docs/concepts/detection.md
lipikaramaswamy: tough to say rn! But when we do more benchmarking and have numbers, we can add.

docs/notebook_source/01_your_first_anonymization.py
lipikaramaswamy: alright folks, see [landing page for tutorials](https://laughing-enigma-o34enml.pages.github.io/latest/tutorials/) and the setup section of each notebook ([e.g. 01](https://laughing-enigma-o34enml.pages.github.io/latest/notebooks/01_your_first_anonymization/))

docs/notebook_source/04_rewriting_biographies.py
lipikaramaswamy: updated teh notebook structure, see the end

docs/notebook_source/01_your_first_anonymization.py
lipikaramaswamy: https://github.com/NVIDIA-NeMo/Anonymizer/pull/84#discussion_r3025944677

docs/notebook_source/04_rewriting_biographies.py
nina-xu: i see so there isn't a particular order to go through the notebooks? you know best. it's fine

nina-xu (approved):
approving the notebooks 👍 i didn't read the rest

alexahaushalter:

docs/index.md
```diff
@@ -1,6 +1,140 @@
...
+export NVIDIA_API_KEY="your-nvidia-api-key"
+```
+By default, Anonymizer uses NVIDIA-hosted models for detection and LLM-based anonymization. You can also [bring your own models](concepts/models.md).
+
+## Anonymize
+
+=== "Python"
+
+    ```python
+    from anonymizer import Anonymizer, AnonymizerConfig, AnonymizerInput, Redact
```
alexahaushalter: @lipikaramaswamy Can we update this example to use Substitute instead of Redact? Maybe we can copy the one from the "Your First Anonymization" tutorial?

docs/tutorials/index.md
```diff
@@ -0,0 +1,88 @@
...
+
+Learn the basic end-to-end flow: load data, configure replace mode, preview, inspect, and run.
+
+### 2. [Inspecting Detected Entities](../notebooks/02_inspecting_detected_entities.ipynb)
+
+Understand what detection produced, how entities are labeled, and where they came from.
+
+### 3. [Choosing a Replacement Strategy](../notebooks/03_choosing_a_replacement_strategy.ipynb)
+
+Compare Redact, Annotate, Hash, and Substitute side by side, including custom templates.
```
alexahaushalter: @lipikaramaswamy Thoughts on if we should lead with Substitute here, as well as in the notebook? I don't feel super strongly on this one since Your First Anonymization already leads with Substitute, but wanted to flag!

docs/notebooks/03_choosing_a_replacement_strategy.ipynb
```diff
@@ -0,0 +1,1311 @@
...
+   "id": "d2935cc6",
+   "metadata": {},
+   "source": [
+    "# 🕵️ Choosing a Replacement Strategy\n",
+    "\n",
+    "Four replacement strategies compared side-by-side on the same data.\n",
+    "\n",
+    "#### 📚 What you'll learn\n",
+    "\n",
+    "- Compare **Redact**, **Annotate**, **Hash**, and **Substitute** on the same input\n",
```
alexahaushalter, line 21: @lipikaramaswamy Debating if in this notebook you think we should lead with Substitute as well, or okay not to since Your First Anonymization already shows Substitute. Up to you but wanted to flag!

docs/tutorials/index.md
alexahaushalter: Also suggest we update to explicitly say this is replace mode, e.g. something like

"Compare replace mode options - Redact, Annotate, Hash, and Substitute - side by side, including custom templates."

docs/notebook_source/01_your_first_anonymization.py
```diff
@@ -0,0 +1,118 @@
...
+# %%
+anonymizer = Anonymizer()
+
+# %% [markdown]
+# ## 📦 Load data and configure
+#
+# - `AnonymizerInput` points to your CSV and names the text column. `data_summary`
+#   gives the LLM context about the kind of text it will process.
+# - `AnonymizerConfig` with `Substitute()` tells Anonymizer to replace detected
+#   entities with LLM-generated synthetic values (e.g. fake names, cities, dates).
```
alexahaushalter: maybe something like "e.g. fake - but context-relevant - names, cities, dates)"

docs/notebook_source/03_choosing_a_replacement_strategy.py
```diff
@@ -0,0 +1,210 @@
...
+    num_records=3,
+)
+annotate_custom_preview.display_record(0)
+
+# %% [markdown]
+# ## #️⃣ Hash
+#
+# - Deterministic -- same input always produces the same hash.
+# - Customize `format_template` (must include `{digest}`),
+#   `algorithm` (`sha256`/`sha1`/`md5`), and `digest_length` (6--64).
```
alexahaushalter: I think 6-64 should only have one hyphen

docs/notebook_source/01_your_first_anonymization.py
```diff
@@ -0,0 +1,118 @@
...
+# %%
+from anonymizer import Anonymizer, AnonymizerConfig, AnonymizerInput, Substitute
+
+# %%
+anonymizer = Anonymizer()
+
+# %% [markdown]
+# ## 📦 Load data and configure
+#
+# - `AnonymizerInput` points to your CSV and names the text column. `data_summary`
```
alexahaushalter, line 60: Can we add something here about how long each text cell can be? Rough estimate is fine if that's what we have (e.g. 2K words or less).

docs/index.md
```diff
@@ -1,6 +1,140 @@
...
+
+=== "Hash"
+
+    &lt;HASH_FIRST_NAME_3bc51062973c&gt; met with &lt;HASH_FIRST_NAME_cd9fb1e148cc&gt; and his daughter to review kindergarten application #&lt;HASH_ID_f2a5f83e2a4c&gt;.
+
+=== "Rewrite"
+
+    The family met with the admissions counselor to review their school application.
+
+---
```
alexahaushalter, line 33: I am not sure exactly the best place, but I think somewhere on this main page we need to add a note about how long the text can be (e.g. <2k words per record). Maybe it should be somehow highlighted like in one of the "info" boxes or using > to create the line like |

.github/workflows/build-notebooks.yml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
+        with:
+          version: "0.9.5"
+      - name: Set up Python
+        run: uv python install 3.11
+      - name: Convert and execute notebooks
+        run: make convert-notebooks
```
mckornfield, line 25: wonder if the make target should be named a bit more explicitly like `make-execute-and-convert-notebooks` but proably fine

.github/workflows/build-notebooks.yml
```diff
@@ -0,0 +1,30 @@
...
+      contents: write
+    env:
+      NVIDIA_API_KEY: ${{ secrets.NVIDIA_API_KEY }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
+        with:
+          version: "0.9.5"
```
mckornfield, line 21: we should probably be setting this version in the `pyproject.toml`, but can be follow up

.github/workflows/build-notebooks.yml
```diff
@@ -0,0 +1,30 @@
+name: Build notebooks
+on:
+  workflow_call:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 12 * * MON"
```
mckornfield, line 6: so what's the value in doing the cron vs. just doing it only on release? in that it's used for the docs?

docs/notebooks/02_inspecting_detected_entities.ipynb
```diff
@@ -0,0 +1,719 @@
...
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "--- Record 0 ---\n",
+      "<first_name>Bobby</first_name> <last_name>Watford</last_name>, a <age>40</age>‑year‑old <race_ethnicity>Mexican</race_ethnicity> <occupation>veterinarian</occupation> living in <city>Denver</city>, <state>Colorado</state>, grew up on the outskirts of the city and developed a love for animals early on. After graduating from <school_name>Jefferson High</school_name>, he earned his <education_level>DVM</education_level> at the <university_name>University of Colorado Boulder</university_name>, where he also completed a research stint in wildlife health. Fluent in <language>English</language>, <first_name>Bobby</first_name> has always described his upbringing as a blend of small‑town curiosity and the vibrant culture of his community, values that continue to shape his compassionate approach to animal care.\n",
+      "\n",
+      "Since finishing his training, <first_name>Bobby</first_name> has worked at <company_name>VCA Animal Hospital</company_name> and later at the <company_name>Colorado Veterinary Clinic</com\n",
```
alexahaushalter: All of the examples in this part of the tutorial seem to end abruptly in the middle of their second paragraphs. Is that something we can fix so it shows the entire text for each record? Or if not, add a note explaining to the user why the records are getting cut off?

docs/notebooks/02_inspecting_detected_entities.ipynb
```diff
@@ -0,0 +1,719 @@
...
+      "8                              DVM  education_level             210           213   detector\n",
+      "9   University of Colorado Boulder  university_name             221           251  augmenter\n",
+      "10                         English         language             324           331   detector\n",
+      "11                           Bobby       first_name             333           338   detector\n",
+      "12                           Bobby       first_name             556           561   detector\n",
+      "13             VCA Animal Hospital     company_name             576           595   detector\n",
+      "14      Colorado Veterinary Clinic     company_name             613           639  augmenter\n",
+      "15              Christian Democrat   political_view             707           725   detector\n",
+      "16                            Maya       first_name             798           802   detector\n",
+      "17                    Aria and Leo       first_name             836           848   detector\n",
```
alexahaushalter: It looks like Anonymizer didn't do a great job here, thinking this is all one first name. Flagging in case we want to update the dataset at all to avoid this!

docs/notebooks/03_choosing_a_replacement_strategy.ipynb
```diff
@@ -0,0 +1,1311 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "d2935cc6",
+   "metadata": {},
+   "source": [
+    "# 🕵️ Choosing a Replacement Strategy\n",
+    "\n",
+    "Four replacement strategies compared side-by-side on the same data.\n",
```
alexahaushalter: Image

In the docs I'm seeing this, which doesn't seem to quite match up to what I am seeing in the PR (i.e. docs shows a brief description of each replacement strategy). If we don't show any description, maybe we can at least link to the replacement.md doc that describes the different strategies? https://laughing-enigma-o34enml.pages.github.io/concepts/replace/#strategy-comparison

Or if we do want to briefly describe them (which I think could be very useful to set the stage), I think a table (like in the link) is maybe more helpful than what the screenshot shows.

docs/notebook_source/03_choosing_a_replacement_strategy.py
alexahaushalter: Although actually I see this in the docs somewhere else now, and there is seems to merge it into one long hyphen. So I'm confused why when I saw it this time I don't recall it doing that! Maybe just ignore

docs/notebook_source/04_rewriting_biographies.py
```diff
@@ -0,0 +1,154 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+
+# %% [markdown]
+# # 🕵️ Rewriting Biographies
+#
+# Instead of replacing entities with tokens, rewrite mode generates a
+# privacy-safe paraphrase of the entire text. The pipeline:
+#
+# 1. Detects entities (same as replace mode)
```
alexahaushalter: This isn't exactly the same as replace mode though right? Since it will look for the latent entities? Maybe we should add a note on that

docs/notebook_source/05_rewriting_legal_documents.py
```diff
@@ -0,0 +1,177 @@
...
+# Rewriting legal text (TAB dataset) with a domain-specific privacy goal
+# and custom entity labels tailored for legal proceedings.
+#
+# #### 📚 What you'll learn
+#
+# - Define domain-specific entity labels for legal text (case numbers, court names, etc.)
+# - Configure rewrite mode with legal-specific privacy goals
+# - Preview and run on court decision documents
+# - Triage flagged records with `needs_human_review`
+#
```
alexahaushalter, line 29: At the top of all 5 notebooks, maybe after this "What you'll learn" before the "Tip" about being a first-timer (anywhere is fine though if you have a different suggestion), could we add a rough estimate of how long each notebook is expected to take to run?

docs/notebook_source/05_rewriting_legal_documents.py
```diff
@@ -0,0 +1,177 @@
...
+result.dataframe.head()
+
+# %%
+result.dataframe[["text_rewritten", "utility_score", "leakage_mass", "needs_human_review"]].head()
+
+# %% [markdown]
+# ## 🚩 Filter by review flag
+#
+# - Records where automated metrics exceed thresholds are flagged for manual review.
+# - Use this to prioritize human attention on the records that need it most.
```
alexahaushalter, line 162: Can we add a suggestion on what they should do when they are human reviewing?

For example, "any_high_leaked" is "true" in the example shown, but how can they debug to understand what "high" info was leaked?

Should we encourage them to re-run the pipeline on just these records to try again, or will they almost surely get the same result as before?

Doesn't need to be super comprehensive, but right now I have no idea what I should do next other than not use those records.

Actually - I think we should add this info into the rewrite.md doc, and then link to that part from here saying something like "Need help figuring out next steps for items that need manual review? Check out these suggestions."

We should add this link to both the Rewrite Biographies tutorial and the Rewrite Legal Documents tutorial in this "Filter by review flag" section.

docs/notebook_source/05_rewriting_legal_documents.py
```diff
@@ -0,0 +1,177 @@
+# ---
```
alexahaushalter, line 4: General comment about the Rewrite Biographies & Rewrite Legal Documents notebooks - showing the logs in the page is so overwhelming and makes it very hard to read the tutorial. It would be ideal if we could collapse them or show just a few example log lines. If that isn't possible though, can we at least suppress them altogether for the purposes of the tutorial pages in the Docs?

docs/notebook_source/04_rewriting_biographies.py
```diff
@@ -0,0 +1,154 @@
...
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+
+# %% [markdown]
+# # 🕵️ Rewriting Biographies
+#
+# Instead of replacing entities with tokens, rewrite mode generates a
+# privacy-safe paraphrase of the entire text. The pipeline:
+#
```
asteier2026, line 22: I'm worried that saying "rewrite mode generates a privacy-safe paraphrase of the entire text." isn't quite accurate. Paraphrase means you change words but keep meaning. But we do modify meaning when necessary to reduce identifiability. I think something like "rewrite mode generates a privacy-safe transformation of the entire text." is more accurate.

docs/concepts/rewrite.md
```diff
@@ -0,0 +1,109 @@
+<!-- SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved. -->
+<!-- SPDX-License-Identifier: Apache-2.0 -->
+
+# Rewrite
+
+Instead of replacing individual entities, rewrite mode generates a privacy-safe paraphrase of the entire text, preserving semantic meaning while obscuring sensitive information.
+
```
asteier2026, line 7: Image

docs/concepts/rewrite.md
```diff
@@ -0,0 +1,109 @@
...
+---
+
+## How it works
+
+[Detection](detection.md) runs first (same as [replace mode](replace.md), plus latent entity detection for context-inferable information). Then the text is classified by domain, each entity gets a sensitivity disposition, and an LLM rewrites the text accordingly. The rewrite is evaluated for quality and privacy leakage, with automatic repair if thresholds are exceeded. A final judge sets a `needs_human_review` flag.
+
```
asteier2026, line 15: Image

docs/notebook_source/05_rewriting_legal_documents.py
lipikaramaswamy: Tbh I don't even have consistent numbers on this because build is unpredictable. Suggest we hold off until the model provider situation has a better solution.

docs/notebook_source/04_rewriting_biographies.py
```diff
@@ -0,0 +1,147 @@
...
+
+# %%
+config = AnonymizerConfig(
+    rewrite=Rewrite(
+        privacy_goal=PrivacyGoal(
+            protect="All direct identifiers and quasi-identifier combinations (names, locations, employers, dates)",
+            preserve="Career trajectory, educational background, and professional accomplishments",
+        ),
+        evaluation=EvaluationCriteria(
+            risk_tolerance=RiskTolerance.strict,
```
memadi-nv: Based on the `RiskTolerance` defined in `rewrite.py`, this will probably raise an AttributeError, since we don't have `strict` as an attribute. Closest to this is probably `minimal` I guess.

docs/notebook_source/05_rewriting_legal_documents.py
```diff
@@ -0,0 +1,179 @@
...
+    detect=Detect(
+        entity_labels=LEGAL_ENTITY_LABELS,
+    ),
+    rewrite=Rewrite(
+        privacy_goal=PrivacyGoal(
+            protect="All personal identifiers, case numbers, court names, and institutional references that could identify parties",
+            preserve="Legal reasoning, procedural facts, statutory references, and the structure of the ruling",
+        ),
+        evaluation=EvaluationCriteria(
+            risk_tolerance=RiskTolerance.strict,
```
memadi-nv: Same as [this](https://github.com/NVIDIA-NeMo/Anonymizer/pull/84/changes#r3034655263).

docs/notebook_source/04_rewriting_biographies.py
```diff
@@ -0,0 +1,147 @@
...
+# - Preview rewritten text and inspect utility / leakage scores
+# - Triage flagged records with `needs_human_review`
+#
+# > **Tip:** First time running notebooks? Start with
+# > [setup instructions](https://nvidia-nemo.github.io/Anonymizer/latest/tutorials/).
+
+# %% [markdown]
+# ## ⚙️ Setup
+#
+# - Check if your `NVIDIA_API_KEY` from [build.nvidia.com](https://build.nvidia.com) is registered for model access.
```
memadi-nv, line 42: Description is here but the code for API key setup is missed.

docs/notebook_source/04_rewriting_biographies.py
```diff
@@ -0,0 +1,147 @@
...
+#   the rewriter retries when evaluation fails.
+
+# %%
+config = AnonymizerConfig(
+    rewrite=Rewrite(
+        privacy_goal=PrivacyGoal(
+            protect="All direct identifiers and quasi-identifier combinations (names, locations, employers, dates)",
+            preserve="Career trajectory, educational background, and professional accomplishments",
+        ),
+        evaluation=EvaluationCriteria(
```
memadi-nv: Passing `evaluation=EvaluationCriteria(...)` to `Rewrite()` will raise a Pydantic ValidationError (extra field not permitted). 
Either need to add `evaluation` to `Rewrite()` pydantic class or remove it from here.

docs/notebook_source/05_rewriting_legal_documents.py
```diff
@@ -0,0 +1,179 @@
...
+config = AnonymizerConfig(
+    detect=Detect(
+        entity_labels=LEGAL_ENTITY_LABELS,
+    ),
+    rewrite=Rewrite(
+        privacy_goal=PrivacyGoal(
+            protect="All personal identifiers, case numbers, court names, and institutional references that could identify parties",
+            preserve="Legal reasoning, procedural facts, statutory references, and the structure of the ruling",
+        ),
+        evaluation=EvaluationCriteria(
```
memadi-nv: Same as [here](https://github.com/NVIDIA-NeMo/Anonymizer/pull/84/changes#r3034677163).

docs/notebook_source/03_choosing_a_replacement_strategy.py
```diff
@@ -0,0 +1,217 @@
...
+# # 🕵️ Choosing a Replacement Strategy
+#
+# Four [replace mode](../concepts/replace.md) strategies compared side-by-side on the same data.
+#
+# | Strategy | What it does |
+# |----------|-------------|
+# | **Substitute** | LLM-generated contextual replacements |
+# | **Redact** | Label-based markers (`[REDACTED_FIRST_NAME]`) |
+# | **Annotate** | Tags entities but keeps original text |
+# # | **Hash** | Deterministic hash digest |
```
memadi-nv: Nit: 
```suggestion
#  | **Hash** | Deterministic hash digest |
```

docs/notebook_source/03_choosing_a_replacement_strategy.py
```diff
@@ -0,0 +1,217 @@
...
+    source="../data/NVIDIA_synthetic_biographies.csv",
+    text_column="biography",
+    data_summary="Biographical profiles",
+)
+
+# %% [markdown]
+# ## 🔄 Substitute
+#
+# - Uses an LLM to generate contextually appropriate synthetic replacements.
+# - Unlike the strategies above, the LLM considers the full document context --
```
memadi-nv: nit: "Unlike the strategies above" -- but Substitute is first. Probably need to change the order of the strategies to make this phrase be accurate.

docs/notebook_source/01_your_first_anonymization.py
```diff
@@ -0,0 +1,119 @@
...
+#     display_name: Python 3
+#     language: python
+#     name: python3
+# ---
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# %% [markdown]
+# # 🕵️ Your First Anonymization
+#
```
memadi-nv, line 19: **Just a nit not necessarily needed to be addressed:** Each notebook's # 🕵️ <Title> uses the same detective emoji, making them harder to distinguish at a glance. The cross-reference links already use distinct emojis -- consider using those for the titles too:

01: 🕵️ Your First Anonymization
02: 🔍 Inspecting Detected Entities
03: 🎯 Choosing a Replacement Strategy
04: ✏️ Rewriting Biographies
05: ⚖️ Rewriting Legal Documents

memadi-nv (changes requested):
Thanks Lipika! Left some feedbacks mainly on the notebooks.
Not for this PR necessarily, but we can add something like the [NSS style_guide](https://github.com/NVIDIA-NeMo/Safe-Synthesizer/blob/main/STYLE_GUIDE.md) in the future so all the docs follow the same pattern across the libraries.

docs/index.md
lipikaramaswamy: Done

docs/tutorials/index.md
lipikaramaswamy: done

docs/notebook_source/01_your_first_anonymization.py
lipikaramaswamy: reworded entirely to eliminate the word fake

docs/notebook_source/01_your_first_anonymization.py
lipikaramaswamy: Added a conservative 2k tokens

lipikaramaswamy: but this is subject to change depending on the default model providers we decide on

docs/index.md
lipikaramaswamy: added

docs/notebook_source/03_choosing_a_replacement_strategy.py
lipikaramaswamy: yeah weird, changed

docs/notebooks/03_choosing_a_replacement_strategy.ipynb
lipikaramaswamy: done

docs/notebook_source/04_rewriting_biographies.py
lipikaramaswamy: it's ordered on the landing page.

docs/notebooks/02_inspecting_detected_entities.ipynb
lipikaramaswamy: removed this print, i think display record does what we need it to better

docs/notebooks/02_inspecting_detected_entities.ipynb
lipikaramaswamy: I thought about this one too but here's why I decided it's not actually a failure. Technically this is the efficient route for the LLM, because there is only one mention of the kids names, so bundling them means less replacement error. yea in theory the entity label should be first_names, but I'd forgive that. 

We can show a different example if you feel strongly.

docs/notebooks/03_choosing_a_replacement_strategy.ipynb
lipikaramaswamy: done

docs/notebook_source/05_rewriting_legal_documents.py
lipikaramaswamy: added

docs/notebook_source/04_rewriting_biographies.py
lipikaramaswamy: changed!

.github/workflows/build-notebooks.yml
lipikaramaswamy: the cron is like a smoke test since we hit live APIs on build.nvidia.com. So we'd catch key expirations, model endpoint changes, etc. early rather than at release time. The cron produces artifacts, the tutorial zips are only published on release when `pack-tutorials.yml` runs

docs/notebook_source/04_rewriting_biographies.py
lipikaramaswamy: added

docs/notebook_source/05_rewriting_legal_documents.py
lipikaramaswamy: they're done, it was temporary :)

.github/workflows/build-notebooks.yml
lipikaramaswamy: PR #93  should handle that no?

docs/concepts/rewrite.md
lipikaramaswamy: edited to match the suggestion

docs/concepts/rewrite.md
lipikaramaswamy: edited to match suggestion

docs/notebook_source/01_your_first_anonymization.py
lipikaramaswamy: I chose to keep them all the same so as to reference anonymizer. we can format these as a follow up if strong opinions emerge :)

docs/notebook_source/04_rewriting_biographies.py
lipikaramaswamy: added, thanks

docs/notebook_source/03_choosing_a_replacement_strategy.py
lipikaramaswamy: updated prose, thanks

docs/notebook_source/05_rewriting_legal_documents.py
lipikaramaswamy: params updated, my guess is not relevant anymore

docs/notebook_source/04_rewriting_biographies.py
lipikaramaswamy: params updated, not an issue anymore

docs/notebook_source/04_rewriting_biographies.py
lipikaramaswamy: changed again

memadi-nv: I _think_ it's not pushed yet, not seeing the change for this one

lipikaramaswamy: check [here](https://github.com/NVIDIA-NeMo/Anonymizer/pull/84/changes), make sure you're viewing all commits, not where this notebook link leads to

memadi-nv (approved):
looking great, thanks Lipika!

PR #93, opened 2026-04-03
chore(ci): line up CI with fixes in safe-synthesizer repo

mckornfield:
* After doing some last minute release fixes for safe-synthesizer, this
brings in some of those changes

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -112,7 +112,7 @@ jobs:
           fetch-depth: 0
 
       - name: Download wheel artifact
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v8
         with:
           name: nemo-anonymizer-${{ needs.publish-wheel.outputs.version }}
           path: dist/
@@ -129,6 +129,6 @@ jobs:
 
           gh release create "v${VERSION}" \
             dist/*.whl \
+            --generate-notes \
             --title "v${VERSION}" \
-            --notes-file CHANGELOG.md \
             $PRERELEASE_FLAG
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,15 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+# ---------------------------------------------------------------------------
+# Release: build wheel, publish to PyPI (or Test PyPI on dry-run), and
+# optionally create a GitHub release.
+#
+# Required repository secrets (Settings → Secrets and variables → Actions):
+#   TEST_PYPI_PERSONAL_TOKEN  – API token scoped to nemo-anonymizer on Test PyPI
+#   PYPI_PERSONAL_TOKEN       – API token scoped to nemo-anonymizer on PyPI
+# ---------------------------------------------------------------------------
+
 name: "Release NeMo Anonymizer"
 
 on:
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -59,6 +59,7 @@ style = "pep440"
 packages = ["src/anonymizer"]
 
 [tool.uv]
+required-version = ">=0.5.0"
```
mckornfield, line 62: ideally we'd raise this. I'm on 0.9.14

.github/workflows/release.yml
```diff
@@ -20,34 +29,106 @@ on:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout at release ref
+        uses: actions/checkout@v6
+        with:
+          ref: ${{ inputs.release-ref }}
+          fetch-depth: 0
+
+      - name: Download wheel artifact
+        uses: actions/download-artifact@v4
```
lipikaramaswamy: should this be `actions/download-artifact@v8` ? https://github.com/actions/download-artifact

.github/workflows/release.yml
```diff
@@ -20,34 +29,106 @@ on:
...
+        run: |
+          PRERELEASE_FLAG=""
+          if echo "$VERSION" | grep -qE '(rc|alpha|beta|\.dev)'; then
+            PRERELEASE_FLAG="--prerelease"
+          fi
+
+          gh release create "v${VERSION}" \
+            dist/*.whl \
+            --title "v${VERSION}" \
+            --notes-file CHANGELOG.md \
```
lipikaramaswamy: This assumes the file exists at the release ref right? What happens at first release?

mckornfield: yeah I think this is kinda stupid. if it doesn't exist it just errors. In our repo it's like a blank file, so we just go in an edit it. I think we can make this just use `generate-release-notes` to start, and you can edit it if you want

mckornfield: I see you have one just like NSS does, but it's also just empty

mckornfield:
As a pre-req for running this pipeline, we need:

`TEST_PYPI_PERSONAL_TOKEN` and `PYPI_PERSONAL_TOKEN` added to secrets, which represent pypi tokens made by whoever is going to control the repo (could be a data-designer-team style one, or just a personal account tied to a person)

lipikaramaswamy:
> As a pre-req for running this pipeline, we need:
> 
> `TEST_PYPI_PERSONAL_TOKEN` and `PYPI_PERSONAL_TOKEN` added to secrets, which represent pypi tokens made by whoever is going to control the repo (could be a data-designer-team style one, or just a personal account tied to a person)

Got these in 
image

mckornfield:
Sweet... once this merges, the steps are:
(1) create the tag
(2) run the job with the tag as an input

image


We could follow up where just pushing the tag does the release too, and maybe sends it to test.pypi and regular pypi, but up to you. I don't love decoupling them but it was good while figuring things out

lipikaramaswamy (approved):
lgtm

.github/workflows/release.yml
lipikaramaswamy: hmm not on [main](https://github.com/NVIDIA-NeMo/Anonymizer)?

mckornfield: ah yeah, it must be in my local dir from some rebases haha
```
$ git status
On branch line-up-ci/mck
Untracked files:
  (use "git add <file>..." to include in what will be committed)
        CHANGELOG.md
        out.txt
```

PR #95, opened 2026-04-07
chore: upgrade data-designer to 0.5.5

### LINKED ISSUE #87: feat: upgrade data-designer to 0.5.4 release
### Priority Level

Medium

## Context

Anonymizer currently pins `data-designer==0.5.0`. DataDesigner v0.5.4 adds an
async engine (`async/async-facade` branch):

- Async model facade (`acompletion`, `agenerate_text_embeddings`, `agenerate`)
- `AsyncConcurrentExecutor` with native asyncio `TaskGroup` + `Semaphore`
- Opt-in via `DATA_DESIGNER_ASYNC_ENGINE=1` (sync path remains the default)

## What to do

1. Bump `pyproject.toml` dependency from `data-designer==0.5.0` to `data-designer==0.5.4`
2. Update `NddAdapter.run_workflow` to use the async path (or verify it works
   transparently with the env-var opt-in)
3. Evaluate whether the multi-step rewrite pipeline (detection, disposition,
   rewrite, evaluation, repair) can run workflow steps concurrently via the async executor

## Why

The rewrite pipeline currently runs each NDD workflow synchronously through
`DataDesigner.create()` / `.preview()`. The async engine allows concurrent
LLM calls within each workflow step, reducing end-to-end latency for
large-batch anonymization runs.

## Blocked on

- [ ] DataDesigner v0.5.4 released

_No response_

### Dependencies

_No response_

andreatnvidia:
## Summary

- Bump `data-designer` from 0.5.0 to 0.5.5
- Work around a logging regression introduced in DD 0.5.5 (NVIDIA-NeMo/DataDesigner#388): DD's `_initialize_interface_runtime()` now calls `configure_logging()` during `DataDesigner.__init__()`, which overwrites the `data_designer` logger level and noisy-logger suppression that Anonymizer sets up. We re-apply our logger levels after constructing `DataDesigner` via `reapply_log_levels()`.
- Clean up `tests_e2e/test_e2e.py` (see below)

Closes #87

## E2E test changes

The original e2e test was a script that ran the full pipeline on all 25 records with debug logging. In practice this took 15+ minutes, produced very verbose output, and had no assertions - making it hard to tell whether things were passing or stuck.

Changes:
- **Fixed broken data path** (`docs/notebook_source/data/` no longer exists, data moved to `docs/data/`)
- **Preview only (3 records)** instead of preview + full run on all 25 records. The preview exercises the same code paths (detection, validation, augmentation, rewrite, evaluation, repair) - the only difference is DD's `preview()` vs `create()`, which is a DD concern, not Anonymizer's.
- **Default logging** instead of debug - less noise, progress is still visible via INFO-level stage markers
- **Added assertions** on output (records returned > 0, expected rewrite columns present)
- **Kept it as a script** (not pytest) so logs stream live - important for a multi-minute LLM-backed test where you need to see progress
- Updated Makefile `test-e2e` target to run the script directly

## Test plan

- [x] 516 unit tests pass
- [x] E2e smoke test passes (3 records, 0 failures, ~6 min)
- [x] Verified logging fix: `data_designer` logger stays at WARNING after `Anonymizer()` init

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ requires-python = ">=3.11"
 license = "Apache-2.0"
 
 dependencies = [
-    "data-designer==0.5.0",
+    "data-designer==0.5.5",
     "pydantic>=2.9,<3",
     "cyclopts>=3",
     "pygments>=2.20.0",
--- a/src/anonymizer/interface/anonymizer.py
+++ b/src/anonymizer/interface/anonymizer.py
@@ -98,11 +98,14 @@ class Anonymizer:
         logger.info(LOG_INDENT + "✅ validator: %s", det.entity_validator)
         logger.info(LOG_INDENT + "🧩 augmenter: %s", det.entity_augmenter)
 
-        self._data_designer = data_designer or DataDesigner(
-            artifact_path=resolved_artifact_path,
-            model_providers=_resolve_model_providers(model_providers),
-        )
-        reapply_log_levels()
+        if data_designer is not None:
+            self._data_designer = data_designer
+        else:
+            self._data_designer = DataDesigner(
+                artifact_path=resolved_artifact_path,
+                model_providers=_resolve_model_providers(model_providers),
+            )
+            reapply_log_levels()
         self._adapter = NddAdapter(data_designer=self._data_designer)
         self._detection_workflow = detection_workflow or EntityDetectionWorkflow(adapter=self._adapter)
         self._replace_runner = replace_runner or ReplacementWorkflow(
--- a/src/anonymizer/interface/anonymizer.py
+++ b/src/anonymizer/interface/anonymizer.py
@@ -41,7 +41,7 @@ from anonymizer.engine.replace.replace_runner import ReplacementWorkflow
 from anonymizer.engine.rewrite.rewrite_workflow import RewriteWorkflow
 from anonymizer.interface.errors import InvalidConfigError
 from anonymizer.interface.results import AnonymizerResult, PreviewResult
-from anonymizer.logging import LOG_INDENT, configure_logging
+from anonymizer.logging import LOG_INDENT, configure_logging, reapply_log_levels
 
 if TYPE_CHECKING:
     import pandas as pd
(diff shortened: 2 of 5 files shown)


lipikaramaswamy:
The `data-designer` 0.5.5 workaround makes sense, but I wanted to flag one thing based on the logging contract we established in the original logging PR. My understanding from that discussion was that `Anonymizer()` should try to avoid invasive logging side effects, and that `configure_logging(enabled=False)` was added specifically to give callers a clean opt-out when they already manage logging themselves.

So thinking about the global-state interaction here -- `configure_logging(enabled=False)` sets `_configured` without clearing or updating `_active_config`, and `Anonymizer.__init__()` now unconditionally calls `reapply_log_levels()`. In a long-lived process, that means constructing `Anonymizer()` can still mutate `data_designer` / noisy logger levels based on stale prior config even when the caller opted out of Anonymizer-managed logging.

One possible workaround would be to scope `reapply_log_levels()` to the branch where `Anonymizer` creates the `DataDesigner` itself, instead of calling it unconditionally. That would avoid overriding logger state for callers who opt out of Anonymizer-managed logging or pass in their own preconfigured runtime. I’d also consider making `reapply_log_levels()` a no-op unless logging was explicitly configured with `enabled=True`, and adding a unit test for the `configure_logging(enabled=False)` path so we know stale global config can’t leak into later `Anonymizer()` construction.

lipikaramaswamy:
Also, 
## Related Issues
Closes #87

andreatnvidia:
@lipikaramaswamy Good catch — you're right that the notebook/long-lived process scenario makes this a real concern.

Addressed in the latest push:

1. `configure_logging(enabled=False)` now clears `_active_config`, so `reapply_log_levels()` can't act on stale prior config
2. `reapply_log_levels()` is now scoped to the branch where Anonymizer creates its own DataDesigner — skipped when a pre-configured instance is passed in
3. Added `test_enabled_false_clears_active_config` covering the stale-config path

lipikaramaswamy (approved):
lgtm, thanks!!

lipikaramaswamy:
~Actually sorry, hold on, e2e test ran with errors.~

Never mind, I had set DATA_DESIGNER_ASYNC_ENGINE=1 🙃 All good without that envvar. I will go ahead and merge.

PR #116, opened 2026-04-10
feat: make preview input loading lightweight

### LINKED ISSUE #28: feat: make preview input loading lightweight
### Priority Level

Medium

### Task Summary

Make data loading in `preview()` lightweight. 

Right now `preview()` appears to load the full input dataset up front and only limits later stages via `preview_num_records`. For larger inputs, that makes preview heavier than expected for a fast inspection path.

We want `preview()` to behave like a genuinely lightweight path end to end, not just in downstream pipeline execution.

Today, the preview flow appears to materialize the full input dataset before `preview_num_records` takes effect in later stages. That means preview can still incur the full cost of reading large inputs even when only a small subset of rows is actually processed.

The goal of this task is to change that contract so preview truncates earlier during input loading.

Proposed work:
- Update the preview path so row limiting happens during input loading rather than after the full dataset has already been materialized
- Thread the preview row limit into the input reader layer
- Keep the `run()` path unchanged
- Update preview logging appropriately
- Add tests covering inputs larger than the requested preview size

Likely affected areas:
- `src/anonymizer/interface/anonymizer.py`
- `src/anonymizer/engine/io/reader.py`
- any format-specific reader helpers used by `AnonymizerInput`
- preview/logging tests

### Dependencies

_No response_

memadi-nv:
## Summary

- **Lightweight preview reads:** Move row-limiting (`nrows`) from `_run_internal` into `read_input` so that `preview()` reads only the requested number of rows from disk instead of loading the full dataset and truncating afterward. For CSV this uses `pd.read_csv(nrows=…)` natively; for Parquet it uses `pyarrow.parquet.ParquetFile.iter_batches()` to read only the necessary row batches, avoiding full-file deserialization. Handles `nrows ≤ 0` gracefully by returning a schema-preserving empty DataFrame for both formats.
- **Preserve downstream preview signal**: Keep `preview_num_records` threaded through `_run_internal` into detection, replacement, and rewrite workflows so that `NddAdapter` still routes to `DataDesigner.preview()` (in-memory execution) rather than `DataDesigner.create()` (artifact-backed execution). `run()` passes `None` to use full execution.
- **Clamp preview to available rows**: When `preview_num_records` exceeds the actual number of loaded rows, cap it to the available count and log a "capped" message. When it fits, log the processing count. This prevents downstream workflows from receiving a row count larger than the dataframe.
- **Simplify `_run_internal`**: Remove the old `effective_records` / `min(preview_num_records, num_records)` logic that lived further down the method — the dataframe is now already the right size on entry, and clamping happens up front if necessarily.
- **Move loading logs into `reader.py`**: The "Loaded N records" log now lives in `_load_dataframe`, with a "Preview mode" prefix when `nrows` is set.
- **Add unit tests**:
  - `read_input` `nrows` behavior: CSV truncation, Parquet truncation, `nrows > len`, `nrows=None`, attribute preservation, and remote-CSV forwarding.
  - Workflow-boundary tests: `preview_num_records` is capped and forwarded in `preview()`, forwarded as-is when smaller than available rows, and `None` in `run()`.
  - Logging assertions: verify preview-mode log appears, large-input preview only loads requested rows, and existing logging tests updated to match the new log format.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing

## Documentation

### DIFF
--- a/src/anonymizer/engine/io/reader.py
+++ b/src/anonymizer/engine/io/reader.py
@@ -51,6 +51,8 @@ def _load_dataframe(input_data: AnonymizerInput, *, nrows: int | None = None) ->
         if suffix == ".csv":
             df = pd.read_csv(source_str, nrows=nrows)
         else:
+            # TODO: pd.read_parquet loads the entire file; for a true partial
+            # read use pyarrow.parquet.ParquetFile(...).iter_batches(...).
             df = pd.read_parquet(source_str)
             if nrows is not None:
                 df = df.head(nrows)
--- a/src/anonymizer/engine/io/reader.py
+++ b/src/anonymizer/engine/io/reader.py
@@ -44,9 +44,18 @@ def _validate_internal_column_collision(dataframe: pd.DataFrame, *, selected_tex
 
 
 def _read_parquet_partial(source: str, *, nrows: int | None = None) -> pd.DataFrame:
-    """Read a Parquet file, stopping early when *nrows* is set."""
+    """Read a Parquet file, stopping early when *nrows* is set.
+
+    Returns a schema-preserving empty DataFrame when *nrows* is zero or negative.
+    """
     if nrows is None:
         return pd.read_parquet(source)
+    if nrows <= 0:
+        schema = pq.ParquetFile(source).schema_arrow
+        empty = pa.table(
+            {name: pa.array([], type=field.type) for name, field in zip(schema.names, schema)},
+        )
+        return empty.to_pandas()
     pf = pq.ParquetFile(source)
     batches: list[pa.RecordBatch] = []
     rows_so_far = 0
--- a/src/anonymizer/engine/io/reader.py
+++ b/src/anonymizer/engine/io/reader.py
@@ -3,6 +3,8 @@
 
 from __future__ import annotations
 
+import logging
+
 import pandas as pd
 
 from anonymizer.config.anonymizer_config import AnonymizerInput, infer_input_source_suffix
@@ -10,10 +12,17 @@ from anonymizer.engine.constants import COL_TEXT
 from anonymizer.engine.io.constants import SUPPORTED_IO_FORMATS
 from anonymizer.interface.errors import AnonymizerIOError, InvalidInputError
 
+logger = logging.getLogger("anonymizer")
 
-def read_input(input_data: AnonymizerInput) -> pd.DataFrame:
-    """Load input into a normalized dataframe with canonical internal text column."""
-    dataframe = _load_dataframe(input_data)
+
+def read_input(input_data: AnonymizerInput, *, nrows: int | None = None) -> pd.DataFrame:
+    """Load input into a normalized dataframe with canonical internal text column.
+
+    Args:
+        input_data: Input source definition.
+        nrows: Maximum rows to read.  ``None`` reads the entire file.
+    """
+    dataframe = _load_dataframe(input_data, nrows=nrows)
     selected_text_column = input_data.text_column
     if selected_text_column not in dataframe.columns:
         raise InvalidInputError(f"Input text column '{selected_text_column}' not found.")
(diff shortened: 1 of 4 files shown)


src/anonymizer/interface/anonymizer.py
```diff
@@ -143,7 +144,8 @@ def preview(
             num_records: Maximum records to process (default 10).
         """
         self._validate_preflight_config(config)
-        result = self._run_internal(config=config, data=data, preview_num_records=num_records)
+        input_df = read_input(data, nrows=num_records)
```
lipikaramaswamy, line 147: Previously, `preview()` propagated `preview_num_records` through detection / replace / rewrite, which means `NddAdapter.run_workflow()` used `DataDesigner.preview(...)` rather than `DataDesigner.create(...)`.

After this refactor, preview reads fewer rows up front, but the downstream workflows no longer know they are in preview mode. In Data Designer 0.5.5, preview vs create is a real execution difference, not just a row-count difference:
  - `create()` goes through `builder.build(...)` / artifact-backed results
  - `preview()` goes through `builder.build_preview(...)` + `process_preview(...)` and stays in-memory

Our goal is to still keep the dataframe in memory for preview. So I'd recommend keeping the lightweight input read, but still preserve the downstream preview signal. So `preview()` can call `read_input(data, nrows=num_records)` and still pass `preview_num_records=num_records` through `_run_internal()`.

src/anonymizer/engine/io/reader.py
```diff
@@ -32,15 +41,28 @@ def _validate_internal_column_collision(dataframe: pd.DataFrame, *, selected_tex
...
     if suffix not in SUPPORTED_IO_FORMATS:
         supported_formats = " or ".join(SUPPORTED_IO_FORMATS)
         raise InvalidInputError(f"Unsupported input format: {suffix}. Use {supported_formats}.")
     try:
         if suffix == ".csv":
-            return pd.read_csv(source_str)
-        return pd.read_parquet(source_str)
+            df = pd.read_csv(source_str, nrows=nrows)
+        else:
+            df = pd.read_parquet(source_str)
```
lipikaramaswamy: So csv loading is lightweight, but we don't actually get the same win for parquets. We'd need a true partial parquet reader (for example via `pyarrow.parquet.ParquetFile(...).iter_batches(...)`). If it's too complex, it's ok to leave out of scope for this PR. But let's clarify in the PR description, and perhaps a TODO that we can follow up on later.

src/anonymizer/interface/anonymizer.py
lipikaramaswamy: Also, let's add a test with one assertion at the workflow boundary that `Anonymizer.preview()` still routes through preview mode downstream (for example by verifying `preview_num_records` is forwarded to the detection workflow / adapter path). That would catch the semantic regression above.

memadi-nv: Makes sense, Thanks for the feedback!

memadi-nv: Added some edge cases as well. I might have overcomplicated the conditions. Please let me know your thoughts @lipikaramaswamy

src/anonymizer/engine/io/reader.py
```diff
@@ -32,15 +43,42 @@ def _validate_internal_column_collision(dataframe: pd.DataFrame, *, selected_tex
     )
 
 
-def _load_dataframe(input_data: AnonymizerInput) -> pd.DataFrame:
+def _read_parquet_partial(source: str, *, nrows: int | None = None) -> pd.DataFrame:
```
lipikaramaswamy, line 46: One small edge case to consider here: `_read_parquet_partial()` does not currently guard `nrows <= 0` /
  empty-batch cases. It may be worth adding a tiny regression test for `nrows=0` (and/or empty parquet input) so this path  returns an empty dataframe cleanly instead of relying on `iter_batches(batch_size=0)` / `Table.from_batches([])` behavior.

lipikaramaswamy (approved):
LGTM. I tested with a few 10000+ record parquets and data loading was indeed faster.


One non-blocking test-structure nit: the new `nrows` coverage in `tests/engine/test_io.py` is solid, but there’s a fair amount of repeated file/data setup inline now. I’d consider extracting a small helper or fixture for "write input dataframe to csv/parquet and return an `AnonymizerInput`" so the tests stay focused on the `nrows` behavior instead of the file-construction boilerplate. Also, in `tests/interface/test_anonymizer_logging.py`, it would be cleaner to use `tmp_path` rather than the local `tempfile.mktemp(...)` helper.

PR #119, opened 2026-04-14
feat: enable detection pipeline for async engine

andreatnvidia:
## Summary

Enable the detection pipeline to run on DataDesigner's async engine (`AsyncTaskScheduler`).

- The async engine builds an `ExecutionGraph` where each side-effect column must have exactly one producer. The detection pipeline previously had stages that wrote to the same side-effect columns (e.g. `_initial_tagged_text`, `_seed_entities_json`), which worked in the sync engine (sequential overwrites) but fails in the async DAG
- Removes dead writes from `parse_detected_entities` (columns that were always overwritten before being read)
- Introduces stage-specific column names (`_seed_tagged_text`, `_seed_validation_candidates`) to replace shared ones
- Bumps `data-designer` to >=0.5.7 (includes async engine with side-effect column support, DD#509 and DD#545)
- Adds a benchmark script comparing sync vs async engine performance

## Type of Change
- [x] Enhancement

## Benchmark Results

Full Anonymizer detection pipeline (`EntityDetectionWorkflow.detect_and_validate_entities`), 10 records, 5 iterations after warmup. Models: GLiNER NIM (self-hosted) + GPT-5.4 on `inference-api.nvidia.com`.

| Engine | Mean | Stdev |
|--------|------|-------|
| Sync   | 30.9s | 6.3s |
| Async  | 25.8s | 1.5s |
| **Speedup** | **1.20x** | |

The async engine is ~20% faster and significantly more consistent (stdev 1.5s vs 6.3s). The speedup is modest because the detection DAG has sequential dependencies between columns - the async engine can't parallelize across those, only within each LLM step's concurrent request dispatch.


andreatnvidia:
The detection pipeline works with the async engine after this PR + NVIDIA-NeMo/DataDesigner#509.

The rewrite pipeline has an additional issue: the evaluate and repair workflows use `@custom_column_generator` functions that call `models[alias].generate()` directly. In async engine mode the HTTP client is async-only, so these sync calls fail. DD's built-in LLM generators avoid this by having separate `generate()`/`agenerate()` methods on the `ColumnGenerator` class, but the `@custom_column_generator` decorator only supports a single (sync or async) function.

We have a working Anonymizer-side fix (c634b8d) that adds a `model_generate()` helper which tries sync first and falls back to scheduling `agenerate()` on DD's engine loop. But the proper long-term fix likely belongs in DD - either by exposing a dual-mode helper in the custom column API, or by having the async scheduler handle sync model-calling generators natively.

Taking a look on DD to see if we can fix it there already.

andreatnvidia:
Btw, rewrite benchmark results using the temporary Anonymizer-side fix (3 records, 3 iterations after warmup, GLiNER NIM + GPT-5.4):

| Engine | Mean | Stdev |
|--------|------|-------|
| Sync   | 244.9s | 98.7s |
| Async  | 180.6s | 50.4s |
| **Speedup** | **1.36x** | |

Same pattern as detection - async is faster and more consistent. The high sync variance (one run hit 358s) suggests API throttling that async handles better with concurrent dispatch.

lipikaramaswamy:
I imagine the speedup will be more remarkable on datasets with high variance in input tokens (eg some examples are 1k tokens, some are 100k). Does that sound right?

andreatnvidia:
Update: removed the `model_compat.py` workaround. DataDesigner#545 (merged today) moves the sync-to-async bridge into DD itself, so custom column generators that call `model.generate()` get auto-wrapped facades that transparently bridge to `agenerate()` under the async engine. All 530 tests pass - just waiting on a DD release that includes #509 + #545, then we pin to that version and merge.

@lipikaramaswamy yes, that's right. The async engine dispatches requests concurrently and processes results as they arrive, so high variance in input length means more slack to exploit - short examples complete and free up slots while long ones are still in flight. The sync engine is bottlenecked by the slowest request in each batch. Real Anonymizer datasets tend to have that kind of variance, so the speedup should land closer to the 1.36x we saw on the rewrite pipeline than the 1.20x from the detection benchmark (which used similar-length inputs).

src/anonymizer/engine/detection/custom_columns.py
lipikaramaswamy: I repro’d this on the PR branch with `DATA_DESIGNER_ASYNC_ENGINE=1`, and the intended validate-before-merge flow is still broken in async execution. Even when `apply_validation_to_seed_entities()` overwrites `COL_SEED_ENTITIES`, `merge_and_build_candidates()` still receives the original GLiNER-parsed seed entities from `parse_detected_entities()`, not the validated version written here. So although the pipeline is
  logically ordered as detect -> validate -> augment -> merge, the async DAG is still resolving `_seed_entities` from the earlier parse step. I think this needs either a distinct post-validation seed column that downstream merge reads, or a workflow change that makes this step the
  explicit producer of the seed-entity column consumed after validation.

src/anonymizer/engine/detection/detection_workflow.py
lipikaramaswamy, line 143: This looks like the workflow registration mismatch behind that behavior: this step is registered as producing `COL_SEED_ENTITIES_JSON`, but the post-validation pipeline still needs `merge_and_build_candidates()` to consume the validated seed entities from `COL_SEED_ENTITIES`.

src/anonymizer/engine/detection/custom_columns.py
andreatnvidia: Good catch, fixed in 71c49ba. Root cause was exactly what you described: `apply_validation_to_seed_entities` was rewriting `COL_SEED_ENTITIES` in-place as an undeclared side effect, which the async engine silently strips. Went with your first suggestion — added a distinct `COL_VALIDATED_SEED_ENTITIES` column that this step produces as a declared side effect, and `merge_and_build_candidates` now consumes it instead of `COL_SEED_ENTITIES`. Verified against the real GLiNER NIM + GPT-5.4-mini that sync and async produce identical output.

src/anonymizer/engine/detection/detection_workflow.py
andreatnvidia: Yeah, exactly. The registered-column-name vs actual-write mismatch is what made the DAG unaware of the post-validation write. Fixed in 71c49ba by having `apply_validation_to_seed_entities` declare `COL_VALIDATED_SEED_ENTITIES` as a side effect and rewiring `merge_and_build_candidates` to consume it — the step's registered name (`COL_SEED_ENTITIES_JSON`) stays as-is since that output is still used by the augmenter prompt.

lipikaramaswamy (approved):
Re-ran the async repro against the latest head (`71c49ba`) and confirmed the fix: `merge_and_build_candidates()` now consumes the validated seed entities via `COL_VALIDATED_SEED_ENTITIES` under `DATA_DESIGNER_ASYNC_ENGINE=1`, so the validate-before-merge flow is preserved in the async DAG.

One follow-up we need: Anonymizer should probably expose its own async toggle (for example `ANONYMIZER_ASYNC_ENGINE=1`) instead of asking users to know about `DATA_DESIGNER_ASYNC_ENGINE`. Requiring Anonymizer users to set a DataDesigner-specific env var feels like a leaky abstraction. I wouldn’t block this PR on it since the core async-safety fix is in place, but I do think we should add this before expecting users to benefit from the speedup. And perhaps in the same PR we can add user-facing docs to add the envvar and to note that the change applies to the detection workflow specifically.

lipikaramaswamy:
Chatted with @andreatgretel -- async will become default on Monday's release of Data Designer, so we can just wait for that to happen and upgrade instead of doing envvar and docs work here.

PR #122, opened 2026-04-15
fix: Input Error when duplicate columns in output data frame

### LINKED ISSUE #121: bug: Duplicate column names in output dataframe
### Priority Level

Medium

### Task Summary

Duplicate column names in output dataframe. This issue is created from item [#14](https://docs.google.com/spreadsheets/d/1hslK7nWMM0zslfCTr502P-DgtzAKb4LDC5lcW1_EEnU/edit?gid=0#gid=0) of the bugbash.

When the input CSV already contains a column named `text_replaced` (which is also the Anonymizer's output column name), the output DataFrame has **duplicate column names**:

```
Output columns: ['text', 'text_replaced', 'text_replaced', 'text_with_spans',
                 'final_entities', 'text_replaced', 'text_replaced']
```

**Impact:** `df['text_replaced']` returns ambiguous results. Users who have pre-existing columns with names matching Anonymizer's output columns will get corrupted DataFrames. This could also cause silent data loss when saving to CSV.

**Recommendation:** Either prefix/suffix Anonymizer output columns to avoid collisions, or raise an error/warning if input columns collide with output column names.

_No response_

### Dependencies

_No response_

memadi-nv:
## Summary

Detects and handles input columns whose names would collide with user facing columns the
Anonymizer pipeline writes, so the user's data is never silently overwritten
and the trace/user dataframes never contain duplicate column names.

- **Collision detection in `read_input`**: input columns matching any of the
  pipeline's output names are detected up front, covering both
  text-column-derived names (`{text_col}_replaced`, `_with_spans`, `_rewritten`)
  and fixed output names (`final_entities`, `utility_score`, `leakage_mass`,
  `weighted_leakage_rate`, `any_high_leaked`, `needs_human_review`).
- **Lossless auto-rename (not a hard error)**: colliding input columns are
  renamed by appending `__input` (e.g. `bio_replaced` → `bio_replaced__input`),
  with a `WARNING` log listing each rename. This mirrors pandas' behaviour for
  duplicate column names and means the pipeline always runs; the user's
  original data is preserved under the new name.
- **`text_column` can itself be a fixed output name**: if the user selects
  `text_column="final_entities"` (or any of the other static output names), the
  input column is renamed the same way (`final_entities → final_entities__input`)
  and `original_text_column` is updated to track the renamed identifier. This
  prevents the end-of-pipeline rename from clashing with the pipeline's own
  `final_entities` output, which previously produced duplicate
  `final_entities` columns in the trace/user dataframes.
- **Type-agnostic rename**: the `__input` suffix is applied to any colliding
  column regardless of dtype — text, numeric (`utility_score`, `leakage_mass`,
  `weighted_leakage_rate`), or boolean (`any_high_leaked`, `needs_human_review`).
- **Secondary-collision fallback**: if `{name}__input` is already taken, the
  column is named `{name}__input_2`, `_3`, etc., so the rename can never
  clobber another input column.
- **Reserved internal column still errors**: the existing hard-error for the
  reserved internal sentinel (`__nemo_anonymizer_text_input__`) is unchanged;
  only user-facing output columns are auto-renamed.

## Tests

- Parametrized per-suffix rename-with-warning tests (`_replaced`, `_with_spans`,
  `_rewritten`).
- Parametrized per-static-column rename-with-warning tests when `text_column`
  is unrelated to the collision.
-  Parametrized test (all six static names) for the
  `text_column == fixed output column` case, asserting the rename happens,
  `original_text_column` reflects the renamed identifier, and no duplicate
  `text` column appears.
- Combined test where `text_column` is a static name *and* other
  static columns collide in the same input, asserting each colliding column
  gets its own independent `__input` rename.
- Multi-collision test asserting all renames happen and each name appears
  in the warning (order-independent).
- Disambiguation test covering the `__input_2` fallback.
- Passthrough test confirming non-colliding columns (e.g. `score`,
  `other_replaced`) are untouched.
- End-to-end integration test verifying the renamed input column survives
  into `trace_dataframe` while the pipeline's own output occupies
  `bio_replaced`.
- End-to-end integration test running `anonymizer.run(...)` with
  `text_column="final_entities"` and asserting that neither `result.dataframe`
  nor `result.trace_dataframe` contains duplicate `final_entities` columns —
  the user's text lives under `final_entities__input` and the pipeline's
  entities payload lives under `final_entities`.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing

## Documentation

### DIFF
--- a/src/anonymizer/engine/io/reader.py
+++ b/src/anonymizer/engine/io/reader.py
@@ -34,13 +34,26 @@ def read_input(input_data: AnonymizerInput, *, nrows: int | None = None) -> pd.D
     return dataframe
 
 
+_OUTPUT_COLUMN_SUFFIXES = ("_replaced", "_with_spans", "_rewritten")
+
+
 def _validate_internal_column_collision(dataframe: pd.DataFrame, *, selected_text_column: str) -> None:
-    if COL_TEXT not in dataframe.columns or selected_text_column == COL_TEXT:
-        return
-    raise InvalidInputError(
-        f"Input contains reserved internal column {COL_TEXT!r} while text_column={selected_text_column!r}. "
-        f"Either set text_column={COL_TEXT!r} or remove/rename {COL_TEXT!r} from input."
-    )
+    if COL_TEXT in dataframe.columns and selected_text_column != COL_TEXT:
+        raise InvalidInputError(
+            f"Input contains reserved internal column {COL_TEXT!r} while text_column={selected_text_column!r}. "
+            f"Either set text_column={COL_TEXT!r} or remove/rename {COL_TEXT!r} from input."
+        )
+    collisions = [
+        f"{selected_text_column}{suffix}"
+        for suffix in _OUTPUT_COLUMN_SUFFIXES
+        if f"{selected_text_column}{suffix}" in dataframe.columns
+    ]
+    if collisions:
+        names = ", ".join(repr(c) for c in collisions)
+        raise InvalidInputError(
+            f"Input column(s) {names} collide with Anonymizer output column names. "
+            f"Please rename or remove these columns from your input data."
+        )
 
 
 def _read_parquet_partial(source: str, *, nrows: int | None = None) -> pd.DataFrame:
--- a/src/anonymizer/engine/io/reader.py
+++ b/src/anonymizer/engine/io/reader.py
@@ -37,7 +37,9 @@ def read_input(input_data: AnonymizerInput, *, nrows: int | None = None) -> pd.D
     if selected_text_column not in dataframe.columns:
         raise InvalidInputError(f"Input text column '{selected_text_column}' not found.")
     _validate_internal_column_collision(dataframe, selected_text_column=selected_text_column)
-    dataframe = _resolve_output_column_collisions(dataframe, selected_text_column=selected_text_column)
+    dataframe, selected_text_column = _resolve_output_column_collisions(
+        dataframe, selected_text_column=selected_text_column
+    )
     dataframe = dataframe.rename(columns={selected_text_column: COL_TEXT})
     dataframe.attrs["original_text_column"] = selected_text_column
     return dataframe
@@ -72,7 +74,9 @@ def _validate_internal_column_collision(dataframe: pd.DataFrame, *, selected_tex
         )
 
 
-def _resolve_output_column_collisions(dataframe: pd.DataFrame, *, selected_text_column: str) -> pd.DataFrame:
+def _resolve_output_column_collisions(
+    dataframe: pd.DataFrame, *, selected_text_column: str
+) -> tuple[pd.DataFrame, str]:
     """Rename input columns whose names collide with Anonymizer output columns.
 
     The pipeline writes a known set of output columns derived from the text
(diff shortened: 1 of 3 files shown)


lipikaramaswamy:
Overall the fix is correct and well-tested — the three output suffixes match what's assigned in `anonymizer.py` and the test coverage (parametrized per suffix, multi-collision, non-colliding passthrough, integration test) is solid.

**The bigger question is whether hard-erroring is the right UX.** 

The current approach fails fast, which is safe, but it can be disruptive — the user has to go modify their input data just to run the pipeline. A friendlier alternative would be to auto-rename the colliding input column with a warning (e.g. `bio_replaced` → `bio_replaced__input`). This is consistent with how tools like pandas handle duplicate column names (appending `.1`, `.2`, etc.) handles column conflicts, they warn and proceed rather than blocking. The result is lossless and the pipeline always runs. Optionally, letting users configure the output suffix via config would give a clean escape hatch for power users.

Some code issues in the current impl: 
  1. Double f-string evaluation in the list comprehension — `f"{selected_text_column}{suffix}"` is computed twice per iteration (once in the condition, once as the item). Simple fix: build the candidate list first, then filter.
  2. `final_entities` is also a hardcoded user-facing output column (`anonymizer.py:367`) but isn't checked here. 
  3. The regex in `test_read_input_output_column_collision_lists_all` (`match="'bio_replaced'.*'bio_rewritten'"`) implicitly assumes tuple insertion order. Fine in practice, but fragile if the tuple is ever reordered.
  4. The PR description has a placeholder `# Closes #123`

src/anonymizer/engine/io/reader.py
```diff
@@ -29,18 +37,86 @@ def read_input(input_data: AnonymizerInput, *, nrows: int | None = None) -> pd.D
...
-        return
-    raise InvalidInputError(
-        f"Input contains reserved internal column {COL_TEXT!r} while text_column={selected_text_column!r}. "
-        f"Either set text_column={COL_TEXT!r} or remove/rename {COL_TEXT!r} from input."
+    """Hard-error if the user's input contains the reserved internal text column."""
+    if COL_TEXT in dataframe.columns and selected_text_column != COL_TEXT:
+        raise InvalidInputError(
+            f"Input contains reserved internal column {COL_TEXT!r} while text_column={selected_text_column!r}. "
+            f"Either set text_column={COL_TEXT!r} or remove/rename {COL_TEXT!r} from input."
+        )
```
memadi-nv, line 85: I think we should keep this as is, since it refers to an internal column that matches with the input data, rather than user-facing columns. That said, I wanted to double-check—since we already addressed the user-facing case in this PR, should we consider handling the internal one as well? @lipikaramaswamy

lipikaramaswamy: I suggest we keep the reserved internal column (`__nemo_anonymizer_text_input__`) as a hard error. That still seems like the right behavior since it’s an implementation-reserved name rather than a normal user-facing output column.

lipikaramaswamy:
The current collision handling still misses the case where the selected `text_column` itself is one of the fixed output column names (`final_entities`, `utility_score`, `leakage_mass`, `weighted_leakage_rate`, `any_high_leaked`, `needs_human_review`).

In that case the pipeline later renames the internal text column back to the original text column name while also writing the real output column with that same name, which produces duplicate columns in the trace/user dataframes.

I reproduced this with `text_column="final_entities"` and also with the other fixed output names above. Example outcome for `final_entities`:
  - trace columns: `['final_entities', 'final_entities', ...]`
  - user columns: multiple `final_entities` columns

  So I think this PR still needs to handle the `text_column == fixed output column` case.

memadi-nv:
> The current collision handling still misses the case where the selected `text_column` itself is one of the fixed output column names (`final_entities`, `utility_score`, `leakage_mass`, `weighted_leakage_rate`, `any_high_leaked`, `needs_human_review`).
> 
> In that case the pipeline later renames the internal text column back to the original text column name while also writing the real output column with that same name, which produces duplicate columns in the trace/user dataframes.
> 
> I reproduced this with `text_column="final_entities"` and also with the other fixed output names above. Example outcome for `final_entities`:
> 
> * trace columns: `['final_entities', 'final_entities', ...]`
> * user columns: multiple `final_entities` columns
> 
> So I think this PR still needs to handle the `text_column == fixed output column` case.

Right! This was the edge case I missed, just added the fix, Thanks!

src/anonymizer/engine/io/reader.py
```diff
@@ -72,7 +74,9 @@ def _validate_internal_column_collision(dataframe: pd.DataFrame, *, selected_tex
         )
 
 
-def _resolve_output_column_collisions(dataframe: pd.DataFrame, *, selected_text_column: str) -> pd.DataFrame:
+def _resolve_output_column_collisions(
+    dataframe: pd.DataFrame, *, selected_text_column: str
+) -> tuple[pd.DataFrame, str]:
```
lipikaramaswamy: The fix looks correct. One small design follow-up: instead of returning a raw `(dataframe, selected_text_column)` tuple from `_resolve_output_column_collisions()`, I’d consider introducing a small typed result object here, e.g. `ResolvedInputColumns(dataframe, selected_text_column)`.

That would make the helper a bit more self-describing than positional tuple unpacking, and it also seems more aligned with the longer-term direction in issue #4 `refactor: replace df.attrs with typed pipeline context dataclass` than adding more ad hoc state passing. I wouldn’t suggest a full pipeline-context refactor in this PR, but a tiny local dataclass here seems like a reasonable step in that direction.

lipikaramaswamy (approved):
One code structure / style comment I think we should address before merging. Otherwise looks good!

PR #123, opened 2026-04-16
feat: add strict_entity_protection

asteier2026:
Changes include:
- A new user supplied param strict_entity_protection (default false)
- This param prevents Sensitivity Disposition from choosing leave-as-is

### DIFF
--- a/src/anonymizer/config/anonymizer_config.py
+++ b/src/anonymizer/config/anonymizer_config.py
@@ -115,7 +115,7 @@ class Rewrite(BaseModel):
     )
     strict_entity_protection: bool = Field(
         default=False,
-        description="If True, strongly discourages leave_as_is dispositions in sensitivity analysis.",
+        description="If True, requires every entity to receive a protective disposition during sensitivity analysis.",
     )
 
     @model_validator(mode="after")
--- a/src/anonymizer/config/anonymizer_config.py
+++ b/src/anonymizer/config/anonymizer_config.py
@@ -113,6 +113,10 @@ class Rewrite(BaseModel):
         ge=0,
         description="Maximum repair rounds. Set to 0 to disable repair.",
     )
+    strict_entity_protection: bool = Field(
+        default=False,
+        description="If True, strongly discourages leave_as_is dispositions in sensitivity analysis.",
+    )
 
     @model_validator(mode="after")
     def populate_default_privacy_goal(self) -> Rewrite:
--- a/src/anonymizer/engine/schemas/rewrite.py
+++ b/src/anonymizer/engine/schemas/rewrite.py
@@ -39,7 +39,6 @@ Supporting enums: Domain, EntitySource, EntityCategory, SensitivityLevel,
 from __future__ import annotations
 
 from enum import Enum
-from typing import Literal
 
 from pydantic import BaseModel, ConfigDict, Field, ValidationInfo, model_validator
 
@@ -228,10 +227,23 @@ class StrictEntityDispositionSchema(BaseModel):
     sensitivity: SensitivityLevel
     entity_label: str = Field(min_length=1)
     entity_value: str = Field(min_length=1)
-    needs_protection: Literal[True]
     protection_reason: str = Field(min_length=10, max_length=500)
     protection_method_suggestion: StrictProtectionMethod
     combined_risk_level: CombinedRiskLevel
+
+    def to_entity_disposition(self) -> EntityDispositionSchema:
+        return EntityDispositionSchema(
+            id=self.id,
+            source=self.source,
+            category=self.category,
+            sensitivity=self.sensitivity,
+            entity_label=self.entity_label,
+            entity_value=self.entity_value,
+            needs_protection=True,
+            protection_reason=self.protection_reason,
+            protection_method_suggestion=self.protection_method_suggestion,
+            combined_risk_level=self.combined_risk_level,
+        )
 
 
 class StrictSensitivityDispositionSchema(BaseModel):
@@ -244,6 +256,11 @@ class StrictSensitivityDispositionSchema(BaseModel):
         for i, entry in enumerate(self.sensitivity_disposition, start=1):
             entry.id = i
         return self
+
+    def to_sensitivity_disposition(self) -> SensitivityDispositionSchema:
+        return SensitivityDispositionSchema(
+            sensitivity_disposition=[e.to_entity_disposition() for e in self.sensitivity_disposition]
+        )
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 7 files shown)


lipikaramaswamy:
I wonder if prompt-only enforcement is strong enough for `strict_entity_protection`. In this PR, the flag is wired through the config and into the sensitivity-disposition prompt, but the actual output contract still allows `needs_protection=False` and `protection_method_suggestion="leave_as_is"`.

  Concrete places this shows up:

  - [src/anonymizer/config/anonymizer_config.py](https://github.com/NVIDIA-NeMo/Anonymizer/blob/dce4e473c10887b0c699293537a910c2e1efa515/src/anonymizer/config/anonymizer_config.py) adds the new flag
  - [src/anonymizer/engine/rewrite/rewrite_workflow.py](https://github.com/NVIDIA-NeMo/Anonymizer/blob/dce4e473c10887b0c699293537a910c2e1efa515/src/anonymizer/engine/rewrite/rewrite_workflow.py) passes the flag into the disposition workflow
  - [src/anonymizer/engine/rewrite/sensitivity_disposition.py](https://github.com/NVIDIA-NeMo/Anonymizer/blob/dce4e473c10887b0c699293537a910c2e1efa515/src/anonymizer/engine/rewrite/sensitivity_disposition.py) only changes prompt text / instructions

The problem is that the schema used by `LLMStructuredColumnConfig` still accepts the non-strict shape, so strict mode is ultimately advisory rather than enforced. Downstream code also respects that parsed output as-is.

I’d suggest making this schema-driven instead:

  - in [`src/anonymizer/engine/schemas/rewrite.py`](https://github.com/NVIDIA-NeMo/Anonymizer/blob/dce4e473c10887b0c699293537a910c2e1efa515/src/anonymizer/engine/schemas/rewrite.py), add strict variants:
      - a StrictProtectionMethod enum that excludes leave_as_is
      - a StrictEntityDispositionSchema with `needs_protection: Literal[True]`
      - a StrictSensitivityDispositionSchema container
  - in [`src/anonymizer/engine/rewrite/sensitivity_disposition.py`](https://github.com/NVIDIA-NeMo/Anonymizer/blob/dce4e473c10887b0c699293537a910c2e1efa515/src/anonymizer/engine/rewrite/sensitivity_disposition.py), switch output_format based on strict_entity_protection instead of only changing the prompt
  - we can keep the stronger prompt text as a secondary nudge, but the real guarantee should come from validation, not prompt compliance

That would make strict mode deterministic: if the model emits any `leave_as_is` or `needs_protection=False`, the structured output parse fails instead of silently weakening the policy.

src/anonymizer/engine/schemas/rewrite.py
```diff
@@ -209,6 +210,42 @@ def format_for_rewrite_context(self) -> str:
...
+
+    sensitivity_disposition: list[StrictEntityDispositionSchema] = Field(min_length=1)
+
+    @model_validator(mode="after")
+    def _normalize_ids(self) -> StrictSensitivityDispositionSchema:
+        for i, entry in enumerate(self.sensitivity_disposition, start=1):
+            entry.id = i
+        return self
+
+
```
lipikaramaswamy, line 265: One follow-up refinement to consider: in strict mode, `needs_protection` is no longer really a model decision, so the strict LLM-facing schema could omit it entirely and normalize it back in after parse.

For example: 
```
  class StrictEntityDispositionSchema(BaseModel):
      """Strict variant: all entities are protected; the model only decides how."""

      model_config = ConfigDict(use_enum_values=True)

      id: int = Field(ge=1)
      source: EntitySource
      category: EntityCategory
      sensitivity: SensitivityLevel
      entity_label: str = Field(min_length=1)
      entity_value: str = Field(min_length=1)
      protection_reason: str = Field(min_length=10, max_length=500)
      protection_method_suggestion: StrictProtectionMethod
      combined_risk_level: CombinedRiskLevel

      def to_entity_disposition(self) -> EntityDispositionSchema:
          return EntityDispositionSchema(
              id=self.id,
              source=self.source,
              category=self.category,
              sensitivity=self.sensitivity,
              entity_label=self.entity_label,
              entity_value=self.entity_value,
              needs_protection=True,
              protection_reason=self.protection_reason,
              protection_method_suggestion=self.protection_method_suggestion,
              combined_risk_level=self.combined_risk_level,
          )
```
And similarly on `StrictSensitivityDispositionSchema`:
```
  class StrictSensitivityDispositionSchema(BaseModel):
          ...
          def to_sensitivity_disposition(self) -> SensitivityDispositionSchema:
              return SensitivityDispositionSchema(
                  sensitivity_disposition=[e.to_entity_disposition() for e in self.sensitivity_disposition]
              )
```
This works for enforcement, but I think we should avoid bifurcating the disposition model layer. Either normalize the strict schemas back into the shared disposition types after parse, or factor shared helpers/behavior into a common base so we don’t end up duplicating that logic across strict and non-strict variants.

The nice part of the suggested approach is that after `to_entity_disposition()` / `to_sensitivity_disposition()`, all the existing validators and helpers on `EntityDispositionSchema` / `SensitivityDispositionSchema` continue to apply, so we avoid duplicating most of that logic in the strict variants - which is required if we don't address it now.

src/anonymizer/config/anonymizer_config.py
```diff
@@ -113,6 +113,10 @@ class Rewrite(BaseModel):
         ge=0,
         description="Maximum repair rounds. Set to 0 to disable repair.",
     )
+    strict_entity_protection: bool = Field(
+        default=False,
+        description="If True, strongly discourages leave_as_is dispositions in sensitivity analysis.",
```
lipikaramaswamy:
```suggestion
        description="If True, requires every entity to receive a protective disposition during sensitivity analysis.",
```

lipikaramaswamy (approved):
Re-reviewed. The strict-mode enforcement and normalization back into the shared disposition schema both look good now. I only noticed a minor wording mismatch in the strict_entity_protection field description, but nothing blocking.


Also, documenting some notes from agent experiments:
  - In non-strict mode, a quasi-identifier like Seattle can remain `needs_protection=false` / `leave_as_is`, and it is excluded from `_rewrite_disposition_block`.
  - In strict mode, the same entity is forced into the strict disposition and shows up in `_rewrite_disposition_block` with a real protection method like generalize.
  - `_replacement_map_for_prompt` still only includes entities with `protection_method_suggestion="replace"`, which is the right behavior. generalize, remove, and suppress_inference entries stay in `_rewrite_disposition_block` but do not leak into the replacement map.
  - Latent entities also flow correctly into `_rewrite_disposition_block` under strict mode. I verified a latent gender=female example with suppress_inference, and it was included in the rewrite instructions while correctly omitted from `_replacement_map_for_prompt`.
  - A mixed strict example with replace, generalize, and remove also behaved as expected: all protected entities appeared in `_rewrite_disposition_block`, and only the replace entity remained in `_replacement_map_for_prompt`.

  So the intermediate-column behavior looks internally consistent: strict mode broadens the protected rewrite instruction surface, while the replacement map remains narrowly scoped to replace actions only.

PR #124, opened 2026-04-16
fix(cli): model config file paths and flag fixes

lipikaramaswamy:
## Summary

- `--model-configs` and `--model-providers` now accept `.yaml`/`.yml` file paths from the CLI — previously any string was parsed as inline YAML, causing the cryptic `Expected YAML mapping, got str` error when a filename was passed
- Missing files now raise a clear `Error: Config file not found: <path>` instead of a traceback
- `FileNotFoundError` added to `_cli_error_handler` so it prints cleanly like other user errors
- `_REPLACE_ONLY_FLAGS` / `_REWRITE_ONLY_FLAGS` converted to `ClassVar` — they were leaking into `--help` as `--replace-only-flags` / `--rewrite-only-flags`
- `--max-repair-iterations` default corrected from `2` to `3` to match the SDK's `Rewrite` model
- Added help text to previously undocumented flags: `--model-configs`, `--model-providers`, `--artifact-path`, `--format-template`, `--output`

## Type of Change
- [x] Bug fix
- [x] New feature

## Testing

18 new tests in `tests/interface/cli/test_cli_model_configs.py` covering file-path loading, missing-file errors, inline YAML regression, and clean CLI exit behaviour.

## Documentation

### DIFF
--- a/src/anonymizer/engine/ndd/model_loader.py
+++ b/src/anonymizer/engine/ndd/model_loader.py
@@ -212,7 +212,7 @@ def _parse_yaml_string(raw: str) -> dict[str, Any]:
 
 
 def _load_yaml_dict(path: Path) -> dict[str, Any]:
-    if not path.exists():
+    if not path.is_file():
         raise FileNotFoundError(f"Config file not found: {path}")
     data = load_config_file(path)
     if not isinstance(data, dict):
--- a/src/anonymizer/engine/ndd/model_loader.py
+++ b/src/anonymizer/engine/ndd/model_loader.py
@@ -52,7 +52,11 @@ def parse_model_configs(raw: str | Path | None) -> ParsedModelConfigs:
     if isinstance(raw, Path):
         parsed = _load_yaml_dict(raw)
     else:
-        parsed = _parse_yaml_string(raw)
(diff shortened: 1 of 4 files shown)


asteier2026 (approved):
Looks great Lipika!

memadi-nv (changes requested):
The flag is documented and intended as "path to a YAML config file", so the guard should enforce that: accept regular files, reject anything else (directory, missing path, broken symlink, etc.) with a single clean error message.

Path.exists() is too loose — it returns True for both files and directories. Path.is_file() is the precise check for "this is a regular file I can open() and read".


**Fix:** Swap `path.exists()` for `path.is_file()` in both spots, and broaden `_cli_error_handler`'s except clause from `FileNotFoundError` to `OSError` so any filesystem-level failure still prints cleanly.

memadi-nv (approved):
Looking great! No functional issues. The PR summary has three small inaccuracies given by curser. Again, not a merge blocker really!

- The test count is 23 not 18. 
- OSError was added rather than just FileNotFoundError.
- Couple of extra wins from commits 2–4 aren't called out

PR #126, opened 2026-04-21
feat(detect): chunked validation with validator pools

### LINKED ISSUE #129: LLM validator response rejected when numeric entity values get returned as JSON numbers instead of strings
### Priority Level

Medium (Annoying but has workaround)

### Describe the bug

The entity validator's output schema declares `value: str`, so DataDesigner's pre-validation demands `type: "string"` for every decision's `value` field. Some LLMs (observed with GPT-5.4-mini) occasionally strip the quotes from numeric-looking entity values when filling in the skeleton, returning `"value": 42` instead of `"value": "42"`. DD rejects the response and the record is dropped from the dataset with `Record missing from workflow output`.

This happens probabilistically — in a small smoke-test run against real endpoints, ~1 in 6 records containing an `age` entity (or any other numeric-looking value) got dropped this way. It affects both the sync and async engines identically.

Run detection on any record containing a numeric quasi-identifier like age, using GPT-5.4-mini (or a similarly liberal model) as `entity_validator`:

```python
from anonymizer import Anonymizer
from anonymizer.config.anonymizer_config import AnonymizerConfig, AnonymizerInput, Detect
from anonymizer.config.replace_strategies import Redact

# (model config wiring GPT-5.4-mini as validator elided)

anonymizer.run(
    config=AnonymizerConfig(detect=Detect(), replace=Redact()),
    data=AnonymizerInput(
        source="input.csv",  # single row: "Patient Bob Smith, 42, was admitted..."
        text_column="text",
    ),
)
```

Representative DD warning (from a real run):

```
Non-retryable failure on _validation_decisions[rg=0, row=1]:
  | Cause: The model output from 'openai/openai/gpt-5.4-mini' could not be
  |   parsed into the requested format while running generation for column
  |   '_validation_decisions'. Validation detail: Response doesn't match
  |   requested <response_schema> 42 is not of type 'string' Failed
  |   validating 'type' in schema['properties']['decisions']['items']
  |   ['properties']['value']: {'default': '', 'description': 'Entity value
  |   (echoed from skeleton)', 'title': 'Value', 'type': 'string'}
  |   On instance['decisions'][2]['value']: 42.
```

### Expected behavior

Records should survive the validator regardless of whether the LLM echoes back a numeric value as a JSON string (`"42"`) or a JSON number (`42`). The `value` field is purely echoed context — `enrich_validation_decisions` in `src/anonymizer/engine/detection/custom_columns.py` overwrites it from the candidate lookup before any downstream consumer reads it — so its type shouldn't gate record survival.

### Additional context

**Suggested fix** (not attempted in the PR that surfaced this):

Drop `value` and `label` from `ValidationDecisionSchema` entirely. They're never read downstream, and asking the LLM to echo them is pure cost and failure surface. Concretely:

- `src/anonymizer/engine/schemas/detection.py`: remove `value` and `label` fields from `ValidationDecisionSchema`.
- `src/anonymizer/engine/detection/detection_workflow.py::_get_validation_prompt`: update the few-shot `Output:` line so the example no longer includes `value`/`label` — the skeleton (`Template:` line) still carries them as context for the LLM, just not the output.
- Add a regression test covering a numeric-looking entity value.

**Approach that does NOT work** (flagging it so we don't repeat the mistake):

Loosening the pydantic field to `str | int | float` + a coercion validator passes DD's pre-validation, but DD stores the **raw LLM dict** (not the pydantic-validated object) in the dataframe. Once some records have `"42"` (string) and others have `42` (int) for the same column, PyArrow can't pick a single Arrow dtype at parquet checkpoint time and the whole batch fails with `Could not convert 'Alice' with type str: tried to convert to int64`.

**Environment:**

- `data-designer >= 0.5.7`
- Observed on `gpt-5.4-mini` served via an internal API gateway. Likely reproducible on any LLM that isn't strictly JSON-schema compliant on numeric-string fields.

lipikaramaswamy:
## Summary

Replaces the single per-row entity-validation LLM call with a chunked dispatch that partitions each row's candidates and rotates them across a configurable pool of validator aliases. This unblocks rows with many candidates (which can exceed the validator's context window or the provider's rate limits as a single call), gives users a quota-aggregation knob, and hardens validation against transient provider issues — all without changing defaults or single-validator behaviour.

- **New `Detect` config knobs.** `validation_max_entities_per_call` (default `100`) bounds per-call candidate count; `validation_excerpt_window_chars` (default `500`) bounds the text context per chunk. Rows under the candidate limit dispatch as one call, exactly equivalent to the prior path.
- **`entity_validator` accepts list-form.** A single alias remains valid (normalized to a one-element list); multiple aliases form a **validator pool**. Pools now serve two jobs:
  1. **Load spreading.** Chunks are dispatched round-robin by chunk index, aggregating quota across equivalent endpoints to work around per-alias rate limits.
  2. **Failover.** If a chunk's assigned alias can't complete the call (unrecoverable rate limit, persistent 5xx, malformed response), the same chunk is automatically retried against the other aliases in the pool before the row is dropped. A chunk only fails once every alias in the pool has failed for it. Single-alias pools behave exactly as before.
- **Internal helper split (not user-facing).** The list-form change to `entity_validator` required splitting `anonymizer.engine.ndd.model_loader.resolve_model_alias` into a scalar helper (unchanged for the 12 still-scalar roles) and a new plural helper `resolve_model_aliases()` that returns `list[str]` for list-valued roles and wraps scalars in a one-element list for uniformity. `resolve_model_alias("entity_validator", ...)` now raises `TypeError` rather than silently returning a list. All in-repo callers are updated. These helpers are not re-exported from `anonymizer.__init__` and are not part of the public package surface, so no supported external consumers are affected; the `TypeError` is a loud, trivially-triagable signal for anyone who had imported the private path directly.
- **Row-level failure contract.** A row is only dropped when *every* alias in the pool has failed on at least one of its chunks. Dropped rows surface on `result.failed_records` with `step="detection"`, not as silent passthrough with un-scrubbed text. `anonymizer validate` reports unknown pool aliases by index.
- **Chunked validation via `CustomColumnConfig`** in a new `chunked_validation` module: position-ordered chunking, per-chunk text excerpts with forced-consistent tag notation per row, per-chunk Jinja prompt rendering, dispatch through sync `ModelFacade.generate` (wrapped in `asyncio.to_thread` for per-row chunk concurrency) + `PydanticResponseRecipe`, and merged output back into `COL_VALIDATION_DECISIONS`. The sync facade path is required because DataDesigner's default thread-pool engine hands custom columns sync-mode `HttpModelClient` facades whose `agenerate` raises; it is also forward-compatible with the async engine after [Anonymizer#119](https://github.com/NVIDIA-NeMo/Anonymizer/pull/119) + DD#545 land, since DD#545 bridges sync `.generate()` calls to `agenerate` transparently. Decisions for ids outside the candidate set are dropped, mirroring the legacy `enrich_validation_decisions` filter.
- **Operator warning** when `len(pool) > 1`, noting that `max_parallel_requests` is enforced per alias so pooling multiplies total in-flight calls — rate-limit budget accordingly.
- **Docs.** New "Chunked validation" section on the detection concept page covering both knobs and tuning direction. Updated "Validator pools" subsection on the models page covering both load-spreading and failover roles with YAML example, plus a "What happens when a row can't be validated" subsection on the detection page documenting the row-drop + `failed_records` contract.
- **Tests.** Unit coverage of every pure helper in the chunked module; async `chunked_validate_row` coverage for pool-of-one, pool-of-two round-robin, multi-chunk reassembly, missing-id drop, and bad-config guards; cross-alias failover coverage (primary-fails-secondary-succeeds, whole-pool-exhausted, single-alias-no-failover, wrap-around order); workflow-level tests for the new `CustomColumnConfig` wiring and kwarg threading; regression test pinning the DD-exposed wrapper as sync (guards against re-introducing the coroutine-return bug); behavioural regression test pinning 1-chunk vs 3-chunk parity for `COL_VALIDATION_DECISIONS` and `COL_DETECTED_ENTITIES` under identical deterministic LLM behaviour.
- **Verified against `build.nvidia.com`.** 3-row smoke run with `validation_max_entities_per_call=3` and a `{gpt-oss-120b, nemotron-30b-thinking}` pool dispatched 30 chunks across the pool, hit a live 429 on one chunk, failed over to the other alias, recovered, and completed with 0 dropped rows.

### What users write
Chunking knobs on the Python `Detect` config — safe defaults mean most users never touch these:
```python
from anonymizer import Anonymizer, AnonymizerConfig, Detect, Redact
config = AnonymizerConfig(
    detect=Detect(
        validation_max_entities_per_call=50,   # smaller chunks -> more calls, less context per call
        validation_excerpt_window_chars=800,   # more text context per chunk
    ),
    replace=Redact(),
)
```
List-form `entity_validator` in the models YAML to spread validation across aliases:
```yaml
selected_models:
  detection:
    entity_detector: gliner-pii-detector
    entity_validator:
      - gpt5-primary
      - gpt5-secondary
    entity_augmenter: gpt5-primary
    latent_detector: claude-sonnet
```

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update
- [ ] Refactoring

## Testing

## Documentation

### DIFF
--- a/src/anonymizer/config/models.py
+++ b/src/anonymizer/config/models.py
@@ -26,12 +26,18 @@ class DetectionModelSelection(BaseModel):
     @field_validator("entity_validator", mode="before")
     @classmethod
     def normalize_entity_validator(cls, value: Any) -> list[str]:
-        """Accept either a scalar alias or a list, return a non-empty list.
+        """Accept a scalar alias, a list of aliases, or a tuple of aliases; return a non-empty list.
 
         Normalizing at parse time keeps every downstream consumer on the
         same shape (``list[str]``) regardless of whether the user wrote
         ``entity_validator: some-alias`` or
-        ``entity_validator: [alias-a, alias-b]``.
+        ``entity_validator: [alias-a, alias-b]``. Tuples are accepted for
+        parity with Pydantic v2's default coercion for ``list[str]`` fields,
+        which lets programmatic callers pass either
+        ``DetectionModelSelection(entity_validator=["a", "b"])`` or
+        ``DetectionModelSelection(entity_validator=("a", "b"))`` without
+        caring about the concrete sequence type. Any other input type
+        raises ``TypeError``.
         """
         if isinstance(value, str):
             aliases: list[str] = [value]
--- a/src/anonymizer/config/models.py
+++ b/src/anonymizer/config/models.py
@@ -3,16 +3,46 @@
 
 from __future__ import annotations
 
-from pydantic import BaseModel
+from typing import Any
+
+from pydantic import BaseModel, field_validator
 
 
 class DetectionModelSelection(BaseModel):
-    """Model aliases for the entity detection pipeline."""
+    """Model aliases for the entity detection pipeline.
+
+    ``entity_validator`` accepts either a single alias or a list of aliases.
+    A list forms a validator *pool*: chunked validation rotates calls
+    across the pool in round-robin order, which is useful for bypassing
+    per-alias TPM/RPM limits. A single scalar is normalized to a
+    one-element list.
+    """
 
     entity_detector: str
-    entity_validator: str
+    entity_validator: list[str]
     entity_augmenter: str
     latent_detector: str
+
+    @field_validator("entity_validator", mode="before")
+    @classmethod
+    def normalize_entity_validator(cls, value: Any) -> list[str]:
+        """Accept either a scalar alias or a list, return a non-empty list.
+
+        Normalizing at parse time keeps every downstream consumer on the
+        same shape (``list[str]``) regardless of whether the user wrote
+        ``entity_validator: some-alias`` or
+        ``entity_validator: [alias-a, alias-b]``.
+        """
+        if isinstance(value, str):
+            aliases: list[str] = [value]
+        elif isinstance(value, (list, tuple)):
+            aliases = [str(item) for item in value]
+        else:
+            raise TypeError(f"entity_validator must be a string or list of strings, got {type(value).__name__}")
+        cleaned = [alias.strip() for alias in aliases if alias.strip()]
+        if not cleaned:
+            raise ValueError("entity_validator must name at least one model alias.")
+        return cleaned
 
 
 class ReplaceModelSelection(BaseModel):
--- a/src/anonymizer/config/models.py
+++ b/src/anonymizer/config/models.py
@@ -3,9 +3,12 @@
 
 from __future__ import annotations
 
+import logging
 from typing import Any
 
 from pydantic import BaseModel, field_validator
+
+logger = logging.getLogger(__name__)
 
 
 class DetectionModelSelection(BaseModel):
@@ -26,7 +29,7 @@ class DetectionModelSelection(BaseModel):
     @field_validator("entity_validator", mode="before")
     @classmethod
     def normalize_entity_validator(cls, value: Any) -> list[str]:
-        """Accept a scalar alias, a list of aliases, or a tuple of aliases; return a non-empty list.
+        """Accept a scalar alias, a list of aliases, or a tuple of aliases; return a non-empty deduplicated list.
 
         Normalizing at parse time keeps every downstream consumer on the
         same shape (``list[str]``) regardless of whether the user wrote
@@ -38,6 +41,11 @@ class DetectionModelSelection(BaseModel):
         ``DetectionModelSelection(entity_validator=("a", "b"))`` without
         caring about the concrete sequence type. Any other input type
         raises ``TypeError``.
+
+        Duplicate aliases are collapsed to the first occurrence (order
+        preserved) and a warning is logged. A duplicate in the pool would
+        burn a failover attempt on an already-exhausted endpoint, which
+        almost certainly isn't what the user wants.
         """
         if isinstance(value, str):
             aliases: list[str] = [value]
@@ -48,7 +56,22 @@ class DetectionModelSelection(BaseModel):
         cleaned = [alias.strip() for alias in aliases if alias.strip()]
         if not cleaned:
             raise ValueError("entity_validator must name at least one model alias.")
-        return cleaned
+        seen: set[str] = set()
+        deduped: list[str] = []
+        for alias in cleaned:
+            if alias in seen:
+                continue
+            seen.add(alias)
+            deduped.append(alias)
+        if len(deduped) != len(cleaned):
+            removed = [alias for alias in cleaned if cleaned.count(alias) > 1]
+            logger.warning(
+                "entity_validator pool contained duplicate aliases %s; collapsing to %s. "
+                "Duplicates burn a failover attempt on an already-exhausted endpoint.",
+                sorted(set(removed)),
+                deduped,
+            )
+        return deduped
 
 
 class ReplaceModelSelection(BaseModel):
--- a/src/anonymizer/engine/detection/chunked_validation.py
+++ b/src/anonymizer/engine/detection/chunked_validation.py
@@ -50,6 +50,7 @@ from __future__ import annotations
 
 import asyncio
 import logging
+from collections.abc import Sequence
 from typing import Any
 
 from data_designer.config import custom_column_generator
@@ -154,13 +155,17 @@ def order_candidates_by_position(
 
 
 def chunk_candidates(
-    ordered: list[tuple[Any, EntitySpan]],
+    ordered: Sequence[tuple[Any, EntitySpan]],
     max_entities_per_call: int,
 ) -> list[list[tuple[Any, EntitySpan]]]:
-    """Partition the ordered (candidate, seed) pairs into chunks of at most ``max_entities_per_call``."""
-    if max_entities_per_call <= 0:
-        raise ValueError(f"max_entities_per_call must be > 0, got {max_entities_per_call}.")
-    return [ordered[i : i + max_entities_per_call] for i in range(0, len(ordered), max_entities_per_call)]
+    """Partition the ordered (candidate, seed) pairs into chunks of at most ``max_entities_per_call``.
+
+    ``max_entities_per_call`` is assumed positive. It is validated upstream by
+    ``ChunkedValidationParams`` (``Field(gt=0)``) and by
+    ``AnonymizerDetectConfig.validation_max_entities_per_call`` (``gt=0``), so
+    we do not re-check it here.
+    """
+    return [list(ordered[i : i + max_entities_per_call]) for i in range(0, len(ordered), max_entities_per_call)]
 
 
 def build_chunk_excerpt(
@@ -369,9 +374,20 @@ async def _dispatch_chunk(
                     exc,
                 )
 
-    # Defensive: facades is non-empty (caller guarantees), so the loop either
-    # returned a successful output or recorded last_exc and re-raises here.
-    assert last_exc is not None
+    # Defensive: ``facades`` is non-empty by caller contract
+    # (``chunked_validate_row`` builds it from the configured pool, which the
+    # config validator requires to be non-empty). After a non-empty pool fully
+    # exhausts, ``last_exc`` is guaranteed set and we re-raise it. If we ever
+    # reach this point with ``last_exc is None``, the precondition was
+    # violated — raise a loud ``RuntimeError`` rather than ``raise None``
+    # (which would produce ``TypeError: exceptions must derive from
+    # BaseException``). Using a real check instead of ``assert`` so the guard
+    # survives ``python -O``.
+    if last_exc is None:
+        raise RuntimeError(
+            "_dispatch_chunk was called with an empty facades list; "
+            "this violates the caller contract (a non-empty validator pool)."
+        )
     raise last_exc
 
 
--- a/src/anonymizer/engine/detection/chunked_validation.py
+++ b/src/anonymizer/engine/detection/chunked_validation.py
@@ -426,6 +426,15 @@ def chunked_validate_row(
     ordered = order_candidates_by_position(candidates, all_spans)
     chunks = chunk_candidates(ordered, params.max_entities_per_call)
 
+    logger.debug(
+        "chunked validation: %d candidate(s) in %d chunk(s) (max=%d per chunk, window=%d chars), pool=%s",
+        len(ordered),
+        len(chunks),
+        params.max_entities_per_call,
+        params.excerpt_window_chars,
+        params.pool,
+    )
+
     dispatch_kwargs_per_chunk: list[dict[str, Any]] = []
     for chunk_index, chunk in enumerate(chunks):
         chunk_candidates_ = [pair[0] for pair in chunk]
--- a/src/anonymizer/engine/detection/chunked_validation.py
+++ b/src/anonymizer/engine/detection/chunked_validation.py
@@ -1,0 +1,412 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Chunked LLM validation for the entity detection pipeline.
+
+Validation is the step that asks an LLM to keep/reclass/drop each candidate
+entity produced by the detector. Historically we ran one LLM call per row over
+the full tagged text. For long documents with many candidates this hits
+context-window and TPM/RPM limits. This module replaces that single call with
+a fan-out: partition the row's candidates into chunks, build a small tagged
+excerpt of text around each chunk, render the existing validation prompt per
+chunk, and dispatch each chunk to an alias selected round-robin from a
+configured validator pool.
+
+The per-chunk decisions are merged into a ``ValidationDecisionsSchema``-shaped
+payload so the downstream ``enrich_validation_decisions`` column keeps
+working unchanged.
+
+The public entry point for DataDesigner wiring is
+:func:`make_chunked_validation_generator`, which produces a generator function
+decorated with ``@custom_column_generator`` and bound to a concrete pool.
+The pure helpers below are exposed for unit testing.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import logging
+from typing import Any
+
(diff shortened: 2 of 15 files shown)


src/anonymizer/engine/detection/chunked_validation.py
```diff
@@ -0,0 +1,529 @@
...
+        tasks.append(
+            _dispatch_chunk(
+                facades=chunk_facades,
+                prompt=prompt,
+                system_prompt=params.system_prompt,
+                chunk_index=chunk_index,
+            )
+        )
+
+    # gather() propagates the first exception, cancelling siblings. That's the
```
andreatnvidia: nit: `gather()` doesn't actually cancel siblings on first exception in Python 3.9+ - it propagates the exception but siblings keep running until `asyncio.run()` cleanup tears down the loop. the end behavior is the same but the comment is slightly misleading. wdyt about updating it?

src/anonymizer/config/models.py
```diff
@@ -3,17 +3,53 @@
...
+        caring about the concrete sequence type. Any other input type
+        raises ``TypeError``.
+        """
+        if isinstance(value, str):
+            aliases: list[str] = [value]
+        elif isinstance(value, (list, tuple)):
+            aliases = [str(item) for item in value]
+        else:
+            raise TypeError(f"entity_validator must be a string or list of strings, got {type(value).__name__}")
+        cleaned = [alias.strip() for alias in aliases if alias.strip()]
```
andreatnvidia, line 56: might be worth deduplicating here - `entity_validator: [v0, v0]` currently passes validation but wastes a failover attempt (both entries resolve to the same facade). could add a uniqueness check or silently deduplicate.

src/anonymizer/engine/detection/chunked_validation.py
```diff
@@ -0,0 +1,529 @@
...
+    skeleton: dict[str, Any],
+    notation: TagNotation,
+) -> str:
+    """Render the validation prompt for a single chunk via Jinja2.
+
+    The template and context match the production ``LLMStructuredColumnConfig``
+    call: dicts are rendered with Python ``str()`` (Jinja2 default), which is
+    how the existing prompt has always served ``{{ _validation_skeleton }}``.
+    """
+    compiled = _PROMPT_ENV.from_string(template)
```
andreatnvidia: suggestion: this compiles the same Jinja template on every chunk. for high-entity-count rows (the exact use case this targets), it'd be nice to compile once. a simple `@functools.lru_cache(maxsize=4)` keyed on the template string would do it.

src/anonymizer/engine/detection/chunked_validation.py
```diff
@@ -0,0 +1,529 @@
...
+    opt-in via ``DATA_DESIGNER_ASYNC_ENGINE=1``). The thread-pool path calls
+    ``generator.generate(record)`` which synchronously invokes our function
+    and passes its return value straight into ``_postprocess_result``. If the
+    outer function were ``async``, the sync caller would receive a coroutine
+    object and DD would reject it with ``"must return a dict, got coroutine"``.
+    The sync wrapper here runs the async row logic on a fresh event loop
+    (safe because each DD worker thread has no ambient loop) and returns the
+    resolved dict, so this works under both the default thread engine and
+    the opt-in async engine (which wraps sync generators in
+    ``asyncio.to_thread``).
```
andreatgretel: once #119 lands (DD >= 0.5.7), the `asyncio.run()` + `asyncio.to_thread(facade.generate)` stack could be simplified to a plain `ThreadPoolExecutor`. with `_AsyncBridgedModelFacade` in DD#545, `facade.generate()` works transparently under both engines, so you don't need async machinery for fan-out. something like:

```python
with ThreadPoolExecutor(max_workers=max_parallel_requests) as pool:
    futures = [pool.submit(_dispatch_chunk, ...) for chunk in chunks]
    results = [f.result() for f in futures]
```

this also solves the `max_parallel_requests` bypass - `max_workers` gives you the concurrency cap that DD's scheduler can't enforce on this `CustomColumnConfig` path. not blocking, but worth considering as a follow-up simplification.

andreatnvidia (commented):
**Additional comments on unchanged lines:**

**`src/anonymizer/engine/ndd/model_loader.py:95`** - `str(value)` in `load_workflow_selections` will silently mangle a list - now that `entity_validator` supports list-form, this becomes a trap if someone adds a pool to the bundled `detection.yaml`. maybe preserve native types here (`return {str(key): value for ...}`) or at least add a comment warning about it?

**`src/anonymizer/engine/detection/custom_columns.py:138`** - `build_validation_skeleton` is no longer used in the detection pipeline (replaced by `build_chunk_skeleton` in the chunked path). worth removing to avoid confusion about which skeleton builder is authoritative?

src/anonymizer/engine/detection/chunked_validation.py
andreatnvidia: the async machinery here could be simplified in two steps as upstream work lands:

**step 1 (after #119 merges / rebase onto DD >= 0.5.7):** the `asyncio.to_thread(facade.generate)` calls can be replaced with plain `facade.generate()` calls inside a `ThreadPoolExecutor`. DD#545's `_AsyncBridgedModelFacade` bridges `generate()` to `agenerate()` transparently under the async engine, so the `asyncio.to_thread` wrapping is no longer needed for engine compatibility. using `ThreadPoolExecutor(max_workers=max_parallel_requests)` also gives you the concurrency cap that DD's scheduler can't enforce on this `CustomColumnConfig` path - addresses the `max_parallel_requests` bypass.

**step 2 (once async engine is default):** the entire sync wrapper + `asyncio.run()` pattern can be dropped in favor of a native async generator:

```python
async def chunked_validate(row, params, models):
    tasks = [models[alias].agenerate(prompt=...) for ...]
    results = await asyncio.gather(*tasks)
    return merge(results, row)
```

DD's `agenerate()` goes through `ThrottledModelClient`, so concurrency control comes from the framework. the sync wrapper rationale (lines 497-508), the engine compatibility docs (lines 35-46), and the `asyncio.to_thread` dance all become unnecessary.

not blocking - the current approach works correctly under both engines. but worth tracking as a follow-up so this doesn't stay more complex than it needs to be.

src/anonymizer/engine/detection/chunked_validation.py
```diff
@@ -0,0 +1,529 @@
...
+            label=entity.label,
+            start_position=entity.start_position,
+            end_position=entity.end_position,
+            score=entity.score,
+            source=entity.source,
+        )
+        for entity in seed_entities_schema.entities
+    ]
+
+    ordered = order_candidates_by_position(candidates, all_spans)
```
andreatnvidia, line 426: optimization idea: if the same entity value+label appears multiple times in a document (e.g. "John Smith" / `first_name` 15 times), each instance gets validated independently across potentially multiple chunks. validating unique `(value, label)` pairs once and propagating the decision to all matching candidates would directly reduce the number of chunks and LLM calls - the most expensive part of this pipeline. for real-world documents with repeated names this could be a significant win.

src/anonymizer/engine/detection/chunked_validation.py
```diff
@@ -0,0 +1,529 @@
...
+        rotated_aliases = [params.pool[(start + offset) % len(params.pool)] for offset in range(len(params.pool))]
+        chunk_facades = [(alias, models[alias]) for alias in rotated_aliases]
+        tasks.append(
+            _dispatch_chunk(
+                facades=chunk_facades,
+                prompt=prompt,
+                system_prompt=params.system_prompt,
+                chunk_index=chunk_index,
+            )
+        )
```
andreatnvidia, line 491: the loop eagerly builds `excerpt`, `skeleton`, and full `prompt` strings for every chunk before the first `await asyncio.gather()`. peak memory scales with `num_chunks * prompt_size` - for a row with 1000 entities and `max_entities_per_call=10`, that's ~100 prompts held simultaneously. moving prompt construction inside each dispatched coroutine (or batching dispatches with a semaphore) would bound peak memory to only the in-flight chunks. not urgent at the default `max_entities_per_call=100`, but worth considering since large documents are the target use case.

src/anonymizer/engine/detection/chunked_validation.py
```diff
@@ -0,0 +1,529 @@
...
+    max_entities_per_call: int,
+) -> list[list[tuple[Any, EntitySpan]]]:
+    """Partition the ordered (candidate, seed) pairs into chunks of at most ``max_entities_per_call``.
+
+    Assumes ``max_entities_per_call > 0``; positivity is enforced upstream at
+    ``ChunkedValidationParams.max_entities_per_call`` and
+    ``AnonymizerDetectConfig.validation_max_entities_per_call`` (both
+    ``Field(gt=0)``).
+    """
+    return [list(ordered[i : i + max_entities_per_call]) for i in range(0, len(ordered), max_entities_per_call)]
```
andreatnvidia, line 170: future optimization: chunks split on entity count regardless of how much text each entity contributes. adjacent chunks with overlapping excerpt windows can resend the same surrounding text to the LLM. chunking by estimated token budget (entities + excerpt size) would produce fewer, better-balanced calls and reduce total tokens sent. not blocking - entity count is a reasonable first cut - but worth tracking if token spend becomes a concern.

src/anonymizer/engine/detection/chunked_validation.py
lipikaramaswamy: Done in f093b99 — wrapped `_compile_template` in `@functools.lru_cache(maxsize=4)` keyed on template source.

src/anonymizer/engine/detection/chunked_validation.py
lipikaramaswamy: good catch — the comment is gone entirely with the async simplification in f093b99. `gather()` got replaced by `ThreadPoolExecutor` + `f.result()`, and the new inline comment correctly notes pending workers finish naturally as the `with` block exits; we just stop observing their results once the first exception re-raises.

src/anonymizer/config/models.py
lipikaramaswamy: done in 841899f — `normalize_entity_validator` collapses duplicates to the first occurrence (order preserved) and logs a warning naming the removed aliases. Went with warn-and-dedupe over hard-reject so a misconfigured pool still boots with something sensible. Tests cover both the dedup path and the no-dup-no-warning path.

src/anonymizer/engine/detection/chunked_validation.py
lipikaramaswamy: step 1 done in f093b99 — dropped `asyncio.run` / `asyncio.to_thread` / `asyncio.gather`; `chunked_validate_row` fans out via `ThreadPoolExecutor` and calls `facade.generate()` directly. Step 2 (full async-native via `facade.agenerate()` + `asyncio.gather`) is deferred behind a `TODO(async-native)` in the module docstring, pinned to "once DD's async engine becomes the default" so we don't keep straddling both engines. 

One place I diverged: `max_workers=len(chunks)` instead of `max_workers=max_parallel_requests`. `max_parallel_requests` is a per-alias knob, and in a pool of N aliases with N different per-alias caps there's no single row-level value that's semantically correct. Per-alias concurrency is already enforced downstream by each facade's `ThrottledModelClient` (AIMD on 429), so the row-level executor is purely for overlapping chunks — not the right layer for a rate cap. Maybe something like `min(len(chunks), sum_of_per_alias_caps)` works too.

src/anonymizer/engine/detection/chunked_validation.py
lipikaramaswamy: Agreed on the shape of the concern, deferring on whether to ship the fix now. Ran the numbers: ~5KB/chunk at defaults (2×window + skeleton + ~3KB template overhead). Even the worst case you flag (1000 entities, `max_entities_per_call=10`) is ~500KB — not where we'd expect OOMs.

The correct fix is also more than lazy construction alone: without a row-level concurrency cap, N workers race construction simultaneously and peak memory stays at O(N). So the real shape is "lazy build + new `validation_max_concurrent_chunks` config knob", and I don't have real-workload data to pick a default.

Module docstring (f093b99) now quantifies the scaling and cross-references the follow-up.

src/anonymizer/engine/detection/chunked_validation.py
lipikaramaswamy: Yea this makes sense, thought about it but decided to keep it simple for now because the consideration is not only input size but output size too. And we need some good testing for it. Follow up issue coming.

src/anonymizer/engine/detection/chunked_validation.py
lipikaramaswamy: The concern with `(value, label)` dedup is that the same surface form + label can carry different correctness across a document. Short names that collide with common English words are the canonical case: the detector tags `"Ira"` as `first_name` in both "Ira said hello" (true positive) and "he grew irate" (false positive from a substring span), producing identical `(value, label)` keys. Same for `Mark`/`mark`, `Bill`/`bill`, `Hope`/`hope`, etc. The validator is precisely the component whose job is to resolve these; deduping erases the question before the validator can answer it, and we have no signal when it does.

andreatnvidia (approved):
LGTM! Two minor follow-ups:

1. `merge_chunk_decisions`: `seen.add()` runs before null-decision filtering - a null entry can shadow a valid decision for the same ID from another chunk. Unlikely since candidates partition cleanly, but the insertion order is technically wrong.

2. Single-chunk rows now go through `build_chunk_excerpt()` with a 500-char window instead of getting the full tagged text. A fast path for `len(chunks) == 1` would preserve parity with the old behavior.

src/anonymizer/engine/detection/chunked_validation.py
lipikaramaswamy: https://github.com/NVIDIA-NeMo/Anonymizer/issues/134

lipikaramaswamy:
[smoke_chunked_validation.py](https://github.com/user-attachments/files/27021891/smoke_chunked_validation.py)
Just tracking the file i used to test chunked validation

PR #128, opened 2026-04-21
add plan-records-drop-silently

### LINKED ISSUE #110: bug: Data Designer failures silently drop records — add visibility and retry
## Summary

The NDD adapter in `src/anonymizer/engine/ndd/adapter.py` wraps Data Designer calls but has two paths where records are dropped silently. Additionally, there is no retry mechanism for failed records despite the adapter already tracking them via `WorkflowRunResult.failed_records`.

## Silent failure paths

**1. Preview mode returns empty DataFrame on None dataset**
In `NddAdapter.run_workflow()`, if `preview_results.dataset is None`, the adapter returns an empty DataFrame (`workflow_input_df.iloc[0:0].copy()`) with no log message explaining why or how many records were lost.

**2. Missing RECORD_ID_COLUMN causes detection bypass**
`_detect_missing_records()` only flags omissions if `RECORD_ID_COLUMN` is present in the output. If Data Designer strips this column, the method short-circuits silently:
```pytn
if RECORD_ID_COLUMN not in output_df.columns:
    return []  # all missing records go unreported
```

## Missing retry logic

The adapter already has the infrastructure needed for retries:
- `WorkflowRunResult.failed_records` tracks every record that didn't appear in output, with `record_id`, `step`, and `reason`
- `_compute_record_id()` generates deterministic UUID5s so failed rows can be precisely identified and re-submitted
- `rewrite_workflow.py` already accumulates `all_failed` across steps

But nothing uses this to retry. Failed records are logged and discarded.

## Proposed fix

**Visibility:**
- [ ] Log a warning with record count and reasons when `preview_results.dataset is None`
- [ ] When `RECORD_ID_COLUMN` is missing from output, warn that missing-record detection is disabled rather than returning empty silently

**Retry loop in `NddAdapter.run_workflow()` (or caller in `rewrite_workflow.py`):**
- [ ] After each `run_workflow()` call, check `result.failed_records`
- [ ] If failed records exist and `retry_attempt < max_retries`, extract the failed rows from the original input DataFrame using their record IDs and re-submit them
- [ ] Merge retry successes back into the main output DataFrame
- [ ] Accumulate records that fail all retries into `all_failed` with reason `"max_retries_exceeded"`
- [ ] Expose `max_retries` as a config parameter (suggested default: 2)
- [ ] Log each retry attempt: `"Retrying %d failed records (attempt %d/%d)"`

## Notes

The RAT-Bench 33-record run lost 4 records due to `nvidia/nemotron-3-nano-30b-a3b` internal server errors with 3–6 hour hangs — a retry with backoff would have recovered these automatically.

memadi-nv:
## Summary
- Wrap NDD `preview` and `create`+`load_dataset` in `try/except` that logs row count, model aliases, and exception type/message before re-raising.
- Warn (instead of silently dropping) when `preview_results.dataset is None`, and let `_detect_missing_records` populate `failed_records`.
- Warn in both `_detect_missing_records` short-circuits (input missing and output missing `RECORD_ID_COLUMN`), with a column-diff breadcrumb on the output case.
- Add `max_retries` kwarg (default 2) to `NddAdapter` with a retry loop that re-submits only the failed subset; leftovers get `reason="max_retries_exceeded"`.
- New unit tests in `tests/engine/test_ndd_adapter.py` covering each visibility path and each retry path.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing

## Documentation

## Related Issues
Closes #110

### DIFF
--- a/plan/plan.md
+++ b/plan/plan.md
@@ -1,67 +1,72 @@
-# Implementation Plan: Add Visibility and Retry to Silent Data Designer Failures in `NddAdapter`
+# Add Visibility to Silent Failures in `NddAdapter`- Implementation Plan
 
 ## Problem
 
-`NddAdapter` in [src/anonymizer/engine/ndd/adapter.py](src/anonymizer/engine/ndd/adapter.py) wraps every Data Designer call the pipeline makes. Three failure paths are silent today, and a retry mechanism that the existing bookkeeping already enables is missing.
+`NddAdapter` in [src/anonymizer/engine/ndd/adapter.py](src/anonymizer/engine/ndd/adapter.py) wraps every underlying workflow call the pipeline makes. Today the adapter has two classes of log gaps when things go wrong - one that is genuinely silent and one that is already surfaced upstream but hard to investigate locally:
 
-1. **`preview` and `create` exceptions propagate without adapter-level context.** Lines 78-83 (`create` + `load_dataset`) and lines 86-89 (`preview`) are both bare today. A user who hits a dead endpoint, auth error, or rate limit sees a raw DD stack trace with no hint of how many records were being processed or against which model alias.
-2. **`preview_results.dataset is None` silently drops all input records.** Lines 90-91 turn it into `workflow_input_df.iloc[0:0].copy()` with no warning. This branch is defensive - today `data_designer.interface.data_designer.preview()` raises `DataDesignerGenerationError` when the processed dataset is empty rather than returning `PreviewResults(dataset=None)`. The branch still needs to be non-silent so that mocks, test doubles, and any future DD behavior change surface loudly.
-3. **`_detect_missing_records` returns `[]` silently on input short-circuit.** Lines 131-132 bypass detection when the **input** lacks `RECORD_ID_COLUMN` (defensive - `_attach_record_ids` always adds it today). The **output** short-circuit (lines 133-141) does flag every record as failed but still emits no log line. Both short-circuits should warn.
-4. **No retry mechanism** despite `WorkflowRunResult.failed_records` + deterministic `_compute_record_id()` already enabling it. Failed records are tracked, logged, and discarded.
+1. **`preview` and `create` exceptions propagate without adapter-level context.** Lines 78-83 (`create` + `load_dataset`) and lines 86-89 (`preview`) are both bare today. A user who hits a dead endpoint, auth error, or rate limit sees a raw stack trace with no hint of how many records were being processed or against which model alias.
+2. **`_detect_missing_records` short-circuits need different treatment.** The two branches are not equivalent:
+   - **Input-missing (lines 131-132) is a correctness-facing silent drop.** When the **input** lacks `RECORD_ID_COLUMN`, detection returns `[]` - so zero `FailedRecord`s surface upstream even though the adapter can no longer verify whether any rows were dropped. This is a genuinely silent state that warrants a WARNING. It is defensive today (`_attach_record_ids` always adds the column under normal `run_workflow` control flow), so reaching it signals an internal invariant violation.
+   - **Output-missing (lines 133-141) is an observability gap, not a correctness bug.** When the **output** lacks `RECORD_ID_COLUMN`, the adapter already flags every input row as a `FailedRecord`, and upstream anonymizer logging already surfaces aggregate failures - so downstream callers are not blind to the problem. What is missing today is a local breadcrumb at the point of detection explaining **why** every row was marked failed (column pass-through dropped, user column config clash, etc.), which would shave investigation time. Adding a WARNING there is a diagnostic improvement.
 
 ## Scope
+
+This PR is strictly a log-visibility pass. No behavior changes, no retry mechanism, no exception-type changes.
 
 In scope:
 
 - Visibility: wrap both `preview` and `create`+`load_dataset` in `try/except` that emits a user-facing WARNING carrying **row count + model aliases + `type(exc).__name__` + message + actionable hint**, plus a parallel DEBUG line with `workflow_name` and column list. Re-raise unchanged.
-- Visibility: on `preview_results.dataset is None`, emit a user-facing WARNING carrying **row count + model aliases + actionable hint** (DEBUG for `workflow_name`). The existing downstream `_detect_missing_records` call will populate `failed_records` so drops are reflected in `WorkflowRunResult` too.
-- Visibility: in `_detect_missing_records`, emit a user-facing WARNING in **both** short-circuits (input-missing and output-missing) naming the row count; DEBUG line carries `workflow_name` and `RECORD_ID_COLUMN` literal.
+- Visibility: in `_detect_missing_records`, emit a user-facing WARNING in both short-circuits naming the row count, but with deliberately different framing:
+  - Input-missing: framed as a correctness-facing silent drop and an internal invariant violation (new information that is not surfaced today).
+  - Output-missing: framed as an observability/diagnostic improvement that complements the existing upstream aggregate logging, not a correctness fix; the WARNING adds local "why every row was marked failed" context (column diff) at the point of detection.
+
+  DEBUG line in either case carries `workflow_name` and `RECORD_ID_COLUMN` literal.
 - Logging convention: any line a user should read to understand a failure is WARNING with counts/models/error-type baked in; any line that exists to help a developer grep (workflow name, tracking column literal, column list) is DEBUG.
-- Retry: bounded retry loop inside `NddAdapter.run_workflow()` that re-submits rows missing from output until they all appear or `max_retries` is exhausted. Final stragglers get `reason="max_retries_exceeded"`.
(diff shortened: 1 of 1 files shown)


plan/plan.md
```diff
@@ -0,0 +1,235 @@
...
+
+## Problem
+
+`NddAdapter` in [src/anonymizer/engine/ndd/adapter.py](src/anonymizer/engine/ndd/adapter.py) wraps every Data Designer call the pipeline makes. Three failure paths are silent today, and a retry mechanism that the existing bookkeeping already enables is missing.
+
+1. **`preview` and `create` exceptions propagate without adapter-level context.** Lines 78-83 (`create` + `load_dataset`) and lines 86-89 (`preview`) are both bare today. A user who hits a dead endpoint, auth error, or rate limit sees a raw DD stack trace with no hint of how many records were being processed or against which model alias.
+2. **`preview_results.dataset is None` silently drops all input records.** Lines 90-91 turn it into `workflow_input_df.iloc[0:0].copy()` with no warning. This branch is defensive - today `data_designer.interface.data_designer.preview()` raises `DataDesignerGenerationError` when the processed dataset is empty rather than returning `PreviewResults(dataset=None)`. The branch still needs to be non-silent so that mocks, test doubles, and any future DD behavior change surface loudly.
```
lipikaramaswamy: nb: we should always raise an Anonymizer error, not propagate data designer errors

plan/plan.md
```diff
@@ -0,0 +1,235 @@
...
+        raise
+    ...
+```
+
+1b. Replace the silent `None`-dataset branch with a WARNING + DEBUG pair. Drop from the message any mention of future DD versions - we describe the current observable effect and an action:
+
+```python
+if preview_results.dataset is None:
+    logger.warning(
+        "Data Designer preview returned no dataset; %d input record(s) were dropped "
```
lipikaramaswamy: Let's never emit a warning with `Data Designer` - a user should not have to know or care that we use DD under the hood, they should see an anonymizer error.

plan/plan.md
```diff
@@ -0,0 +1,235 @@
...
+
+1. **`preview` and `create` exceptions propagate without adapter-level context.** Lines 78-83 (`create` + `load_dataset`) and lines 86-89 (`preview`) are both bare today. A user who hits a dead endpoint, auth error, or rate limit sees a raw DD stack trace with no hint of how many records were being processed or against which model alias.
+2. **`preview_results.dataset is None` silently drops all input records.** Lines 90-91 turn it into `workflow_input_df.iloc[0:0].copy()` with no warning. This branch is defensive - today `data_designer.interface.data_designer.preview()` raises `DataDesignerGenerationError` when the processed dataset is empty rather than returning `PreviewResults(dataset=None)`. The branch still needs to be non-silent so that mocks, test doubles, and any future DD behavior change surface loudly.
+3. **`_detect_missing_records` returns `[]` silently on input short-circuit.** Lines 131-132 bypass detection when the **input** lacks `RECORD_ID_COLUMN` (defensive - `_attach_record_ids` always adds it today). The **output** short-circuit (lines 133-141) does flag every record as failed but still emits no log line. Both short-circuits should warn.
```
lipikaramaswamy: So _detect_missing_records() does not silently ignore missing output IDs anymore; if the output lacks `RECORD_ID_COLUMN`, it marks every input row as failed at src/anonymizer/engine/ndd/adapter.py:133. So this is a missing warning, not a silent drop.

plan/plan.md
```diff
@@ -0,0 +1,235 @@
...
+        )
+        logger.debug(
+            "NDD workflow '%s' preview failure context: columns=%s",
+            workflow_name, col_names,
+        )
+        raise
+    ...
+```
+
+1b. Replace the silent `None`-dataset branch with a WARNING + DEBUG pair. Drop from the message any mention of future DD versions - we describe the current observable effect and an action:
```
lipikaramaswamy: 1b seems fine as a defensive logging improvement, but I don’t think `dataset=None` is part of Data Designer’s intended `preview()` contract today. In the current implementation, `preview()` raises `DataDesignerGenerationError` when the processed dataset is empty and otherwise returns `PreviewResults(dataset=processed_dataset, ...)`, so this branch looks defensive rather than a real expected runtime path, which is ok too

plan/plan.md
```diff
@@ -0,0 +1,235 @@
...
+    )
+    logger.debug(
+        "NDD workflow '%s' preview dataset=None; emitting empty output", workflow_name,
+    )
+    output_df = workflow_input_df.iloc[0:0].copy()
+```
+
+The downstream `_detect_missing_records` call will then emit a `FailedRecord` per input row, so the drop is visible in `WorkflowRunResult` too.
+
+1c. In `_detect_missing_records`, add a WARNING + DEBUG pair to **both** short-circuits (not only the input case). Each pair also surfaces **why** the tracking column went missing, using the information available at each call site. Counts come from `len(input_df)` since every input row is at risk:
```
lipikaramaswamy: On `1c`, I think the two short-circuit branches should be treated differently.

  - Warning on the `input_df` missing `RECORD_ID_COLUMN` path makes sense. That branch currently returns `[]` silently and indicates an internal invariant violation.
  - I would be more careful with the `output_df` missing `RECORD_ID_COLUMN` path. That case is not fully silent today: the adapter already returns a `FailedRecord` for every input row, and upstream anonymizer logging already surfaces aggregate failures. A warning there could still be useful, but I’d describe it as an observability improvement rather than a correctness fix.

And again, I don’t think anonymizer logs should reference Data Designer directly. If we add warnings here, they should stay at the anonymizer abstraction level, e.g. “workflow output is missing the record-tracking column, so missing-record detection could not be verified,” rather than naming DD or exposing DD-specific behavior.

plan/plan.md
```diff
@@ -0,0 +1,235 @@
...
+        len(input_df),
+    )
+    logger.debug(
+        "NDD workflow '%s' detection skipped: input missing '%s'; input_columns=%s",
+        workflow_name, RECORD_ID_COLUMN, list(input_df.columns),
+    )
+    return []
+```
+
+### Step 2 - Retry loop
```
lipikaramaswamy: On Step 2, I’m not convinced the retry loop belongs in `NddAdapter`.

`NddAdapter` is a generic wrapper used across detection, replacement, and rewrite flows, and those stages do not necessarily want the same retry semantics. Retrying “rows missing from output” at the adapter layer feels like it is imposing a policy across very different workflows rather than fixing the underlying failure mode.

A few concerns:
- It can significantly increase cost and latency by rerunning expensive workflow steps for failed rows.
- It may introduce nondeterministic behavior, since a retry can produce a different output for the same row.
- It risks masking where the failure should actually be handled. If the root problem is transient model/API failure, that usually belongs in the lower-level client/backoff layer, not in the anonymizer adapter.
- Rewriting final failure reasons to `"max_retries_exceeded"` may throw away the original failure context, which is often the most useful debugging signal.

What we need more than retries is a better understanding of why a step failed, i.e. root-cause diagnostics. If we want to make this actionable, we likely need a way to capture and expose DD task traces and processor artifacts, or explicitly defer that work for now.

Today, the practical way to investigate these failures is probably to enable full debug logging and tie each affected record back to its failure log. IMO that's an acceptable short-term approach until the library and its diagnostics surface are more mature.

plan/plan.md
memadi-nv: I'm afraid it might be misleading with what is actually happening under the hood. As you mentioned there's a difference between the time where the dataset is empty (causing the `DataDesignerGenerationError` ) vs when the dataset is `None` (the default value setup for the dataset. [here ](https://github.com/NVIDIA-NeMo/DataDesigner/blob/d75113c7f522ab9e15f586a31e0c4d8adfd39884/packages/data-designer-config/src/data_designer/config/preview_results.py#L23) in DD), so I think it's better to log nothing than an inaccurate one. Seems like this `output_df = workflow_input_df.iloc[0:0].copy()` fallback is effectively dead code anyways! If the preview dataset faces any DD warning, we'll just log that (not in the name of DD of course), wyt?

plan/plan.md
memadi-nv: Considering my thoughts from above, this part wil be removed.

plan/plan.md
memadi-nv: makes sense, agreed! I mainly added the retry loop since it was mentioned as the proposed fix for this issue. If that's the case, I think the logs added in the plan- step 1- should be sufficient. Let me know your thoughts. Thanks!

plan/plan.md
lipikaramaswamy: That sounds good

plan/plan.md
lipikaramaswamy: Yup, good to drop. Btw. issues are often agent drafted so always worth thinking through scope when starting on an issue :)

lipikaramaswamy:
Good to start implementation with the adjustment to scope 👍

PR #130, opened 2026-04-22
fix: improved small-model support

mvansegbroeck:
## Summary

Anonymizer's rewrite pipeline was dropping records on every non-`gpt-oss-120b` model due to schema strictness that small models couldn't satisfy. Root cause: DataDesigner runs `jsonschema.validate()` on raw LLM output *before* pydantic's before-validators run — any strict `enum`, `required`, `type`, `minLength`, or uniqueness check becomes an un-coercible gate for small-model drift.

This PR applies a two-layer pattern to every LLM-facing schema that was dropping records:

1. **Loose wire schema**: typed as `str` (not Enum), defaults instead of `required`, nullable where useful, no strict lengths. The emitted JSON Schema stops rejecting drifted output. For some shapes (disposition / meaning units) the wire schema is widened to `oneOf({wrapper}, {array})` via `__get_pydantic_json_schema__` so DD's pre-validate accepts both shapes.
2. **Server-side reconstruction / before-validators normalize drift**: coerces None/int → str, lowercases enum drift, substring-matches merged enums, dedups IDs, pads missing IDs, truncates overlong / coerces empty strings, reconciles inconsistent field pairs. Disposition uses the reconstructor as the single normalization gate; remaining schemas use before-validators. The reconstructor column wraps construction in `try/except ValidationError` so an orphans-only response emits an empty disposition rather than dropping the row.

## Schemas touched

| schema | role | change |
|---|---|---|
| `EntityDispositionSchema` | disposition_analyzer | **Architecture: strict internal schema, reconstructor as single gate.** Wire schema is the loose `SimpleDispositionItem` (`oneOf` widening accepts bare-list shape). `disposition_derivation.reconstruct_full_disposition` builds the strict `EntityDispositionSchema` server-side via `_normalize_category` + `_normalize_method` (both case-fold + substring fallback) + `derive_needs_protection` (tautologically consistent (needs, method) pair). `category` restored to `EntityCategory` enum; `_coerce_small_model_output` deleted; `combined_risk_level` deleted. Pessimistic `needs_protection` default for high-risk entities with omitted method. Reconstructor column guards `ValidationError` from orphans-only responses. |
| `SimpleDispositionResult` | disposition_analyzer (wire) | New schema. Wire shape widened to `oneOf({wrapper-object}, {bare-array})` so models that omit the `sensitivity_disposition` key (observed: nemotron-3-nano:4b, 5/5 records) still pass DD's jsonschema pre-validate. Before-validator normalizes the bare-list shape back to wrapper. |
| `MeaningUnitsSchema` | meaning_extractor | Same `oneOf` + `_accept_bare_list` widening (qwen3.5:4b legal_court drift). `_ensure_list` reassigns sequential ids 1..N when LLM omits ids or emits collisions. |
| `RawValidationDecisionSchema` | chunked-validation path | Adds `_normalize_decision` (preserves None for the chunk-merger's "no answer, skip" semantics; prose drift → `"keep"`) and `_coerce_proposed_label` (None → `""`, int → str). Mirrors `ValidationDecisionSchema`'s coercers which existed only on the non-chunked wire path. |
| `ValidationDecisionSchema` | validator (wire) | `decision` enum → str; `proposed_label` nullable; drop dead `value`/`label` (already server-overridden in `enrich_validation_decisions`). |
| `LatentEntitySchema` | latent_detector | category/confidence enums → str; `rationale` truncation + empty default; `evidence` list clamp; sentinel injected on empty list to keep pyarrow happy. |
| `MeaningUnitSchema` | meaning_extractor | 16-value `aspect` enum → str + normalize. |
| `DomainClassificationSchema` | domain_classifier | 24-value `Domain` enum → str + normalize; string-to-float confidence coercion (accepts `"0.95"`, `"85%"`, falls back to 0.5 for `"high"`). |
| `QualityAnswersSchema` / `PrivacyAnswersSchema` / `QACompareResultsSchema` | evaluator / judge | `@model_validator(mode="before")` dedups IDs, pads missing with pessimistic defaults, drops extras. `PrivacyAnswerItemSchema.reason` truncates at 200 chars and now also coerces empty/whitespace-only strings (was rejected by `min_length=1`). |

`_ENTITY_LABEL_TO_CATEGORY` is derived from three frozensets (`_DIRECT_ID_LABELS`, `_QUASI_ID_LABELS`, `_SENSITIVE_ATTR_LABELS`); a CI-asserted coverage test guards against future divergence from `DEFAULT_ENTITY_LABELS`.

## Drift modes uncovered post-merge (round-2 fixes)

The initial PR was reviewed and revised (commits 19efe6e..16bff86). After merging origin/main (chunked-validation feature #126) and re-benching against 5 candidate models on DGX Spark, four additional drift modes surfaced. Each is now covered:

| Drift mode | Affected schema | Fix commit | Result |
|---|---|---|---|
| Bare list at top level instead of `{sensitivity_disposition: [...]}` | `SimpleDispositionResult` | `dc0bc84` | nemotron-3-nano:4b rewrite 0/5 → 5/5 |
| `proposed_label: null` from chunked-validation path | `RawValidationDecisionSchema` | `576de6a` | gemma4-e4b redact 1/5 → 4/5 |
| Free-form prose in `decision` field (`"No specific entity type..."`) | `RawValidationDecisionSchema` | `5fa19b7` | gemma4-e4b 4/5 → 5/5 |
| Bare list at top level instead of `{units: [...]}` | `MeaningUnitsSchema` | `329d488` | qwen3.5:4b legal_court rewrite ❌ → ✅ |

Round-2 hardening (commit `5e55aea`) closes greptile's automated review:

* Lowercase + substring-match `protection_method_suggestion` before strict-schema construction (`"REPLACE"` no longer drops record)
* Empty-string `_truncate_reason` coerces to placeholder (was failing `min_length=1`)
* `_reconstruct_full_disposition_column` guards `ValidationError` from orphans-only responses
* Drop dead `endswith("_identifiers")` branch in `_normalize_category`

## Bench results — 5 candidate models on DGX Spark (Ollama Q4_K_M, post-fix)

| Model | Replace | Rewrite | Mean rewrite wall | Mean utility | Mean leakage |
|---|---|---|---|---|---|
| `gemma4-e2b` (2B) | 5/5 ✅ | 5/5 ✅ | 92 s | 0.81 | 0.13 |
| `nemotron-3-nano:4b` | 5/5 ✅ | 5/5 ✅ | 164 s | 0.63 | 0.28 |
| **`gemma4-e4b`** ⭐ | **5/5 ✅** | **5/5 ✅** | 199 s | **0.92** | 0.13 |
| `qwen3.5:4b` | 5/5 ✅ | 5/5 ✅ | 204 s | 0.69 | 0.19 |
| `qwen3.5:9b` | 5/5 ✅ | 5/5 ✅ | 256 s | 0.82 | **0.04** |

**Recommended sweet spot for rewrite: `gemma4:e4b`** — best small-model utility, 5/5 reliable, mid-tier latency. `gemma4:e2b` is the speed pick for replace mode (~21s avg).

NVFP4 / vLLM second-half was bench-tested (separate commit set, on cosmicproc/E4B + prithivMLmods/E2B community quants). Result: **no Blackwell speedup observed at single-stream batch with community W4A4 quants** — both gemma4 sizes were equal-or-slower than Ollama Q4_K_M with -14 to -22pp utility drop. Documented as future work; not in this PR's scope.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactoring

## Testing
- [x] `make test` passes locally — **514 tests pass** (108 new since the original PR open: schema coverage, label coverage, privacy reason truncation including empty-string, domain confidence string coercion, MeaningUnit id renumbering, `_normalize_category` + `_normalize_method` drift modes, class-K reconstructor reconciliation, `_flatten_context` enumeration alignment, `_pad_empty_latent_column`, end-to-end `enrich_validation_decisions` with dropped wire fields, bare-list shapes for both disposition and meaning, `RawValidationDecisionSchema` None/prose/int coercers, reconstruct-column orphans-only guard)
- [x] `make check` passes locally (format + lint + typecheck)

## Documentation
- [ ] If docs changed: `make docs-build` passes locally — no doc changes in this PR

## Review response

**@memadi-nv (round 1, 2026-04-24)**: 7 inline comments + 6 missing-test gaps. All addressed by commits 19efe6e..16bff86. Brief mapping in commit messages.

**greptile-apps[bot] (rounds 1-2)**: 3 P1 + 1 P2 from the round-2 review on commit 329d488. All closed by commit `5e55aea`.

**Round-2 reviewers (Amy / Andre)**: Andre's engineering nits (`_normalize_method` lowercase, dead branch, ValidationError guard, empty-string `_truncate_reason`, cross-ref comment) all addressed. Amy's prompt-restoration concerns (COVERAGE / QUALITY clauses dropped from disposition prompt) and the requested A/B utility comparison are noted as follow-up — recommended for a separate PR with isolated bench evidence rather than reverting prompts that bench shows produce 0.92 utility on the headline model.

## Pattern shared via helper (commit `4a54b85`)

The two schemas (`SimpleDispositionResult`, `MeaningUnitsSchema`) that need bare-list tolerance now share a single helper pair in `engine/schemas/shared.py`:

* `loose_list_wrapper_json_schema(handler, schema, *, list_field)` — emits `oneOf({wrapper}, {array})`, gracefully degrades + logs a warning if pydantic ever moves the inline property schema behind a `$ref` (defensive against future restructure).
* `accept_bare_list(*, list_field)` — builds the `mode="before"` validator that wraps a top-level bare list back into the canonical wrapper dict.

Each call site is now one line each. Future loose-wire schemas drop in trivially.

Also in `4a54b85`: `RawValidationDecisionSchema._normalize_decision` renamed to `_normalize_decision_preserve_none` so the divergence vs the canonical `ValidationDecisionSchema._normalize_decision` (which returns "keep" instead of preserving `None`) is visible at call sites without requiring docstring comparison.

## Open follow-up

Amy's prompt-restoration request (re-add COVERAGE / QUALITY clauses to the disposition prompt + run isolated A/B utility comparison) is deferred to a separate PR. Aggregate bench evidence on this branch shows util preserved (0.92 mean on `gemma4:e4b` headline; 5/5 reliable across all 5 candidate models). The follow-up PR will pair restored clauses with side-by-side bench evidence to isolate the utility delta.

## Related

Analysis trail and full benchmark matrix are documented externally in `notes/lipika-heads-up-local-serving.md` §5/§5b in the `anonymizer-strategy` workspace.


src/anonymizer/engine/schemas/rewrite.py
```diff
@@ -114,10 +157,87 @@ class ProtectionMethod(str, Enum):
...
 
 
-class CombinedRiskLevel(str, Enum):
-    low = "low"
-    medium = "medium"
-    high = "high"
+# Best-effort mapping from Anonymizer entity-labels to EntityCategory, used when
+# the model outputs an entity_label in the `category` field (observed with small
+# Gemma models on the disposition step). Values are from DEFAULT_ENTITY_LABELS in
+# config/entity_labels.py; any label not in this table falls back to
```
memadi-nv: There is no `config/entity_labels.py`. `DEFAULT_ENTITY_LABELS` lives at `src/anonymizer/engine/constants.py`.

src/anonymizer/engine/schemas/rewrite.py
```diff
@@ -114,10 +157,87 @@ class ProtectionMethod(str, Enum):
...
+# "quasi_identifier" which is the most conservative (protect-cautiously) choice.
+_ENTITY_LABEL_TO_CATEGORY: dict[str, str] = {
+    # Direct identifiers: strong re-id on their own.
```
memadi-nv: ### `_ENTITY_LABEL_TO_CATEGORY` is a hand-maintained copy and is already out of sync

The PR's own comment calls `DEFAULT_ENTITY_LABELS` the source of truth, but the table in `schemas/rewrite.py` contains labels that do not exist in `ENTITY_LABEL_EXAMPLES` — e.g. `full_name`, `passport_number`, `drivers_license`, `address`, `zip_code`, `credit_card_number`. Conversely it likely omits labels someone will add tomorrow.

Two reasonable options:

- (preferred) Define three small sets `_DIRECT_IDS`, `_QUASI_IDS`, `_SENSITIVE_ATTRS` and derive `_ENTITY_LABEL_TO_CATEGORY` as `{lbl: "direct_identifier" for lbl in _DIRECT_IDS} | ...`. Then add a test: `assert set(DEFAULT_ENTITY_LABELS) <= set(_ENTITY_LABEL_TO_CATEGORY)`.
- At minimum add the coverage assertion as a unit test so the next label-addition PR breaks CI rather than silently losing fallback coverage.

src/anonymizer/engine/rewrite/disposition_derivation.py
```diff
@@ -0,0 +1,269 @@
...
+        if not src or not lbl or not val:
+            logger.warning(
+                "reconstruct_full_disposition: orphan simple item id=%s "
+                "(missing or drifted source/label/value, out of context range); skipping",
+                item.id,
+            )
+            continue
+
+        # Derive derived fields.
+        method = (item.protection_method_suggestion or "").strip() or "leave_as_is"
```
memadi-nv: When the small model omits the method, this silently derives `needs_protection=False,` even for a `direct_identifier` with `sensitivity="high`". That's exactly the kind of failure mode this PR exists to fix — and it's inconsistent with the pessimistic defaults the PR uses elsewhere (e.g. `PrivacyAnswersSchema `pads missing with `"yes"`, `_normalize_decision` defaults to `"keep"` = preserve detection).

**Suggested fix** — pessimistic default based on category/sensitivity:

```suggestion
raw_method = (item.protection_method_suggestion or "").strip()
if raw_method:
    method = raw_method
elif category in ("direct_identifier", "sensitive_attribute") or sensitivity in ("medium", "high"):
    method = "replace"
else:
    method = "leave_as_is"
```
Add a test for `SimpleDispositionItem(id=1, category="direct_identifier", sensitivity="high", protection_method_suggestion="") `→ reconstructed with `needs_protection=True`.

src/anonymizer/engine/schemas/rewrite.py
```diff
@@ -129,16 +249,134 @@ class EntityDispositionSchema(BaseModel):
...
 
+    # Coerce small-model output before pydantic's strict checks run:
+    #   - normalize `category` display-label drift (e.g. "DIRECT IDENTIFIERS"
+    #     or "last_name") into the expected enum string where possible
+    #     (common Gemma-family failure mode).
+    #   - coerce unquoted ints on string fields (Gemma 4 26B emits postcodes
+    #     like 98101 as bare integers).
+    @model_validator(mode="before")
+    @classmethod
+    def _coerce_small_model_output(cls, data):
```
memadi-nv: ### Two layers of coercion for the same drift — duplicated logic, only one is reachable in prod

`EntityDispositionSchema._coerce_small_model_output` does ~70 lines of normalization (category drift, int→str, needs↔method reconciliation, etc.) — but since the workflow now routes every disposition through `SimpleDispositionResult` + `reconstruct_full_disposition`, the reconstructor always pre-massages the fields before calling `EntityDispositionSchema(...)`. The strict-schema's before-validator never sees raw LLM output in the production path; it only runs in the unit tests.

Two downsides:

- ~70 lines of production-untested code billed as a live safety net.
- `template_protection_reason` vs `protection_reason` default (`"auto: model omitted protection_reason"`) are two different fallbacks for the same field, reachable via different paths. Currently the reconstructor path wins in prod; the stringly-typed default only kicks in if someone constructs `EntityDispositionSchema` directly from LLM output — a combination no one should be doing anymore.

Pick one gate:

- (a) Drop `_coerce_small_model_output` from `EntityDispositionSchema`, make `category: EntityCategory` again, rely exclusively on the reconstructor. Simpler, one source of truth.
- (b) Keep `_coerce_small_model_output` as the single unified gate, remove the templating/defaulting from `reconstruct_full_disposition`, and route every item through `EntityDispositionSchema.model_validate`.

(a) seems like a better pick. Reason: the reconstructor has access to context (category, sensitivity, method together) that the field-by-field coercion doesn't, so its reason templating is strictly better than the static default.

tests/engine/test_disposition_derivation.py
```diff
@@ -0,0 +1,270 @@
...
+    # SensitivityDispositionSchema requires min_length=1, so zero-item
+    # reconstruction should raise at the strict-schema layer — caller is
+    # expected to short-circuit before disposition when no entities exist.
+    simple = SimpleDispositionResult(sensitivity_disposition=[])
+    with pytest.raises(ValidationError):
+        reconstruct_full_disposition(simple, [], [])
+
+
+def test_multi_label_entity_flattens_to_multiple_slots() -> None:
+    simple = _make_simple(
```
memadi-nv, line 388: ### Prompt enumeration order must match `_flatten_context` — not verified

`reconstruct_full_disposition` pairs LLM items with context by `context[item.id - 1]`, and `_flatten_context` emits **one slot per `(value, label)` pair**. If the jinja rendering of `<<FINAL_ENTITIES>>` presents entities to the LLM as **one row per value** (multi-label values merged), the LLM's numbering will not align with the context index — multi-label entities will either get the wrong pairing or fall into the orphan path and be silently dropped.

The new reconstructor has a unit test (`test_multi_label_entity_flattens_to_multiple_slots`), but there is no test asserting the **prompt rendering** enumerates in the same order and cardinality.

Please add an end-to-end test that renders the prompt for a multi-label `entities_by_value` row, asserts the entity list the model sees is numbered `1..N` identically to `_flatten_context`, and feeds a matching hand-crafted LLM response through `reconstruct_full_disposition` with zero orphans. If the rendering turns out to be one-per-value, `_flatten_context` should collapse to match.

src/anonymizer/engine/schemas/rewrite.py
```diff
@@ -288,16 +531,64 @@ class MeaningUnitAspect(str, Enum):
...
 
 class MeaningUnitsSchema(BaseModel):
-    """LLM output schema for meaning unit extraction step."""
+    """LLM output schema for meaning unit extraction step.
+
+    Outer list has default_factory=list (was min_length=1); if the model
+    emits an empty list the record still survives so the pipeline can
+    decide what to do downstream. A @model_validator(mode="before") drops
+    units with empty unit text to reduce noise.
+    """
```
memadi-nv, line 583: ###  `MeaningUnitsSchema` docstring describes behavior that isn't implemented

The docstring claims *"A `@model_validator(mode="before")` drops units with empty unit text to reduce noise"*, but no such validator exists — only `_ensure_list` (list-wrap). Either implement the drop or fix the docstring.

Related: `MeaningUnitSchema.id: int = Field(ge=1, default=1)` — if the LLM emits multiple units whose ids all default to `1`, any downstream id-keyed dedup collapses them. Consider `default=None` with sequential reassignment in `_ensure_list`, or document that the `default=1` is purely a wire-layer escape hatch.

src/anonymizer/engine/schemas/detection.py
```diff
@@ -138,34 +187,146 @@ class LatentConfidence(str, Enum):
...
+        return out
+
+    @field_validator("rationale", mode="before")
+    @classmethod
+    def _cap_rationale(cls, v: object) -> str:
+        """Truncate verbose (Nemotron-observed 260 chars > 150 max); pad
+        terse (Qwen-observed sub-10 chars) with an ellipsis to reach the
+        10-char soft floor. Exact floor kept as internal expectation but
+        not enforced at wire — server reads rationale for context only.
+        """
```
memadi-nv, line 368: Only the upper bound is enforced. Either implement the padding or drop the promise from the docstring — non-blocking (wire no longer has `min_length`) but confusing for whoever reads this next.

memadi-nv (commented):
### Missing tests for claims in the PR summary / commits

- PR summary: *"5 stale class-K assertions refreshed to assert reconciliation instead of raising"* — the diff only shows 2 refreshed tests (`test_entity_disposition_reconciles_no_protection_but_method_set`, `test_entity_disposition_reconciles_needs_protection_but_leave_as_is`). Wcould not find the other 3!
- No test for `_pad_empty_latent_column` in `detection_workflow.py` (the belt-and-braces DataFrame sentinel).
- No test exercising the entity-label-stuffed-into-category path in `_coerce_small_model_output` via `_ENTITY_LABEL_TO_CATEGORY` — this was called out as a key Gemma failure and gets its own ~70 lines of table, but tests only cover display-label variants (`"DIRECT IDENTIFIER"`, plural, `"_identifiers"`), not the `"last_name"` → `direct_identifier` mapping.
- No test that `ValidationDecisionSchema` dropping `value`/`label` from the wire still produces correct output through `enrich_validation_decisions` (the PR claims those fields were already server-overridden — worth an end-to-end assertion).
- No test for `PrivacyAnswerItemSchema._truncate_reason` (200+ char inputs, the specific small-model drift the method exists for).
- No test for `DomainClassificationSchema._coerce_confidence` string inputs (`"0.95"`, `"85%"`, `"high"`), despite `qa_generation.py`'s `_DOMAIN_KEY` lookup being downstream of this.

mvansegbroeck:

src/anonymizer/engine/rewrite/sensitivity_disposition.py
```diff
@@ -187,6 +203,67 @@ def _get_sensitivity_disposition_prompt(
...
+    # entity_label/value echoes (orphans-only response), the reconstructor
+    # produces an empty list, which trips
+    # SensitivityDispositionSchema.min_length=1 and raises ValidationError.
+    # That exception would propagate from this generator and drop the
+    # record — exactly the failure mode this PR is meant to prevent. Catch
+    # and emit an empty disposition column instead so downstream consumers
+    # treat it as "no entities to protect" rather than dropping the row.
+    if not simple.sensitivity_disposition:
+        logger.warning("reconstruct: empty SimpleDispositionResult for row; emitting empty sensitivity_disposition")
+        row[COL_SENSITIVITY_DISPOSITION] = {"sensitivity_disposition": []}
```
lipikaramaswamy, line 247: Downstream parsing still routes through `SensitivityDispositionSchema.model_validate(...)` in [parse_sensitivity_disposition()](https://github.com/NVIDIA-NeMo/Anonymizer/blob/maarten/bugfix/small-model-support/src/anonymizer/engine/rewrite/parsers.py#L120), so I think this guard will just cause a failure later when the QA and rewrite steps call this function.

I also think that in the case of missing sensitivity disposition, we should be forcing all direct identifiers to be replaced and all quasi identifiers to be generalized - that's the rough guidance in the sensitivity disposition prompt too (cc @asteier2026)

src/anonymizer/engine/schemas/rewrite.py
```diff
@@ -115,11 +160,112 @@ class ProtectionMethod(str, Enum):
...
 
+# Mapping from Anonymizer entity-labels to EntityCategory, used when the
+# disposition LLM outputs an entity_label in the `category` field (observed
+# with small Gemma models). Derived from three category sets so the source
+# of truth lives in one place per category — see
+# test_entity_label_to_category_covers_default_labels for the CI guard
+# that fires when a label is added to DEFAULT_ENTITY_LABELS without a
+# category assignment here. Any label not in this table falls back to
+# "quasi_identifier" which is the most conservative (protect-cautiously)
+# choice.
```
lipikaramaswamy, line 183: Two things I need to think more about in case the LLM at the sensitivity disposition step fails to generate output (and @asteier2026 probably knows more here) 

1. Is this list of direct identifiers, quasi and sensitive attrs right? 
2. Do we want any user supplied label to just be generalized instead of replaced?

src/anonymizer/engine/rewrite/sensitivity_disposition.py
asteier2026: That would be optimal, I agree with "in the case of missing sensitivity disposition, we should be forcing all direct identifiers to be replaced and all quasi identifiers to be generalized "

PR #132, opened 2026-04-22
docs: self-hosting GLiNER locally

mvansegbroeck:
## Summary

Add `docs/concepts/self-hosting-gliner.md` — a Concepts page on self-hosting `nvidia/gliner-pii` locally for Anonymizer's `entity_detector` role.

Covers the OpenAI-compatible request/response contract, a minimal FastAPI reference server (CPU by default, auto-uses GPU when available), two `curl` examples (one checks `/v1/models`, one
 exercises the full detection contract and shows the decoded `entities` array), a minimal provider YAML with the `skip_health_check: true` footnote, and performance notes (CPU 5–20
ms/note, GPU 1–3 ms at ~1 GB VRAM).

## Type of Change
- [x] Documentation update

## Testing
N/A — docs-only.

## Documentation
- [x] `make docs-build` passes locally

## Related Issues
None.


docs/concepts/self-hosting-gliner.md
```diff
@@ -0,0 +1,171 @@
...
+DEVICE = "cuda" if torch.cuda.is_available() else "cpu"
+app = FastAPI()
+model = GLiNER.from_pretrained(MODEL_NAME).to(DEVICE)
+
+@app.get("/v1/models")
+def list_models():
+    return {"object": "list", "data": [{"id": MODEL_NAME, "object": "model"}]}
+
+@app.post("/v1/chat/completions")
+async def chat_completions(request: Request):
```
lipikaramaswamy, line 62: This matches Anonymizer’s parser, but it currently ignores `chunk_length`, `overlap`, and `flat_ner`, even though Anonymizer sends those on the detector path. That can make long-text behavior diverge from the expected GLiNER setup.The [Guardrails implementation](https://github.com/NVIDIA-NeMo/Guardrails/blob/a5c765899d3a407bf076f0226d9b784324e86eda/examples/deployment/gliner_server/src/gliner_server/server.py) is a useful reference for chunking/overlap handling, but it still needs an Anonymizer-specific adapter because this chat endpoint must return entities as `text` / `label` / `start` / `end` / `score` inside `message.content`.

mlcrubio:
looking forward to get this merged.

docs/concepts/self-hosting-gliner.md
lipikaramaswamy: The reference server now honors all three fields from the detector request (chunk_length, overlap, flat_ner) via _create_text_chunks, offset shifting, and _finalize_entities (subset removal when flat_ner=false). See `tools/serve_gliner.py`

tools/gliner_pii_utils.py
```diff
@@ -0,0 +1,94 @@
...
+def create_text_chunks(text: str, chunk_length: int = 384, overlap: int = 128) -> tuple[list[str], list[int]]:
+    """Split text into overlapping chunks (character-based, same as Guardrails)."""
+    chunks: list[str] = []
+    offsets: list[int] = []
+    start = 0
+    while start < len(text):
+        chunks.append(text[start : start + chunk_length])
+        offsets.append(start)
+        start += chunk_length - overlap
+    return chunks, offsets
```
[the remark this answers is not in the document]
lipikaramaswamy: Fixed. That file is no longer in the PR; logic lives in `tools/serve_gliner.py`. `_validate_chunk_params` rejects invalid pairs with HTTP 422 (overlap must be less than `chunk_length`) before chunking runs. See L104-L111.

tools/serve_gliner.py
```diff
@@ -0,0 +1,400 @@
...
+        "model_name": MODEL_NAME,
+        "device": _device,
+        "batch_mode": BATCH_MODE,
+    }
+
+
+@app.post("/v1/chat/completions")
+async def chat_completions(request: Request) -> dict[str, Any]:
+    if model is None:
+        raise RuntimeError("GLiNER model is not loaded")
```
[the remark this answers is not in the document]
lipikaramaswamy: Fixed — chat_completions raises `HTTPException(status_code=503, detail="GLiNER model is not loaded")` when `model is None`

tools/serve_gliner.py
```diff
@@ -0,0 +1,400 @@
...
+                if not job.future.done():
+                    job.future.set_result(entities)
+
+
+detector = BatchDetector()
+
+
+@app.on_event("startup")
+async def startup() -> None:
+    _load_model()
```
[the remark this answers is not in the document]
lipikaramaswamy: Fixed — startup uses `await asyncio.to_thread(_load_model)` so model download/load doesn’t block the uvicorn event loop.

docs/concepts/self-hosting-gliner.md
```diff
@@ -0,0 +1,177 @@
...
+    inference_parameters:
+      max_parallel_requests: 8   # send concurrent rows; the reference server batches them
+```
+
+Set `skip_health_check: true`: Anonymizer's default probe sends `prompt="Hello!"` with no `labels` field, which is not a valid GLiNER request.
+
+---
+
+## Performance notes
+
```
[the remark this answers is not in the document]
lipikaramaswamy: Fixed — Performance notes now document DEVICE (auto, cuda, mps, cpu), e.g. `DEVICE=cuda python tools/serve_gliner.py`, matching `_resolve_device()` in the reference server. See the “Performance notes” section at the bottom of `docs/concepts/self-hosting-gliner.md`.

tools/serve_gliner.py
```diff
@@ -0,0 +1,459 @@
...
+def _dedupe_entities(entities: list[dict[str, Any]]) -> list[dict[str, Any]]:
+    best: dict[tuple[str, str], dict[str, Any]] = {}
+    for entity in entities:
+        label = str(entity.get("label", ""))
+        text = str(entity.get("text", ""))
+        key = (label, text.strip().lower())
+        score = float(entity.get("score", 0.0))
+        if key not in best or score > float(best[key].get("score", 0.0)):
+            best[key] = entity
+    return list(best.values())
```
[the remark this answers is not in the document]
lipikaramaswamy: Fixed in 44b3e45. Dedupe now keys on (label, text, start, end) after chunk offsets are shifted to document-global coordinates, so overlap-window duplicates collapse but the same surface form at different positions is kept. Intent: dedupe redundant detections from overlapping chunks, not repeated mentions elsewhere in the doc.

tools/serve_gliner.py
```diff
@@ -0,0 +1,462 @@
...
+from typing import Any
+
+import torch
+import uvicorn
+from fastapi import FastAPI, HTTPException, Request
+from gliner import GLiNER
+
+MODEL_NAME = "nvidia/gliner-pii"
+HOST = "0.0.0.0"
+PORT = 8001
```
mvansegbroeck: should we make this an argument with 8001 the default?

tools/serve_gliner.py
```diff
@@ -0,0 +1,462 @@
...
+DEFAULT_CHUNK_LENGTH = 384
+DEFAULT_OVERLAP = 128
+DEFAULT_FLAT_NER = False
+DEFAULT_INFERENCE_BATCH_SIZE = 8
```
mvansegbroeck, line 46: just double checking if these are the right defaults here.

lipikaramaswamy: Yup 384, 128, flat_ner=false match what Anonymizer injects on the detector path `detection_workflow._inject_detector_params`. `DEFAULT_INFERENCE_BATCH_SIZE=8` is server-side only and used when the request doesn’t pass batch_size.

tools/serve_gliner.py
lipikaramaswamy: Good call, added `--port` / `--host` CLI override with the same defaults, though it's not explicitly reqd for anonymizer wiring.

tools/serve_gliner.py
```diff
@@ -0,0 +1,481 @@
...
+from dataclasses import dataclass
+from typing import Any
+
+import torch
+import uvicorn
+from fastapi import FastAPI, HTTPException, Request
+from gliner import GLiNER
+
+MODEL_NAME = "nvidia/gliner-pii"
+DEFAULT_HOST = "0.0.0.0"
```
memadi-nv: **Default bind 0.0.0.0 contradicts the data-residency premise** — serve_gliner.py:40, DEFAULT_HOST = "0.0.0.0", no auth.
The main point of this PR is "PHI never leaves the host." But binding 0.0.0.0 exposes the detector — and all the text flowing through it — on every network interface, so anyone on the LAN can reach it. For a tool whose reason to exist is data residency, the safe default is 127.0.0.1 (localhost-only), with 0.0.0.0 as an opt-in for the multi-host case. At minimum the docs should call out the exposure if we decide to leave it as is.

memadi-nv (approved):
Tested with a couple of entity text rows and all works great! Just a small feedback for the default bind that makes the detector network-reachable with no auth and is better to be changed for the purpose of this PR.

lipikaramaswamy (approved):

PR #136, opened 2026-04-24
fix: add visibility to silent failures in NDD adapter

### LINKED ISSUE #110: bug: Data Designer failures silently drop records — add visibility and retry

memadi-nv:
## Summary

Surface previously silent failure modes in `NddAdapter` via structured logs, and introduce a typed error wrapper so callers don't depend on backend exception types.

### Changes
- **`src/anonymizer/engine/ndd/adapter.py`**
  - Wrap `create()` / `load_dataset()` / `preview()` in a `try/except` that logs a `WARNING` (row count, model aliases, exception) + `DEBUG` (workflow name, columns) and re-raises as `AnonymizerWorkflowError` with `__cause__` preserved.
  - Add `WARNING` + `DEBUG` logs for the two silent short-circuits in `_detect_missing_records` (input missing tracking column, output missing tracking column, with a column-diff breadcrumb).
- **`src/anonymizer/interface/errors.py`**
  - New `AnonymizerWorkflowError` for wrapping workflow-step failures.
- **`tests/engine/test_ndd_adapter.py`**
  - Five new tests covering the three exception-wrap paths and the two detection short-circuit warnings, plus an assertion that log messages don't leak backend product names.

## Type of Change
- [x] Bug fix

## Testing
- [x] `make check` passes locally

## Related Issues
Plan for this issue created in draft PR #128 => some parts partially changed
Closes #110

### DIFF
--- a/src/anonymizer/engine/ndd/adapter.py
+++ b/src/anonymizer/engine/ndd/adapter.py
@@ -153,8 +153,7 @@ class NddAdapter:
             logger.warning(
                 "Missing-record detection skipped: input DataFrame lacks the record-tracking "
                 "column, so the adapter cannot verify whether any of %d input record(s) were "
-                "dropped. This indicates an internal invariant violation - `_attach_record_ids` "
-                "was not called on this DataFrame before detection.",
+                "dropped.",
                 len(input_df),
             )
             logger.debug(
--- a/src/anonymizer/engine/ndd/adapter.py
+++ b/src/anonymizer/engine/ndd/adapter.py
@@ -64,6 +64,7 @@ class NddAdapter:
         logger.debug("NDD workflow '%s' starting with %d records", workflow_name, len(workflow_input_df))
         col_names = [c.name for c in columns]
         logger.debug("NDD workflow '%s': %d columns %s", workflow_name, len(col_names), col_names)
+        model_aliases = [m.alias for m in model_configs]
 
         with tempfile.TemporaryDirectory(prefix=f"anonymizer_{workflow_name}_") as tmp_dir:
             seed_path = str(Path(tmp_dir) / "seed.parquet")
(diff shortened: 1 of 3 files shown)


src/anonymizer/engine/ndd/adapter.py
```diff
@@ -75,18 +76,50 @@ def run_workflow(
...
+                    output_df = run_results.load_dataset()
+                except Exception as exc:
+                    logger.warning(
+                        "Workflow execution failed for %d input record(s) on model(s) %s: %s: %s. "
+                        "Check endpoint reachability, credentials, and quota.",
+                        len(workflow_input_df),
+                        model_aliases,
+                        type(exc).__name__,
+                        exc,
+                    )
```
lipikaramaswamy: We can keep the useful DD error detail without exposing `DataDesignerError` (or other backend exception types) directly by wrapping it in an anonymizer-owned exception and chaining the cause.

Something along the lines of:

  ```python
  except Exception as exc:
      logger.warning("Workflow execution failed: %s", exc)
      raise AnonymizerError(f"Workflow execution failed: {exc}") from exc
```

Can add a new error type like `AnonymizerExecutionError` or `AnonymizerWorkflowError` in src/anonymizer/interface/errors.py and use that instead of generic `AnonymizerError`. That way we won't be presenting a very broad surface like endpoint reachability, credentials, quota... we'll be relaying the exact error.

File: src/anonymizer/engine/ndd/adapter.py [hunk 2]
```diff
@@ -74,23 +77,41 @@ def run_workflow(
...
+            except Exception as exc:
+                logger.warning(
+                    "Workflow failed for %d input record(s) on model(s) %s: %s",
+                    record_count,
+                    model_aliases,
+                    exc,
                 )
-                output_df = run_results.load_dataset()
-            else:
-                effective_preview = min(preview_num_records, len(workflow_input_df))
```
lipikaramaswamy, line 85: let's not lose this logic

lipikaramaswamy: nvm i see it above, though

lipikaramaswamy (approved):
Looks good to approve. Were you able to check using a small model or config with very low timeout to simulate a data designer failure?

memadi-nv:
> Were you able to check using a small model or config with very low timeout to simulate a data designer failure?

Instead of driving an actual low-timeout model, the tests added in  `tests/engine/test_ndd_adapter.py` mock `DataDesigner` and inject `side_effect` exceptions at each of the three points where the backend can fail:

1. **`preview` failure** — `test_preview_exception_wraps_in_workflow_error_and_logs`
   - `mock_dd.preview.side_effect = DataDesignerRuntimeError("endpoint unreachable")`
2. **`create` failure** — `test_create_exception_wraps_in_workflow_error_and_logs`
   - `mock_dd.create.side_effect = DataDesignerRuntimeError("quota exceeded")`
3. **`load_dataset` failure (post-create)** — `test_load_dataset_exception_wraps_in_workflow_error_and_logs`
   - `mock_create_results.load_dataset.side_effect = DataDesignerRuntimeError("corrupt parquet")`

Each test asserts that:
- The failure is wrapped in `AnonymizerWorkflowError` with the original `DataDesigner` exception preserved as `__cause__`.
- The warning log includes the record count, model alias, and error message.
- The debug log includes the workflow context (columns, workflow name).
- Neither log leaks backend identifiers (`Data Designer`, `DataDesigner`, etc.).

I assume that's sufficient, but let me know if it's not the case.

PR #137, opened 2026-04-27
feat: enhance repair

asteier2026:
Changes include:
- Add evidence to privacy qa reanswer
- Enhance privacy qa reanswer prompt
- Enhance repair prompt

### DIFF
--- a/src/anonymizer/engine/rewrite/evaluate.py
+++ b/src/anonymizer/engine/rewrite/evaluate.py
@@ -105,7 +105,8 @@ Fill in the "answer" field for each item. Do not add or remove items.
 def _render_privacy_reanswer_prompt(row: dict[str, Any]) -> str:
     qa = parse_privacy_qa(row.get(COL_PRIVACY_QA))
     skeleton = [
-        {"id": item.id, "question": item.question, "answer": "", "confidence": 1.0, "reason": ""} for item in qa.items
+        {"id": item.id, "question": item.question, "answer": "", "confidence": 1.0, "reason": "", "evidence": []}
+        for item in qa.items
     ]
 
     prompt = """You are a privacy auditor. Read the text and answer each question with "yes" or "no".
@@ -116,7 +117,8 @@ def _render_privacy_reanswer_prompt(row: dict[str, Any]) -> str:
 - You must commit to "yes" or "no". Do not hedge.
 - Set confidence from 0.0 to 1.0 reflecting how strongly the answer is supported
 - Provide a brief reason grounded in the text
-- You MUST provide answer, confidence, and reason for EVERY item in the template below
+- Provide a list of short verbatim quotes from the text that support your answer (empty list if none)
+- You MUST provide answer, confidence, reason and evidence for EVERY item in the template below
 </rules>
 
 <text>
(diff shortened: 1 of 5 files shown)


src/anonymizer/engine/rewrite/repair.py
```diff
@@ -92,48 +68,26 @@ def _leaked_items_text(
...
-
     prompt = """You are helping to rewrite text for privacy protection.
 
 <privacy_goal>
 <<PRIVACY_GOAL>>
 </privacy_goal>
 
-<protection_decisions>
-<<PROTECTION_BLOCK>>
-</protection_decisions>
```
lipikaramaswamy, line 135: Does removing the structured protection context here intentionally change repair behavior for rows whose planned method is `replace`? Without `protection_decisions` and `replacement_map`, it looks like repair no longer has access to the selected method or the synthetic replacement values, so any repaired row may effectively fall back to generic inference suppression instead of honoring the earlier rewrite plan.

src/anonymizer/engine/rewrite/evaluate.py
```diff
@@ -105,28 +105,56 @@ def _render_quality_reanswer_prompt(row: dict[str, Any]) -> str:
...
+- For each question, focus on the specified entity label and target entity value.
+
+- Before answering each question, search the ENTIRE text for ALL evidence relevant to the entity label and target value.
+- Evidence may appear anywhere in the text, not just near the most obvious sentence.
+- Look for direct mentions, synonyms, paraphrases, and contextual cues.
+- Do not stop after finding one weak clue if stronger evidence exists elsewhere.
+- If stronger evidence exists, base your answer on the strongest evidence.
+- Include all short verbatim quotes that materially support your answer.
+
+- Ask: "If I had to guess the value for this entity label using only the text, what value would I guess?"
```
lipikaramaswamy, line 124: Have you seen any increase in false-positive leak calls with the stronger "most likely guess" standard here, especially for the geography/country-style examples? I can see the motivation for catching subtler inference, but since this output drives leakage scoring and repair, I’m curious whether the broader rubric has caused safe rewrites to be flagged more often in practice

lipikaramaswamy (commented):
Btw the failing test jobs seem expected from the current diff, since this PR removes repair prompt sections/helpers that the existing repair tests still assert on. If we do truly want to remove those sections, then the tests will need an update too. But let's discuss

src/anonymizer/engine/rewrite/repair.py
asteier2026: Yes, what I found is that simple things like replace and generalize are done on the first rewrite. By the time we get to repair, what we have left are complex context issues where we really just need to suppress inference. If we keep the original sensitivity disposition, that plan is no longer relevant and often just gets in the way.

src/anonymizer/engine/rewrite/evaluate.py
asteier2026: I kept a keen eye out for this, and I just don't see it happening. We get roughly the same number of repairs. I needed to push the prompt farther in making deductions. I was seeing NA pass QA, but the same entity fail  (really succeed) in the RAT-Bench "guess the value" type prompt. I also did regression testing on TAB and BIO and those datasets still look good.

lipikaramaswamy (approved):
LGTM 🚢 

Focusing repair on residual inference issues instead of carrying forward all first-pass rewrite guidance makes sense, and the tighter privacy-evaluation approach seems well justified 👍

PR #139, opened 2026-04-27
feat: re-export ModelProvider from anonymizer public API

### LINKED ISSUE #115: feat: re-export ModelProvider from anonymizer public API
## Summary

Users currently have to import `ModelProvider` directly from the Data Designer internals:

```python
from data_designer.config.models import ModelProvider
```

This leaks a dependency implementation detail into user code. `ModelProvider` should be re-exported from the `anonymizer` package so the full setup is self-contained:

```python
from anonymizer import Anonymizer, AnonymizerConfig, AnonymizerInput, Redact, Substitute, ModelProvider, configure_logging
```

## Current state

Configuring model providers programmatically (rather than via YAML) requires a `data_designer` import. This means users need to know about and directly depend on an internal Data Designer type, even though they are only configuring the Anonymizer.

## Proposed change

Add `ModelProvider` (and any other Data Designer types that surface in the public Anonymizer API) to `src/anonymizer/__init__.py` as explicit re-exports:

```python
from data_designer.config.models import ModelProvider as ModelProvider
```

This keeps the internal dependency on Data Designer but hides it behind the Anonymizer public surface. Users never need to know where `ModelProvider` comes from.

## Acceptance criteria

- [ ] `from anonymizer import ModelProvider` works
- [ ] All example notebooks updated to remove `data_designer` imports
- [ ] Any other Data Designer types currently imported directly by users are audited and re-exported if needed
- [ ] Public API surface documented in `__init__.py` with explicit `as` re-exports (PEP 484 `__all__` or explicit aliases) so the re-export is visible to type checkers

memadi-nv:
## Summary
- Re-export `ModelProvider` from the top-level `anonymizer` package so users no longer need to import `data_designer` internals when configuring providers programmatically.
- Users can now keep all public-API imports on a single line:
  ```python
  from anonymizer import (
      Anonymizer,
      AnonymizerConfig,
      AnonymizerInput,
      ModelProvider,
      Redact,
      Substitute,
      configure_logging,
  )
  ```
- Also fixes a pre-existing docs bug where the YAML provider example used fields that don't exist on the `ModelProvider` pydantic model.
## Changes
- **`src/anonymizer/__init__.py`**
  - Added `from data_designer.config.models import ModelProvider as ModelProvider` (explicit `as` alias so strict type checkers treat it as an intentional public re-export).
  - Added `"ModelProvider"` to `__all__`.
- **`docs/concepts/models.md`**
  - Rewrote the *Custom providers* section with two subsections — *YAML file* and *Programmatic* — sharing a single env-var setup block.
  - Fixed the YAML example: replaced invalid `base_url` / `api_key_env_var` with the real fields `name` / `endpoint`, matching the `ModelProvider` schema and the tested code path (`_resolve_model_providers`).
  - Added a programmatic example using `from anonymizer import Anonymizer, ModelProvider`.
  - Updated the *Custom models* example so the `model_providers` kwarg comment references the `providers` variable from the section above.
- **`tests/interface/cli/test_cli_model_configs.py`**
  - Updated `test_model_providers_list_passthrough` to import `ModelProvider` from `anonymizer` instead of `data_designer.config.models`, covering the new public re-export.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing

## Documentation

### DIFF
--- a/docs/concepts/models.md
+++ b/docs/concepts/models.md
@@ -31,60 +31,58 @@ Each pipeline stage has a **role** mapped to one of these aliases. See the full 
 
 ## Custom providers
 
-To use models from a different provider (OpenAI, OpenRouter, etc.), configure one or more providers when constructing the `Anonymizer`. Each provider maps a `name` to an API endpoint; the name is later referenced from your model configs.
+Use `model_providers` to define named API endpoints for hosted models such as OpenAI or OpenRouter.
 
-Providers can be defined either in a YAML file or programmatically in Python — both forms accept the same fields and are equivalent.
-
-Set your provider API keys as environment variables first:
+Set your API keys first:
 
 ```bash
 export OPENAI_API_KEY="your-openai-api-key"
 export OPENROUTER_API_KEY="your-openrouter-api-key"
 ```
 
-### YAML file
+=== "YAML"
 
-Define the providers in a YAML file and pass the path. API keys are read from the environment by the underlying client, so they don't need to appear in the file.
+    Define providers in a YAML file and pass the path to `Anonymizer`.
 
-```yaml
-# my_providers.yaml
-providers:
-  - name: openai
-    endpoint: https://api.openai.com/v1
-  - name: openrouter
-    endpoint: https://openrouter.ai/api/v1
-```
+    ```yaml
+    providers:
+      - name: openai
+        endpoint: https://api.openai.com/v1
+      - name: openrouter
+        endpoint: https://openrouter.ai/api/v1
+    ```
 
-```python
-from anonymizer import Anonymizer
+    ```python
+    from anonymizer import Anonymizer
 
-anonymizer = Anonymizer(model_providers="my_providers.yaml")
-```
+    anonymizer = Anonymizer(model_providers="my_providers.yaml")
+    ```
 
-### Programmatic
+=== "Python"
 
-Build a list of `ModelProvider` instances in code. Useful when you want your setup to be self-contained, or need to pull secrets from a vault rather than the environment.
+    Construct `ModelProvider` objects directly in code.
 
-```python
-import os
+    ```python
+    import os
+    from anonymizer import Anonymizer, ModelProvider
 
-from anonymizer import Anonymizer, ModelProvider
+    providers = [
+        ModelProvider(
+            name="openai",
+            endpoint="https://api.openai.com/v1",
+            api_key=os.environ["OPENAI_API_KEY"],
+        ),
+        ModelProvider(
+            name="openrouter",
+            endpoint="https://openrouter.ai/api/v1",
+            api_key=os.environ["OPENROUTER_API_KEY"],
+        ),
+    ]
 
-providers = [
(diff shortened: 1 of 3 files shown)


docs/concepts/models.md
```diff
@@ -31,28 +31,59 @@ Each pipeline stage has a **role** mapped to one of these aliases. See the full
 
 ## Custom providers
```
lipikaramaswamy, line 32: Suggest we do something like below - it'll make tabs like we have on the main docs page

---

  ## Custom providers

  Use `model_providers` to define named API endpoints for hosted models such as OpenAI or OpenRouter.

  Set your API keys first:

  ```bash
  export OPENAI_API_KEY="your-openai-api-key"
  export OPENROUTER_API_KEY="your-openrouter-api-key"
  ```
  === "YAML"

      Define providers in a YAML file and pass the path to `Anonymizer`.

      ```yaml
      providers:
        - name: openai
          endpoint: https://api.openai.com/v1
        - name: openrouter
          endpoint: https://openrouter.ai/api/v1
      ```

      ```python
      from anonymizer import Anonymizer

      anonymizer = Anonymizer(model_providers="my_providers.yaml")
      ```

  === "Python"

      Construct `ModelProvider` objects directly in code.

      ```python
      import os
      from anonymizer import Anonymizer, ModelProvider

      providers = [
          ModelProvider(
              name="openai",
              endpoint="https://api.openai.com/v1",
              api_key=os.environ["OPENAI_API_KEY"],
          ),
          ModelProvider(
              name="openrouter",
              endpoint="https://openrouter.ai/api/v1",
              api_key=os.environ["OPENROUTER_API_KEY"],
          ),
      ]

      anonymizer = Anonymizer(model_providers=providers)
      ```

  After defining providers, reference them from your model configs as described below.

src/anonymizer/__init__.py
```diff
@@ -7,6 +7,9 @@
 
 __version__ = version("nemo-anonymizer")
 
+# Re-exports from Data Designer so users don't import it directly.
```
lipikaramaswamy:
```suggestion
```

lipikaramaswamy: Suggest removing comment for brevity

lipikaramaswamy (approved):
Awesome, makes the interface cleaner for users! 

Suggest checking that docs changes - esp tabs - are rendering correctly. You can run `  make docs-serve` and open the local address to check.

src/anonymizer/__init__.py
memadi-nv: Thanks for catching this.

memadi-nv:
> Awesome, makes the interface cleaner for users!
> 
> Suggest checking that docs changes - esp tabs - are rendering correctly. You can run ` make docs-serve` and open the local address to check.

Tabs working as expected after the doc's change. I added the `nvidia` provider and its `API_KEY` in the docs, since it was called in the config but not defined in providers.

PR #141, opened 2026-04-28
refactor: replace df.attrs with typed ResolvedInput context

### LINKED ISSUE #4: refactor: replace df.attrs with typed pipeline context dataclass

memadi-nv:
## Summary

Replaces the experimental `df.attrs` mechanism for threading the user-facing
text-column name through the pipeline with a typed `ResolvedInput` dataclass
owned by the orchestrator. Workflows no longer need to preserve `attrs`
across pandas operations that silently drop them.

## Changes

- **New `ResolvedInput` dataclass** (`engine/resolved_input.py`): frozen
  dataclass holding the input `dataframe`, the user's `requested_text_column`,
  and the `resolved_text_column` (post-collision rename). `read_input` returns
  this directly.
- - **Reader (`engine/io/reader.py`)**: populates both requested and resolved
  text-column fields, and now resolves output-column collisions to a fixed
  point. If the selected text column is renamed, derived output names from the
  resolved text column are also reserved, preventing duplicate labels such as
  `final_entities__input_replaced`.
- **Orchestrator (`interface/anonymizer.py`)**: `_run_internal` accepts
  `ResolvedInput` and pulls `resolved_text_column` from it for output-column
  restoration — no more `df.attrs` reads.
- **Workflows** (`detection_workflow.py`, `llm_replace_workflow.py`,
  `rewrite_workflow.py`, `row_partitioning.py`): drop their `attrs=` plumbing
  / `merge_and_reorder` no longer takes the `attrs` parameter.
- **Public result types** (`interface/results.py`): `AnonymizerResult` and
  `PreviewResult` gain `resolved_text_column` (replaces `original_text_column`).
  `render_record_html` parameter renamed accordingly.

## Breaking Changes

- `AnonymizerResult.original_text_column` → `AnonymizerResult.resolved_text_column`
- `PreviewResult.original_text_column` → `PreviewResult.resolved_text_column`
- `render_record_html(..., original_text_column=...)` → `render_record_html(..., resolved_text_column=...)`

## Tests

- New `tests/engine/test_resolved_input.py`: covers construction, immutability,
  the requested-vs-resolved asymmetry, and the headline invariant that
  metadata survives pandas merge/concat/groupby.
- Workflow-level `assert "..." not in df.attrs` regression guards removed
  (they enforced an implementation detail, not the orchestrator's contract).
- Replaced with a behavioral orchestrator test that injects deliberately
  misleading `attrs` into mocked workflow output and asserts `Anonymizer.run()`
  + `display_record` still produce correct user-facing columns.

## Follow-up

- Add `renamed_input_columns` field to `ResolvedInput` so users can see all
  collision-driven renames, not just the text column (deferred per review).


lipikaramaswamy:
IMO we should tighten the abstraction before merging. The name `PipelineContext` indicates something broader than what this object currently is. It is not really passed through the whole pipeline; it is the result of input normalization plus metadata the top-level orchestrator needs when turning internal columns back into user facing columns. A name like `ResolvedInput`, `ResolvedInputFrame`, `InputContext`, or similar would describe the current role better.

Related: now that I think about it, `original_text_column` is misleading. In collision cases this can become something like `final_entities__input`, which is not the original requested column name anymore. It is the resolved user facing/output text column. I think the object can capture the full input-resolution result, for example: `dataframe`, `requested_text_column`, `resolved_text_column`, and `renamed_input_columns`. Ok to punt some of that as follow-up, but if we're introducing the dataclass now, it's probably worth getting the core vocabulary right.

Re tests: `assert "original_text_column" not in result.dataframe.attrs` enforces an implementation detail that workflows must not preserve attrs. But the real invariant is that the orchestrator no longer depends on attrs. We should instead test that mocked downstream workflow outputs have no attrs and `Anonymizer.run()` still restores the correct user-facing columns/display behavior.

src/anonymizer/engine/io/reader.py
```diff
@@ -97,19 +84,22 @@ def _resolve_output_column_collisions(dataframe: pd.DataFrame, *, selected_text_
...
-    text column whose name matches a fixed output column (e.g.
-    ``text_column='final_entities'``) it is renamed the same way, and
-    ``ResolvedInputColumns.selected_text_column`` reflects its new name so
-    downstream rename steps use the non-colliding identifier.
+    If the selected text column itself collides with a fixed output name (e.g.
+    ``text_column='final_entities'``) it is renamed the same way; the returned
+    ``ResolvedInput.resolved_text_column`` reflects the non-colliding
+    identifier while ``requested_text_column`` retains what the user asked for.
     """
     candidate_output_columns: list[str] = [f"{selected_text_column}{suffix}" for suffix in _OUTPUT_COLUMN_SUFFIXES]
```
lipikaramaswamy: Blocking edge case: collision detection is computed from `selected_text_column` before the selected text column may be renamed.

If `text_column="final_entities"` gets resolved to `final_entities__input`, an input column like `final_entities__input_replaced` is not treated as reserved. Later `_rename_output_columns(..., resolved_text_column="final_entities__input")` renames the internal replacement output to `final_entities__input_replaced`, creating duplicate user-facing column labels.

So I think the reader needs to also reserve output names derived from the final `resolved_text_column`, or keep resolving collisions until the resolved text column and all output names derived from it are collision-free.

lipikaramaswamy:
Note the conflict in `src/anonymizer/engine/replace/llm_replace_workflow.py`. When resolving, we should preserve `main`’s recent `_INTERNAL_COLUMNS` cleanup and only remove the old `attrs=` plumbing, otherwise this PR may reintroduce workflow scratch columns into returned trace dataframes.

src/anonymizer/engine/io/reader.py
memadi-nv: Addressed! Thanks for catching this.

src/anonymizer/engine/io/reader.py
```diff
@@ -85,48 +72,80 @@ def _validate_internal_column_collision(dataframe: pd.DataFrame, *, selected_tex
...
+
+    # Iterate until no new collisions surface. Termination is guaranteed:
+    #   - Pass 1 resolves collisions against `selected_text_column`'s derived
+    #     names plus the static names. If the text column itself was renamed,
+    #     a second pass re-derives candidates from the new name and catches
+    #     any secondary collisions.
+    #   - No further pass is possible because every rename target produced by
+    #     `_next_available_name` contains the `__input` infix, and no candidate
+    #     output name does. A third-pass collision would require those sets to
+    #     overlap, which they cannot.
```
lipikaramaswamy: Tiny comment nit: I think this termination explanation is a little too specific now that post-rename derived output names can contain `__input` themselves. Maybe simplify it to describe the fixed-point intent instead:

  ```suggestion
      # Iterate until no new collisions surface. If the selected text column is
      # renamed, the derived output names change too, so the loop re-checks those
      # post-rename names before returning.
```

lipikaramaswamy (approved):
Looks good. I rechecked the resolved-text-column collision case locally: `final_entities__input_replaced` is now reserved after `final_entities` resolves to `final_entities__input`, so the previous duplicate-column edge case is fixed. Small nit, but good to go.

PR #150, opened 2026-05-11
feat: improve sensitivity disposition calibration

asteier2026:
Changes include:
- Refactor entity_categories to assign explicit sensitivity tiers (high/medium/low) for               
  quasi-identifiers and latent identifiers, grounded in re-identification risk rather than content      
  sensitivity                                                                                           
  - Add combined_risk_assessment block clarifying the distinction between sensitivity and               
  combined_risk_level, and establishing that needs_protection follows from combined_risk_level          
  - Add bundle over-protection guidance to prevent medium assignment to every element of an identifying 
  bundle                                                                                                
  - Strengthen output_requirements to enforce combined_risk_level as the driver of needs_protection and 
  require protection_reason to demonstrate persistence after high-risk anchors are suppressed


asteier2026:
I made the changes greptile-apps recommended Wed morning.

lipikaramaswamy:
Q: in strict entity protection mode, should `combined_risk_level="low"` still be allowed?

I noticed `StrictEntityDispositionSchema` accepts `combined_risk_level="low"`, but `to_entity_disposition()` always converts strict entities with `needs_protection=True`. With the new validator, that creates `low + needs_protection=True`, which raises because low combined risk now requires `needs_protection=False`.

I may be missing the intended strict-mode semantics here, but should strict mode either constrain `combined_risk_level` to `high`/`medium`, or handle this conversion differently?

asteier2026:
Good catch Lipika, I made the adjustments.

src/anonymizer/engine/rewrite/sensitivity_disposition.py
```diff
@@ -39,6 +39,8 @@ def _get_sensitivity_disposition_prompt(
 
 All entities MUST be protected — you only decide HOW. Choose the most appropriate
 protection_method_suggestion for each entity. leave_as_is is not available.
+combined_risk_level must be 'high' or 'medium'. 'low' is not valid in strict mode
+because it implies no protection is needed.
 
 Ignore the MINIMUM NECESSARY CHANGE principle — it does not apply in strict mode.
 Ignore the QUASI-IDENTIFIERS guidance that says "NOT automatically sensitive" —
```
lipikaramaswamy: Tiny wording consistency nit to match L152

  ```suggestion
  Ignore the QUASI-IDENTIFIERS guidance that says "Not automatically protected" —
  ```

lipikaramaswamy (approved):
Looks good. I rechecked the strict-mode edge case: `combined_risk_level="low"` is now rejected by the strict schema, and `medium`/`high` convert cleanly with `needs_protection=True`. Tests ran great, and I'm assuming the end-to-end pipeline works with the updated prompt/schema.

Issue #152, opened 2026-05-11
bug: result.trace_dataframe is not persistable via to_parquet


lipikaramaswamy:
## Symptom

```python
result = anonymizer.preview(...)
result.trace_dataframe.to_parquet("trace.parquet")   # succeeds
pd.read_parquet("trace.parquet")                      # fails
```

```
ValueError: format number 1 of "list<element: struct<labels: list<element: string>, labels_str: string, value: string>>[pyarrow]" is not recognized
```

The error refers to a pandas extension dtype with no registered loader, and it makes the entire file unreadable — every column fails individually with the same error, because the parquet's pandas-metadata is parsed up front.

## Root cause

Only 3 of the 51 columns in `trace_dataframe` carry pyarrow-backed pandas extension dtypes:

| Column | dtype recorded in parquet pandas-metadata |
|---|---|
| `_entity_examples` | `string[pyarrow]` |
| `_entities_for_replace` | `list<element: struct<labels: list<element: string>, labels_str: string, value: string>>[pyarrow]` |
| `_entities_for_replace_json` | `string[pyarrow]` |

All three are constructed in `src/anonymizer/engine/replace/llm_replace_workflow.py:53-55`:

```python
working_df["_entity_examples"] = parsed_entities.apply(_create_entity_examples)
working_df["_entities_for_replace"] = parsed_entities.apply(_enrich_entities_for_template)
working_df["_entities_for_replace_json"] = working_df["_entities_for_replace"].apply(json.dumps)
```

These inherit a pyarrow-backed dtype from `parsed_entities` (which comes from arrow-typed entity columns upstream).

## Why it matters

- Users debugging via the trace (e.g. *"why did the validator keep this entity?"*, *"why didn't repair converge on this row?"*) can't persist or share their analysis.
- The forthcoming agent-skill workflow saves the trace by default for exactly this debugging use case.

`result.dataframe` is unaffected — it was already normalized to plain Python types in #21 (the `final_entities` `{"entities": [...]}` wrapper). The internal trace columns never got the same treatment.

## Suggested fix

Two options:

1. **Strip before returning** (minimal): drop these 3 workflow-internal columns from `trace_dataframe` in `interface/anonymizer.py` before constructing `PreviewResult` / `AnonymizerResult`. They're never user-facing and were never intended for export.
[...]

Issue #148, opened 2026-05-09
chore: use COL_* constants for intermediate columns in llm_replace_workflow.py


lipikaramaswamy:
The `STYLEGUIDE.md` rule *"All column names are constants in `engine/constants.py`. Never use string literals for column names"* is violated in `src/anonymizer/engine/replace/llm_replace_workflow.py:53-55`:

```python
working_df["_entity_examples"] = ...
working_df["_entities_for_replace"] = ...
working_df["_entities_for_replace_json"] = ...
```

These are workflow-internal intermediate columns.

### Fix

1. Add `COL_ENTITY_EXAMPLES`, `COL_ENTITIES_FOR_REPLACE`, `COL_ENTITIES_FOR_REPLACE_JSON` to `src/anonymizer/engine/constants.py`.
2. Replace the three string literals with those constants.

Mechanical fix; no behavior change.

Surfaced while reviewing the new `STYLEGUIDE.md` in #114.

PR #149, opened 2026-05-09
docs: add AGENTS.md, STYLEGUIDE.md, agent-assisted contribution (#114)

lipikaramaswamy:
## Summary

  - Add `AGENTS.md` — architecture overview, pipeline diagrams, and structural invariants for agents working in
  the codebase.
  - Add `STYLEGUIDE.md` — code conventions ruff and ty cannot enforce (Pydantic vs dataclass, error handling,
  column-name constants, prompt construction).
  - Add `CLAUDE.md` — 3-line redirect to `AGENTS.md` so Claude Code picks it up.
  - Add an "Agent-Assisted Development" subsection to `CONTRIBUTING.md` establishing the
  `plans/<issue-number>/<short-name>.md` convention for non-trivial changes.
  - Ignore `.claude/worktrees/` (Claude Code session worktrees).

  Filed #148 as a follow-up for three pre-existing column-name string-literal violations in
  `llm_replace_workflow.py:53-55` that the new STYLEGUIDE rule covers.

  Closes #114.

  ## Test plan

  - [ ] CI passes (format-check, `mkdocs build --strict`)
  - [ ] Reviewer can navigate the AGENTS.md ↔ STYLEGUIDE.md ↔ CONTRIBUTING.md cross-links
  - [ ] `git check-ignore .claude/worktrees/foo` returns the path (worktrees properly ignored)

PR #153, opened 2026-05-11
docs: add anonymizer Claude Code skill and supporting concept docs

lipikaramaswamy:
## Summary

- **Claude Code skill at `skills/anonymizer/`** — elicits dataset context, recommends mode + strategy, drafts a runnable Python script, iterates with the user via the workflow in `interactive.md`.
- **`docs/concepts/choosing-a-strategy.md`** (215L) — decision guide for mode (Replace vs Rewrite), strategy choice, privacy goal phrasing, and detection knobs. Doubles as the primary agent reference.
- **`docs/troubleshooting.md`** (210L) — symptom-first guide for dropped rows, leakage, low utility, and pipeline failures.
- **`README.md`** — new "Using with Claude Code" section pointing at the `skills.sh` installer.
- **`mkdocs.yml`** — adds the two new docs to navigation.
- **`src/anonymizer/__init__.py`** — exports `PrivacyGoal` at the top level (previously only accessible via deep import despite being referenced in docs and the skill's output template).
- **`docs/concepts/detection.md`** — minor wording polish.

## Testing notes

Tested end-to-end with 20 PMC-Patients clinical case reports. The interactive workflow was exercised: install verify → provider check → data inspect → clarify → plan → build → preview → results inspection. Two real workflow gaps surfaced and fixed during testing:

- Provider configuration wasn't checked proactively → added step 1 environment verification + step 3 provider question (commit `8ccd277`)
- Trace data wasn't persisted by default → script now saves `preview.parquet` for inspection (commit `5c5d5a0`)

The skill loader was also verified: `npx skills add` (skills.sh) discovers the skill correctly via the SKILL.md frontmatter.

## Related issues filed during this work

- #148 — replace string-literal column names in `llm_replace_workflow.py:53-55` with `COL_*` constants (existing STYLEGUIDE rule violation)
- #152 — `result.trace_dataframe` not persistable via `to_parquet`. **Blocks the skill's `preview.parquet` save from being readable until fixed.** Same 3 lines as #148.

## Test plan

- [ ] CI passes (`format-check`, `copyright-check`, `mkdocs build --strict`)
- [ ] `from anonymizer import PrivacyGoal` succeeds
- [ ] Skill discoverable via `npx skills add` against this branch
- [ ] Skill workflow produces a sensible script for a sample dataset (verified manually with PMC-Patients data)


lipikaramaswamy:
Once #149 merges, this PR needs a follow-up edit to AGENTS.md:
1. The dev-vs-user redirect at the top should also mention the bundled skill at skills/anonymizer/
2. A note that public-API changes (especially imports referenced in skills/anonymizer/SKILL.md's output template) may require corresponding skill updates.

docs/concepts/choosing-a-strategy.md
```diff
@@ -0,0 +1,215 @@
...
+    data_summary="De-identified inpatient progress notes from a US oncology service",
+)
+```
+
+What to include:
+
+- The domain (clinical, legal, financial, customer support, etc.)
+- The genre (notes, transcripts, opinions, biographies)
+- Anything about the source the engine couldn't infer from a single record (e.g. "transcribed phone calls — expect disfluencies")
+- *(Augmenter-only)* When you leave `entity_labels=None`, this is currently also the only place to nudge the augmenter LLM **away** from inventing labels you don't care about (e.g. "do not tag generic anatomical terms, medication class names, or job titles as PII"). Treat it as a soft do-not-tag list.
```
asteier2026: I'm not following. I thought entity_labels=None means use the default list and the augmenter can add what ever other entities/labels it deems necessary.

docs/concepts/choosing-a-strategy.md
```diff
@@ -0,0 +1,215 @@
...
+|---|---|---|
+| `None` (default) | Detect all `DEFAULT_ENTITY_LABELS`; the augmenter LLM can also infer new labels not in the default set | General-purpose — almost always the right starting point |
+| Explicit list | **Strict mode** — only the labels you list are detected, augmenter cannot invent new ones | You have a domain-specific entity that the defaults don't cover, or you want to *narrow* detection to a known short list |
+
+Common ways to extend the default list:
+
+- Healthcare: `mrn`, `clinical_facility`, `diagnosis_code`, `medication_name`
+- Legal: `case_number`, `court_name`, `docket_number`, `judge_name`
+- Customer support: `ticket_id`, `internal_user_id`, `transaction_id`
+- Internal: `employee_id`, `cost_center`, `internal_project_codename`
```
asteier2026: mrn is in the default list, as is employee-id and possibly case_number and court_name

docs/concepts/choosing-a-strategy.md
lipikaramaswamy: What I was trying to say here is that when entity_labels=None, yes the augmenter goes and finds whatever, but the only way to tell it what not to find is by saying it in data_summary. This is what i did for nemotron logs.

docs/concepts/choosing-a-strategy.md
```diff
@@ -0,0 +1,215 @@
...
+
+Both modes start from the same [detection](detection.md) pipeline. The difference is what happens after entities are detected.
+
+| Question | Replace | Rewrite |
+|---|---|---|
+| Is the goal "scrub the entities and keep everything else"? | ✅ | — |
+| Is the goal "produce a privacy-safe version of this text that downstream models can train on"? | — | ✅ |
+| Are there inferable / latent identifiers that aren't explicitly stated (e.g. "ringing the bell" → cancer treatment)? | ❌ leaves them | ✅ removes them |
+| Cost per record | ~1 LLM call (Substitute) or 0 LLM calls (Redact/Annotate/Hash) | Many LLM calls (domain → disposition → QA → rewrite → evaluate → repair → judge) |
+| Output text length | ≈ same as input | Often shorter / restructured |
```
asteier2026, line 104: There's still an LLM call for redact/annotate/hash.

skills/anonymizer/workflows/interactive.md
```diff
@@ -0,0 +1,48 @@
...
+
+3. **Clarify** — Ask the user the questions you need to write a config. Keep the set short; batch related questions; provide concrete options where possible. Common things to make precise:
+   - **Goal**: scrub PII for retention/sharing? produce training data? demos? compliance?
+   - **Mode**: Replace (entities only) vs Rewrite (full text transformation that also removes inferable identifiers like "ringing the bell" → cancer)? Default per the rule in `SKILL.md`.
+   - **For Replace**: which strategy? (`Substitute` for realistic-looking, `Redact` for explicit `[REDACTED_…]` markers, `Hash` for stable cross-row identifiers, `Annotate` for inspection only).
+   - **For Rewrite**: what must be protected? what must be preserved? how strict (`risk_tolerance`)? Read [`docs/concepts/choosing-a-strategy.md`](../../../docs/concepts/choosing-a-strategy.md) sections 5–6 with the user's answers in mind.
+   - **Domain-specific entity labels** the defaults won't cover (e.g. `"medical record number"`, `"case number"`, `"internal project codename"`). If yes, read [`docs/concepts/choosing-a-strategy.md`](../../../docs/concepts/choosing-a-strategy.md) section 2.
```
asteier2026: Again, medical record number and case number are bad examples

skills/anonymizer/SKILL.md
```diff
@@ -0,0 +1,178 @@
...
+- **`Annotate` is for inspection, not production.** Its output keeps the original entity text and is not privacy-safe. Use it during iteration to confirm detection is working, then switch.
+- **Always set `AnonymizerInput.data_summary`**, even briefly. It is the single cheapest quality lever and it improves both detection and rewrite.
+- **Never claim privacy guarantees.** Anonymizer is best-effort. Outputs may need human review depending on `risk_tolerance`. Tell the user this when you finalize.
+
+# Usage Tips and Common Pitfalls
+
+- **`Detect.entity_labels=None` (the default) is permissive** — the augmenter LLM may invent labels not in `DEFAULT_ENTITY_LABELS`. Setting an explicit list switches to **strict mode** where *only* the listed labels are detected. To add domain labels, *extend* the default, don't replace it: `entity_labels=DEFAULT_ENTITY_LABELS + ["medical record number", ...]`.
+- **GLiNER is zero-shot** — entity labels are natural-language strings (`"medical record number"`, `"internal project codename"`), not codes or enum values. Any label you can describe in English is a label GLiNER can detect.
```
asteier2026: Again, medical record number is in the default list

docs/concepts/choosing-a-strategy.md
lipikaramaswamy: Good catch — fixed in de0ab26. Cross-checked all examples against the actual DEFAULT_ENTITY_LABELS list and dropped redundant ones (`mrn`, `court_name`, `employee_id`). Also switched all examples to snake_case to match the convention — the validator only strips/lowercases, so `medical record number` and `medical_record_number` would have been treated as different labels. 

Note on `case_number`: verified it's not in `DEFAULT_ENTITY_LABELS`, so I kept that one

docs/concepts/choosing-a-strategy.md
lipikaramaswamy: I had meant that to be number of llm calls over and above detection here. But rephrased to clarify.

docs/concepts/choosing-a-strategy.md
lipikaramaswamy: Ok, rewrote the bullet to lead with what data_summary is for

skills/anonymizer/workflows/interactive.md
lipikaramaswamy: fixed in https://github.com/NVIDIA-NeMo/Anonymizer/commit/de0ab26c7700e6f0f82435209e3aa3c630a64e32

docs/concepts/choosing-a-strategy.md
```diff
@@ -0,0 +1,215 @@
...
+6. (Rewrite) Risk tolerance?         → Rewrite(risk_tolerance="minimal" | "low" | "moderate" | "high")
+```
+
+Steps 1–2 govern the [detection](detection.md) stage that runs first in **both** modes — usually the highest-leverage way to improve overall quality on a new dataset. Steps 3–6 shape the mode-specific transformation that follows detection.
+
+---
+
+## 1. (Detection) `data_summary`
+
+`AnonymizerInput.data_summary` is an optional one-line description that flows into LLM prompts. It is the single cheapest quality lever you have, and it improves both detection and rewrite.
```
alexahaushalter: Can it not also improve replace? If not, should we make a brief comment as to why?

docs/concepts/choosing-a-strategy.md
```diff
@@ -0,0 +1,215 @@
...
+
+- The domain (clinical, legal, financial, customer support, etc.)
+- The genre (notes, transcripts, opinions, biographies)
+- Anything about the source the engine couldn't infer from a single record (e.g. "transcribed phone calls — expect disfluencies")
+- `data_summary` is the only way to provide a soft do-not-tag list for the augmenter when `entity_labels=None` — the augmenter is free to invent labels beyond `DEFAULT_ENTITY_LABELS`, so use it to tell the LLM what *not* to tag (e.g. "do not tag generic anatomical terms, medication class names, or job titles as PII").
+
+What to leave out:
+
+- Lists of entity types **you want detected** (those go in `Detect.entity_labels`)
+- Privacy/utility goals (those go in `Rewrite.privacy_goal`)
```
alexahaushalter, line 51: You mention earlier that data summary can help with rewrite but it isn't exactly clear how (especially since these are things we aren't supposed to be including). Should we add another sentence on that?

docs/concepts/choosing-a-strategy.md
```diff
@@ -0,0 +1,215 @@
...
+- Generic phrasing ("text data" adds no signal)
+
+---
+
+## 2. (Detection) Detection knobs
+
+For most datasets the [detection](detection.md) defaults work. The main reason to tune `entity_labels` is when your data has **domain-specific entities that can be described in plain English** — GLiNER is zero-shot, so any concept you can name (e.g. `"clinical_facility"`, `"internal_project_codename"`) becomes a thing it can find. Match the snake_case convention of `DEFAULT_ENTITY_LABELS`. If the entities you care about aren't in the default list, write them down and add them. Tune `gliner_threshold` only when you see a specific recall or precision problem in preview.
```
alexahaushalter: "tune entity_labels" sounds like it is something you are using a model to fine-tune. Should we just say "adjust" instead?

And maybe "entity" instead of "thing"?

docs/concepts/choosing-a-strategy.md
```diff
@@ -0,0 +1,215 @@
...
+
+- Healthcare: `clinical_facility`, `diagnosis_code`, `medication_name`, `lab_test_code`
+- Legal: `case_number`, `docket_number`, `statute_citation`, `judge_name`
+- Customer support: `ticket_id`, `internal_user_id`, `transaction_id`
+- Internal: `cost_center`, `internal_project_codename`, `experiment_id`
+
+```python
+from anonymizer import DEFAULT_ENTITY_LABELS, Detect
+
+detect = Detect(entity_labels=DEFAULT_ENTITY_LABELS + ["mrn", "clinical_facility", "medication_name"])
```
alexahaushalter: I think normally we write out "medical_record_number" - is there a reason to use the abbreviation here instead?

docs/concepts/choosing-a-strategy.md
```diff
@@ -0,0 +1,215 @@
...
+### `gliner_threshold`
+
+Default `0.3`. The validator catches false positives downstream, so erring low is safe.
+
+| Symptom | Move | Try |
+|---|---|---|
+| Entities are being missed | Lower | `0.2` or even `0.15` |
+| Validator is slow / expensive — it's being handed a huge candidate list | Raise | `0.4`–`0.5` |
+
+A low threshold doesn't hurt *accuracy* — the validator runs in batches of `validation_max_entities_per_call` (default `100`, tunable on `Detect`), so a long candidate list just becomes more validator calls, not a worse validator. The cost you're paying for `gliner_threshold=0.2` is latency and tokens, not precision. Raise the threshold when that cost matters; otherwise, leave it low.
```
alexahaushalter: This suggests to me that I should always raise the threshold because it seems to save on cost without hurting accuracy. Can you add in what the trade-off is?

docs/concepts/choosing-a-strategy.md
```diff
@@ -0,0 +1,215 @@
...
+| Question | Replace | Rewrite |
+|---|---|---|
+| Is the goal "scrub the entities and keep everything else"? | ✅ | — |
+| Is the goal "produce a privacy-safe version of this text that downstream models can train on"? | — | ✅ |
+| Are there inferable / latent identifiers that aren't explicitly stated (e.g. "ringing the bell" → cancer treatment)? | ❌ leaves them | ✅ removes them |
+| Additional LLM calls (beyond shared detection) | ~1 (Substitute) or 0 (Redact/Annotate/Hash) | Many (domain → disposition → QA → rewrite → evaluate → repair → judge) |
+| Output text length | ≈ same as input | Often shorter / restructured |
+| Best for | Structured records, log scrubbing, known-list redaction | Free-text data with implicit identifiers (clinical notes, biographies, depositions, support transcripts) |
+
+**Default:** if the user hasn't said otherwise, prefer Rewrite for higher privacy standards.
```
alexahaushalter: I am a hesitant to prefer Rewrite since in my conversations so far, it seems like the vast majority want Replace. I also worry Rewrite has been less tested on different types of text, for example, do how do we expect it to perform for structured log data?

docs/concepts/choosing-a-strategy.md
```diff
@@ -0,0 +1,215 @@
...
+## 4. (Replace) Which strategy
+
+The four strategies are summarised in [Replace](replace.md#strategy-comparison). The decision rule:
+
+| You want… | Use | Why |
+|---|---|---|
+| Realistic-looking text safe for sharing or training | **Substitute** | LLM-generated synthetic values preserve readability |
+| Clear visual marking that an entity was removed | **Redact** | `[REDACTED_FIRST_NAME]` is unambiguous |
+| To inspect what was detected without losing the original | **Annotate** | Original text is preserved next to the label — **not privacy-safe on its own** |
+| Deterministic re-identification across rows (same person → same token) | **Hash** | Same input always produces the same hash digest |
```
alexahaushalter: Should we say "across documents" instead of "across rows"? No strong opinion but it seems like Anonymizer is more about text whereas rows sounds tabular to me.

docs/concepts/choosing-a-strategy.md
```diff
@@ -0,0 +1,215 @@
...
+
```
alexahaushalter, line 123: Should we add to prefer Substitute when not specified?

docs/concepts/choosing-a-strategy.md
```diff
@@ -0,0 +1,215 @@
...
+- The genre (notes, transcripts, opinions, biographies)
+- Anything about the source the engine couldn't infer from a single record (e.g. "transcribed phone calls — expect disfluencies")
+- `data_summary` is the only way to provide a soft do-not-tag list for the augmenter when `entity_labels=None` — the augmenter is free to invent labels beyond `DEFAULT_ENTITY_LABELS`, so use it to tell the LLM what *not* to tag (e.g. "do not tag generic anatomical terms, medication class names, or job titles as PII").
+
+What to leave out:
+
+- Lists of entity types **you want detected** (those go in `Detect.entity_labels`)
+- Privacy/utility goals (those go in `Rewrite.privacy_goal`)
+- Generic phrasing ("text data" adds no signal)
+
```
alexahaushalter, line 54: Should we add to this list that you should leave out requests related to substitution, such as all names should be Portuguese, and that those go in Substitute(instructions)?

docs/concepts/choosing-a-strategy.md
```diff
@@ -0,0 +1,215 @@
...
+`Rewrite` ships with sensible defaults for `protect` and `preserve` (auto-populated when you pass `Rewrite()` with no arguments). Override them when you can be more specific than the generic defaults.
+
+### How to write `protect`
+
+`protect` answers: **"What should not appear in the output, even by inference?"**
+
+| Pattern | Example |
+|---|---|
+| Direct identifiers + quasi-identifiers | `"All patient names, MRNs, dates of birth, and any combinations of attributes that could re-identify an individual"` |
```
alexahaushalter: Similar comment about writing out MRN. No strong opinion but I don't think that is as known of an acronym.

docs/concepts/choosing-a-strategy.md
```diff
@@ -0,0 +1,215 @@
...
+By default, low-risk quasi-identifiers may be left unchanged when the engine judges them safe in context. Set `strict_entity_protection=True` to force every detected entity into an active protection method.
+
+Use it when:
+
+- The data is from a regulated domain (medical, legal, financial) and a blanket policy is required
+- You're producing data for external sharing where any unchanged identifier is a compliance risk
+- Audit requires "every entity was actively protected"
+
+Don't use it when utility matters more than blanket protection — it tends to increase modifications and can lower `utility_score`.
+
```
alexahaushalter, line 185: Should we add that they should prefer setting strict_entity_protection=false if not otherwise specified?

I also worry that it will assume if the dataset is medical, legal, or financial it will then think it should set this to true. Wonder if/how we can make the "AND a blanket policy is required" more strong?

docs/concepts/choosing-a-strategy.md
```diff
@@ -0,0 +1,215 @@
...
+## 6. (Rewrite) Risk tolerance
+
+`risk_tolerance` is a rewrite-only knob — it selects a coherent bundle of repair and review thresholds. The full table is in [Rewrite > Risk tolerance](rewrite.md#risk-tolerance); the choice rule is below.
+
+| User signal | Pick |
+|---|---|
+| "Medical / legal / financial / external release" | `minimal` |
+| Default for most privacy-sensitive data | `low` |
+| "I want utility prioritised, this is internal-only" | `moderate` |
```
alexahaushalter: nit: US spelling is usually prioritized

docs/concepts/choosing-a-strategy.md
```diff
@@ -0,0 +1,215 @@
...
+- `high` does **not** auto-repair single high-sensitivity leaks. Use only when you have downstream review.
+- `max_repair_iterations` (default 3) caps cost. Set to 0 to skip repair entirely while still computing leakage / utility metrics — useful for audits.
+
+---
+
+## Goal → starting config cheat sheet
+
+A starting point for common scenarios. Always run `preview` and iterate from here.
+
+| Goal | Mode | Strategy / config |
```
alexahaushalter, line 211: Could add one more for structured data generally that it should use Replace / Substitute.

docs/concepts/choosing-a-strategy.md
```diff
@@ -0,0 +1,215 @@
...
+|---|---|---|
+| "Scrub PII from logs for retention" | Replace | `Redact()` |
+| "De-identify clinical notes for research sharing" | Rewrite | `Rewrite(privacy_goal=PrivacyGoal(protect="all PHI and any context that could imply a specific patient or facility", preserve="clinical findings, treatments, and outcomes"), risk_tolerance="minimal", strict_entity_protection=True)` |
+| "Produce realistic-looking biographies for demos" | Replace | `Substitute(instructions="Names and locations should remain plausible for the original cultural context.")` |
+| "Anonymize legal opinions for an SFT dataset" | Rewrite | `Rewrite(privacy_goal=PrivacyGoal(protect="party names, case numbers, judge names, and locations below the state level", preserve="argument structure and procedural posture"), risk_tolerance="low")` |
+| "Allow re-joining records by identifier without keeping the identifier" | Replace | `Hash(algorithm="sha256", digest_length=16)` |
+| "I just want to see what the detector finds" | Replace | `Annotate()` (preview only — never ship Annotate output as anonymised data) |
```
alexahaushalter: nit: anonymized (with z)

docs/concepts/choosing-a-strategy.md
alexahaushalter, line 211: I worry slightly that only one example suggests Substitute when I expect that will be like 60-90% of actual usage of the library. Maybe I'm way off though!

docs/concepts/detection.md
```diff
@@ -69,7 +69,7 @@ Pools also act as **failover**. If a chunk's assigned alias can't complete the c
 
 #### What happens when a row can't be validated
 
-If validation can't get a complete answer for a row — every alias in the pool has failed on at least one of that row's chunks — the row is **dropped from the output** rather than passed through with some entities unvalidated. This is deliberate: the alternative would be writing the original text back out with those entities still un-scrubbed, which is exactly the outcome you're trying to avoid.
+If validation can't get a complete answer for a row — every alias in the pool has failed on at least one of that row's chunks — the row is **dropped from the output** rather than passed through with some entities unvalidated. This is deliberate: the alternative would be writing the original text back out with those entities still un-scrubbed, which is an undesired outcome.
```
alexahaushalter, line 72: For later, should we consider making this a parameter?

docs/troubleshooting.md
```diff
@@ -0,0 +1,210 @@
+<!-- SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved. -->
+<!-- SPDX-License-Identifier: Apache-2.0 -->
+
+# Troubleshooting
+
+Symptom-first guide to common problems and how to fix them. Each entry says how to diagnose, what knob to turn, and what to verify after.
+
+When something looks wrong, **first confirm the run completed cleanly** (no dropped rows). Once you know the pipeline ran, **run `preview` on the failing rows** — the trace columns it produces are how you'll diagnose nine cases out of ten.
```
alexahaushalter: Are dropped rows and failed rows different? My first assumption is they are the same, and then it is confusing to me to make sure there were no dropped rows but to then investigate any that failed.

docs/troubleshooting.md
```diff
@@ -0,0 +1,210 @@
...
+---
+
+## Detection problems
+
+### Detection missed an entity I expected
+
+Try in order:
+
+1. **Lower `gliner_threshold`** from `0.3` to `0.2` (or `0.15`). False positives get caught downstream by validation.
+2. **Add the entity's label to `entity_labels`** if it's not in `DEFAULT_ENTITY_LABELS`. Domain-specific labels (`mrn`, `case_number`, `internal_project_codename`) won't be detected reliably without being listed.
```
alexahaushalter: Wondering about mrn being written out here too

alexahaushalter: Is it easy to add a single label to entity_labels and still include the entire default list?

I recall (maybe incorrectly) that if a user provides a list themselves, we are strict and only use that list. So do they have to copy the entire default list in now?

docs/troubleshooting.md
```diff
@@ -0,0 +1,210 @@
...
+
+### Replacement looks repetitive or unrealistic (Substitute)
+
+Symptoms: every "Alice" becomes "Maya," every city becomes "Springfield," every email becomes `john@example.com`.
+
+1. **Add domain hints to `Substitute.instructions`** — see [Choosing a strategy > Writing Substitute.instructions](concepts/choosing-a-strategy.md#writing-substituteinstructions).
+2. **Check the `replacement_generator` model** — small models default-collapse to high-frequency names. Try a stronger model for this role.
+3. **`Substitute` is consistent within a row, not across rows.** Repeated mentions of the same value in one row collapse to one replacement (so a person's name stays consistent in a single document). Across rows the LLM has no shared state, so "Alice" in different rows usually gets different replacements. If you need stable cross-row mappings, use `Hash` or post-process `result.trace_dataframe["_replacement_map"]`.
```
alexahaushalter: Would suggest reordering #3, e.g.

Use `Hash` or post-process `result.trace_dataframe["_replacement_map"]` if you need stable cross-row mappings. Recall: **`Substitute` is consistent within a row, not across rows.** Repeated mentions of the same value in one row collapse to one replacement (so a person's name stays consistent in a single document). Across rows the LLM has no shared state, so "Alice" in different rows usually gets different replacements.

This focuses on the action first and then explains that the other part about Substitute is just information.

docs/troubleshooting.md
```diff
@@ -0,0 +1,210 @@
...
+
+### Hash output isn't stable across runs
+
+Hashes are deterministic given the same `algorithm`, `digest_length`, and input text. Instability almost always means:
+
+- The detected text differs (different whitespace, casing, surrounding context). Look at the `<your_text_column>_with_spans` column to confirm.
+- The label differs between runs (label normalization is lossy — e.g. `first_name` vs `FIRST_NAME` collapse to the same value, but `name` vs `first_name` do not).
```
alexahaushalter: What can they do to fix this?

I am also confused because the next line explains that the label doesn't matter for hashing, whereas this line suggests that the label differing is causing a problem.

skills/anonymizer/workflows/interactive.md
```diff
@@ -0,0 +1,48 @@
...
+   - For deeper inspection: the script saves `preview.parquet` (trace dataframe — superset of the user-facing columns). Load with `pd.read_parquet("preview.parquet")` to inspect per-entity validation decisions, sensitivity dispositions, and other trace columns.
+
+7. **Iterate**
+   - Ask the user how the preview looks.
+   - If quality is off, consult [`docs/troubleshooting.md`](../../../docs/troubleshooting.md) at the relevant symptom (`leakage_mass too high`, `utility_score too low`, `Most rows have needs_human_review=True`, etc.). Apply the suggested knob change to the script, re-preview.
+   - Loop until the user is satisfied.
+
+8. **Finalize** — Tell the user they can run the full pipeline with:
+   - `python <script>.py --full`
+   - Caution that wall-clock and cost depend heavily on dataset size, model choice, and (for Rewrite) `max_repair_iterations`. Suggest they start by running on a larger sample (e.g. `--num-records 100`) before committing to the full dataset.
```
alexahaushalter, line 48: What is wall-clock? Is this a known term?

skills/anonymizer/SKILL.md
```diff
@@ -0,0 +1,179 @@
...
+
+# Rules
+
+- **Always preview before running the full pipeline.** Preview is cheap; a full run can be expensive and slow.
+- **If `result.failed_records` is non-empty after preview, fix that *before* tweaking strategy.** Dropped rows are a model/provider/infra problem (rate limits, auth, etc.), not a config problem. Strategy knobs won't help. See `docs/troubleshooting.md` "Did the run actually complete cleanly?".
+- **For cross-record consistency** (same value → same replacement everywhere), use `Hash`, not `Substitute`. `Substitute` is consistent within a row only.
+- **Default to `Rewrite`** for free-text content longer than ~2 sentences (clinical notes, depositions, biographies, support transcripts). **Default to `Replace`** for short or structured fields (log lines, single-cell PII, structured records).
+- **`Annotate` is for inspection, not production.** Its output keeps the original entity text and is not privacy-safe. Use it during iteration to confirm detection is working, then switch.
+- **Always set `AnonymizerInput.data_summary`**, even briefly. It is the single cheapest quality lever and it improves both detection and rewrite.
```
alexahaushalter, line 33: Same question as earlier - does this not help with Substitute?

docs/concepts/choosing-a-strategy.md
alexahaushalter: I think what you mention later in the Skill is a better happy medium:

- **Default to `Rewrite`** for free-text content longer than ~2 sentences (clinical notes, depositions, biographies, support transcripts). **Default to `Replace`** for short or structured fields (log lines, single-cell PII, structured records).

docs/troubleshooting.md
alexahaushalter: I see later you write:

- **`Detect.entity_labels=None` (the default) is permissive** — the augmenter LLM may invent labels not in `DEFAULT_ENTITY_LABELS`. Setting an explicit list switches to **strict mode** where *only* the listed labels are detected. To add domain labels, *extend* the default, don't replace it: `entity_labels=DEFAULT_ENTITY_LABELS + ["clinical_facility", ...]`. Match the snake_case convention of `DEFAULT_ENTITY_LABELS`.

Maybe we can make the phrasing you have here match more similarly to what I pasted above? I specifically like the word "extend"

skills/anonymizer/SKILL.md
```diff
@@ -0,0 +1,179 @@
...
+- [`docs/concepts/detection.md`](../../docs/concepts/detection.md) — detection internals (GLiNER threshold semantics, label catalogue, augmenter/validator behavior).
+- [`docs/concepts/models.md`](../../docs/concepts/models.md) — model role configuration, validator pools.
+
+# Troubleshooting
+
+Environment-level issues only. Quality and pipeline issues are in `docs/troubleshooting.md`.
+
+- **`anonymizer` not installed:** Tell the user `nemo-anonymizer` is not in this Python environment (requires Python ≥ 3.11). Ask if they want you to install it (`pip install nemo-anonymizer`) or do it themselves. Do not install without permission.
+- **Model aliases not configured:** Anonymizer can't run without `models.yaml` and `providers.yaml`. Tell the user to set these up — see `docs/concepts/models.md`. If they don't have a config yet, point them at `src/anonymizer/config/default_model_configs/` for the shipped defaults.
+- **LLM calls failing at preview:** Probably auth / network / wrong base URL. See `docs/troubleshooting.md` "Validation passed but `preview` errors at LLM call".
```
alexahaushalter: Should it say something more like "Probably auth error / network error / wrong base URL." ?

skills/anonymizer/SKILL.md
```diff
@@ -0,0 +1,179 @@
...
+    Anonymizer,
+    AnonymizerConfig,
+    AnonymizerInput,
+    DEFAULT_ENTITY_LABELS,
+    Detect,
+    # Pick what you need:
+    # Replace mode:
+    Substitute, Redact, Annotate, Hash,
+    # Rewrite mode:
+    Rewrite, PrivacyGoal,
```
alexahaushalter, line 91: Should there be a comma after PrivacyGoal? I would've guessed no since it is last in the list.

skills/anonymizer/SKILL.md
```diff
@@ -0,0 +1,179 @@
...
+    data = AnonymizerInput(
+        source="TODO: path to .csv / .parquet / .jsonl",
+        text_column="TODO: name of the text column",
+        data_summary="TODO: one-line description of the data (domain, genre, anything non-obvious)",
+    )
+
+    detect = Detect(
+        # Add domain labels by *extending* the default, not replacing it.
+        # entity_labels=DEFAULT_ENTITY_LABELS + ["clinical_facility", "diagnosis_code"],
+        gliner_threshold=0.3,  # default; lower (0.2) for recall, raise (0.5) for cost
```
alexahaushalter: Should we say "for cost savings"? to be more explicit (rather than just "for cost"?

README.md
```diff
@@ -126,6 +126,20 @@ AnonymizerConfig(replace=Hash(algorithm="sha256", digest_length=8))
 
 ---
 
+## Using with Claude Code
+
+This repo ships a Claude Code skill at [`skills/anonymizer/`](skills/anonymizer/SKILL.md) that elicits your dataset's privacy requirements, recommends Replace vs Rewrite plus a strategy, and drafts a runnable script for you to iterate on. While the skill should work with other coding agents that support skills, development and testing has focused on Claude Code at this stage.
```
alexahaushalter: Should we say "Rewrite vs Replace plus a strategy" instead? The strategy only applies when picking Replace, right?

docs/concepts/choosing-a-strategy.md
lipikaramaswamy: Reframed in 0c46cc0 — data_summary improves detection (precursor to both Replace and Rewrite), and in Rewrite mode additionally informs meaning preservation.

docs/concepts/choosing-a-strategy.md
lipikaramaswamy: Added

docs/concepts/choosing-a-strategy.md
lipikaramaswamy: Added

docs/concepts/choosing-a-strategy.md
lipikaramaswamy: Expanded, feel free to wordsmith

docs/concepts/choosing-a-strategy.md
lipikaramaswamy: added

docs/concepts/detection.md
lipikaramaswamy: Something like `keep_failed` and return blanks?

skills/anonymizer/workflows/interactive.md
lipikaramaswamy: yup standard, wall-clock time includes CPU/GPU time, waiting for CPU/GPU, I/O, and other processes

skills/anonymizer/SKILL.md
lipikaramaswamy: It doesn't, because we simply use instructions specified in Substitute(). It'll always help entity detection.

skills/anonymizer/SKILL.md
lipikaramaswamy: added

skills/anonymizer/SKILL.md
lipikaramaswamy: Doesn't matter

docs/concepts/choosing-a-strategy.md
lipikaramaswamy: Addressed in the same 0c46cc0 rewrite of the data_summary section — it now says "In Rewrite mode it additionally provides context on what the data contains, helping the rewriter preserve meaning."

docs/concepts/choosing-a-strategy.md
lipikaramaswamy: I think the user should decide rather than us codifying a heuristic. One extra question on mode is cheap compared to picking wrong by rule.

README.md
lipikaramaswamy: Fixed in 19e8d84 — reworded to "Rewrite, or Replace with a strategy" so the strategy clause is unambiguously tied to Replace.

skills/anonymizer/SKILL.md
lipikaramaswamy: Fixed in 19e8d84 — "for cost" → "for cost savings".

docs/concepts/choosing-a-strategy.md
lipikaramaswamy: Fixed in 0c46cc0 — "tune" → "adjust", "thing" → "entity".

docs/concepts/choosing-a-strategy.md
lipikaramaswamy: Fixed in 0c46cc0 — example now uses non-default labels (clinical_facility, diagnosis_code, medication_name). 63c36ae also fixed the underlying tuple + list TypeError.

docs/concepts/choosing-a-strategy.md
lipikaramaswamy: Reworked in 0c46cc0 — the paragraph now makes both directions' costs explicit: lowering pays latency/tokens, raising trades for recall risk. Plus a nudge to verify high-threshold setups with an Annotate preview.

docs/concepts/choosing-a-strategy.md
lipikaramaswamy: Fixed in 0c46cc0 — "across rows" → "across documents" in the Hash row.

docs/concepts/choosing-a-strategy.md
lipikaramaswamy: Fixed in 0c46cc0 — "MRNs" → "medical record numbers" in the PrivacyGoal example.

docs/concepts/choosing-a-strategy.md
lipikaramaswamy: Fixed in 0c46cc0 — "prioritised" → "prioritized".

docs/concepts/choosing-a-strategy.md
lipikaramaswamy: Fixed in 0c46cc0 — "anonymised" → "anonymized".

docs/concepts/choosing-a-strategy.md
lipikaramaswamy: Added two more Substitute rows in 0c46cc0 (survey responses, customer support transcripts for fine-tuning), so the cheat sheet now has three distinct Substitute examples clustered together.

docs/troubleshooting.md
lipikaramaswamy: Good catch — they were the same thing. Unified in 0c46cc0: the intro now uses one term ("rows that didn't make it through the pipeline, usually a rate-limit / infra issue") and explicitly contrasts with quality-issue rows (high leakage_mass, low utility_score, needs_human_review).

docs/troubleshooting.md
lipikaramaswamy: Fixed in 0c46cc0 — the bullet no longer uses "mrn"; the worked example now shows clinical_facility as the extension label.

docs/troubleshooting.md
lipikaramaswamy: Done in 19e8d84 — bullet now leads with the fix ("use Hash or post-process result.trace_dataframe['_replacement_map']") and the per-row vs cross-row background follows.

docs/troubleshooting.md
lipikaramaswamy: Good catch — the bullets contradicted each other. Rewrote the whole section in 19e8d84 to distinguish the digest (text-only) from the output wrapper (templated with label, default <HASH_{label}_{digest}>). Added the format_template="<HASH_{digest}>" workaround for users who want label-free output.

docs/troubleshooting.md
lipikaramaswamy: The troubleshooting bullet now leads with "Extend the default list" and matches the SKILL.md framing (strict-mode explanation + the [*DEFAULT_ENTITY_LABELS, ...] pattern). Fixed in 0c46cc0 / 63c36ae.

skills/anonymizer/SKILL.md
lipikaramaswamy: Fixed in 19e8d84 — "for cost" → "for cost savings".

README.md
lipikaramaswamy: Fixed in 19e8d84 — reworded to "Rewrite, or Replace with a strategy" so the strategy clause is unambiguously tied to Replace.

docs/concepts/choosing-a-strategy.md
alexahaushalter: Looks good! We can always adjust later if we see the skill does this too much.

docs/concepts/choosing-a-strategy.md
alexahaushalter: Awesome thanks!

docs/concepts/detection.md
alexahaushalter: Yeah that's what I'm thinking! Ping me if you want me to make a separate issue for this. Certainly not urgent.

alexahaushalter (approved):

docs/concepts/detection.md
lipikaramaswamy: Yup I'm resolving now but please do make another issue!

Issue #106, opened 2026-04-09
feat(rewrite): formal judge rubric selection and score distribution analysis


lipikaramaswamy:
## Background

The rewrite pipeline's final judge currently uses Option A: **privacy, quality, naturalness** on a 1–10 scale, adapted from an earlier research repo. PR #86 improved the judge prompt wording and iterated on scoring behavior as an interim fix, but the formal rubric decision was never made. This issue closes that out.

Closes #37.

## Options under consideration

| Option | Dimensions | Scale | Notes |
|--------|-----------|-------|-------|
| **A (current)** | privacy, quality, naturalness | 1–10 | Baseline; `quality` and `naturalness` overlap in practice |
| **B** | privacy, utility, naturalness | 1–10 with improved anchors | Renames quality → utility for clarity; adds explicit anchor descriptions per score band |
| **C** | privacy, utility, faithfulness, fluency | 1–5 | More granular separation of reservation (faithfulness) vs. readability (fluency); narrower scale may reduce score clustering |

## Scope of work

- [ ] Run score distribution analysis on TAB and NVIDIA Synthetic Biographies using current rubric (Option A) — check for score clustering, ceiling effects, and correlation between dimensions
- [ ] Evaluate Options B and C against the same datasets and compare differentiation
- [ ] Document chosen rubric with justification and sample score distributions
- [ ] Update anchor point descriptions for chosen option
- [ ] Update `FinalJudgeWorkflow` implementation
- [ ] Update test coverage

## Related
- Supersedes #37
- See #98 for LLM-as-a-judge coverage for REPLACE mode

PR #158, opened 2026-05-14
feat: Add LLM-as-a-judge for Anonymizer replace evaluation

### LINKED ISSUE #98: feat: LLM-as-a-judge for REPLACE evaluation
### Priority Level

Medium (Nice to have)

Hard to tell if all the PII was found and replaced well when using Replace.

The evaluation for rewrite is great and we would like to offer the option to extend it to replace in some capacity.

Ideal eval is a human reviewing, but in absence of that, can an LLM optionally review to help answer "Did it actually (1) find all the PII and (2) replace it in a contextually relevant way?"

This could also be very helpful when trying to run Anonymizer on a language one is unfamiliar with to have a sense of whether Anonymizer can perform well on that language. And ultimately be able to provide more benchmark info.

_No response_

### Additional context

_No response_

memadi-nv:
## Summary

Adds a new opt-in step — `Anonymizer.evaluate(output)` — that scores a `preview()` / `run()` output along four metrics, plus the supporting workflows, prompts, schemas, and display rendering. Evaluation is intentionally **separate from anonymization** so the user can iterate on judges without re-paying the detection + replacement cost.

## What the evaluation looks like

For Substitute, four LLM-as-judge metrics run per record, each surfaced as a "LLM alignment score" with a colored verdict badge:

- **Detection Validity** — were the detected entities real PII (no false positives, mislabels, or boundary errors)?
- **Type Fidelity** — does each synthetic preserve the entity class and format?
- **Relational Consistency** — do cross-entity relationships (city↔country, dob↔age, name↔email) stay coherent?
- **Attribute Fidelity** — do salient within-entity attributes (gender of name, age bucket) survive?

For non-Substitute strategies (Annotate / Redact / Hash), only Detection Validity runs — the other three require a replacement map. 

## Public API and Evaluation Preview:

```python
preview = anonymizer.preview(config=cfg, data=src, num_records=15)   # detection + replacement, no judges
evaluated = anonymizer.evaluate(preview)                              # opt-in evaluation
evaluated.display_record(0)
```

Screenshot 2026-05-22 at 18 17 51

The replace strategy travels on AnonymizerResult / PreviewResult (replace_method field), so callers don't restate the mode and can't mis-state it. Pickle round-trips of preview / result preserve everything — supports the "anonymize once, re-evaluate many times" workflow.

## Runtime efficiency

The evaluation step is the LLM-heavy part of this PR, so a few choices keep wall time as low as possible:

- **One DataDesigner workflow, four columns.** All four judges are added as columns of a single `adapter.run_workflow(...)` call; DD schedules the LLM calls concurrently within that workflow up to the model's `max_parallel_requests`. No Python threads, no shared-state races.
- **Per-judge passthrough at flatten time.** Each judge's `postprocess()` overrides trivially-passthrough rows (e.g., 0 entities, < 2 replacements) with default verdicts so degenerate rows don't muddy the score.
- **No judge work during anonymization.** `preview()` / `run()` no longer fire any judges — that 4× LLM cost is paid only when the user explicitly asks for evaluation.
- **Non-critical evaluation preserves all rows.** When the LLM drops a row during the judge workflow, it survives in the output with `*_valid=None` ("Unavailable") via a left-merge on record id; the drop is still surfaced through `failed_records`. Anonymization row counts never shrink because of evaluation-side failures.
- **Failure isolation preserved.** If one judge column's structured-output parse fails for a row, only that metric becomes Unavailable; the other three still report.
- **Display is robust to inconsistent LLM responses.** When the LLM marks a row `valid=False` but enumerates no specific failures, the badge renders Not Satisfied (red) rather than letting the count math produce a misleading green.

## Module layout

Judges live under a dedicated evaluation package so the layout reflects "evaluation is its own axis," not a sub-concern of replace:

src/anonymizer/engine/evaluation/
    detection_judge.py                 # mode-agnostic
    replace/
        type_fidelity_judge.py         # Substitute-only
        relational_consistency_judge.py
        attribute_fidelity_judge.py

Model selection follows the same split: judge aliases moved out of `replace.yaml` into a new `evaluate.yaml` with `detection_validity_judge` and three `replace_*_judge` roles. `preview()` / `run()` validate the roles needed to produce anonymized output; `evaluate(...)` validates the roles needed to score it.

## Updated tutorial notebooks
- **Notebooks 1–3** (`01_your_first_anonymization`, `02_inspecting_detected_entities`, `03_choosing_a_replacement_strategy`) gained an "Optional: Evaluate" section explaining the opt-in nature and showing `anonymizer.evaluate(...)`.

## Validation on real datasets

Validated end-to-end on two datasets with `num_records=15` using replace/substitute mode. This experiment uses `gpt-oss-120b` as the LLM judge from internal inference models (not build.nvidia.com) and sets `max_parallel_requests=16` to reduce anonymization and evaluation runtime. 

## Job run time (anonymization / evaluation)

| Dataset | Anonymization | Evaluation |
|---|---|---|
| `NVIDIA_synthetic_biographies` | 1m 08s | 58s |
| `TAB_legal_sample25` | 1m 42s | 1m 16s |

### Performance — % of records with LLM alignment score = 100% across the four metrics

| Metric | `NVIDIA_synthetic_biographies` | `TAB_legal_sample25` |
|---|---|---|
| `detection_valid` | 26.67% | 0.00% |
| `type_fidelity_valid` | 86.67% | 93.33% |
| `relational_consistency_valid` | 93.33% | 80.00% |
| `attribute_fidelity_valid` | 93.33% | 80.00% |

`detection_valid` is the lowest-scoring metric on both datasets — most records don't reach 100% detection accuracy. Besides, there are times that false positives in `detection validity` score might leak into the `Type Fidelity` score.  We're hoping that enabling a more capable LLM for entity detection during anonymization can significantly reduce this issue - #165.

## Test plan

- 4 new test files (one per judge) — passthrough behavior, malformed-payload tolerance, schema/alias plumbing.
- Extended `test_replace_runner.py` — single-DD-workflow dispatch, detection-only path for non-Substitute, row-preservation on LLM drop, error on missing trace columns.
- `_verdict_badge` unit tests covering all four color paths including the `valid=False, correct==total` precedence rule.
- Empty-section render regression test (no "Detection Judge" heading when no evaluation has run).
- **821 tests pass locally**; pipeline validated end-to-end via the demo notebook + tutorial notebooks 1–3.

## Files (34 changed, +4332 / −231)

- **New**: `src/anonymizer/engine/evaluation/{detection_judge.py, replace/*}` and their tests; `src/anonymizer/config/default_model_configs/evaluate.yaml`; demo notebook (`evaluate-replace-demo.ipynb`).
- **Modified**: `interface/anonymizer.py` (new `evaluate` method), `engine/replace/replace_runner.py` (judge dispatch via single DD workflow), `interface/display.py` (four metric sections + verdict badge + self-wrapping detection section), `engine/constants.py`, `config/default_model_configs/replace.yaml`, `config/models.py`, `interface/results.py` (carries `replace_method`), notebooks 01–03.

## Future improvements

- **`EvaluateConfig` is an empty placeholder today.** Reserved for evaluation-specific knobs we plan to add — selected-metric subset, per-judge model/prompt overrides, scoring thresholds. The class lives in the public surface so future fields can land without changing imports.
- **Rewrite-side evaluation will reuse the same `anonymizer.evaluate(...)` entry point** in upcoming PRs (likely as part of #106). The mode metadata on `AnonymizerResult` already includes a forward slot for `rewrite`, and `EvaluateModelSelection` has a natural place for `rewrite_*_judge` fields when those land.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing

## Documentation

## Related Issues
Closes #98


docs/notebooks/evaluate-replace-demo.ipynb
memadi-nv, line 1: This notebook is just for the purpose of demo and experimenting with the replace evaluation. Once all changes are approved, it'll be removed.

src/anonymizer/engine/evaluation/replace/attribute_fidelity_judge.py
```diff
@@ -0,0 +1,343 @@
...
+        )
+        run_result = self._adapter.run_workflow(
+            with_replacements,
+            model_configs=model_configs,
+            columns=[self.column_config(selected_models)],
+            workflow_name="replace-attribute-fidelity-judge",
+            preview_num_records=effective_preview_num_records,
+        )
+
+        judged_df = run_result.dataframe.copy()
```
lipikaramaswamy, line 337: Because evaluation is non-critical, I don’t think we should let dropped judge rows shrink the anonymization result.

If `adapter.run_workflow(...)` returns fewer rows than `prepared`, `run_result.failed_records` is populated, but `judged_df = run_result.dataframe` means the returned `AnonymizerResult` loses those records entirely. That changes the row count of a previously successful `preview()` / `run()` just because evaluation failed for a subset of rows.

Let's join judge columns back onto `prepared` by record id / uuid col and set the missing rows’ judge verdicts to unavailable (`*_valid=None`, invalid lists empty), while preserving all original rows. This is similar to how rewrite treats final judge as non-critical.

docs/notebooks/evaluate-replace-demo.ipynb
```diff
@@ -0,0 +1,364 @@
+{
+ "cells": [
```
lipikaramaswamy: reminder to delete this file before merging

src/anonymizer/interface/display.py
```diff
@@ -509,6 +1032,14 @@ def _normalize_disposition(raw: object) -> list[dict[str, str]]:
...
 border-radius:6px;opacity:0.85">{replaced_html}</div>
       </div>
+      <div style="margin-bottom:16px">
+        <div style="font-size:0.8em;font-weight:600;text-transform:uppercase;\
+letter-spacing:0.05em;margin-bottom:6px;opacity:0.5">Detection Judge</div>
+        {detection_judge_html}
+      </div>
+      {type_fidelity_section}
+      {attribute_fidelity_section}
+      {relational_consistency_section}
```
lipikaramaswamy: This wrapper renders even when no evaluation has been run, because `_render_detection_judge_section()` returns `""` but the template still includes the outer “Detection Judge” heading.

So ordinary `preview().display_record(...)` will show an empty judge section, even though the PR intentionally keeps judges out of `preview()` / `run()`.

Can we move the wrapper/heading into `_render_detection_judge_section()` itself, like the other judge sections, so the whole section is omitted when `COL_DETECTION_VALID` is absent?

src/anonymizer/engine/evaluation/detection_judge.py
```diff
@@ -0,0 +1,336 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
lipikaramaswamy, line 1: I don’t think `DetectionJudgeWorkflow` should live under `engine/replace/`, since detection validity evaluates the detection artifacts rather than replacement behavior. It also runs for non-Substitute strategies and could plausibly be / should reused by rewrite. 

I considered `engine/detection/`, but I think that package should stay focused on producing detection outputs. This judge is a post-hoc evaluation workflow that consumes detection outputs, so an evaluation-oriented package seems cleaner.

Could we move these under something like:

  ```text
  src/anonymizer/engine/evaluation/
    detection_judge.py
    replace/
      type_fidelity_judge.py
      relational_consistency_judge.py
      attribute_fidelity_judge.py
  ```

Names are suggestions but the module layout should reflect evaluation as its own workflow axis, separate from replacement execution.

src/anonymizer/config/default_model_configs/replace.yaml
```diff
@@ -3,3 +3,7 @@
 
 selected_models:
   replacement_generator: gpt-oss-120b
```
lipikaramaswamy, line 5: I think these judge aliases should live in a separate evaluation model-selection section rather than under `replace`.

`replace.yaml` should describe the models needed to perform replacement. These roles are only needed for the separate opt-in evaluation step, so putting them under `replace` means anonymization-time config now owns evaluation-only models.

Could we add a first-class `evaluate` section instead? Something like:

  ```yaml
  selected_models:
    replace:
      replacement_generator: gpt-oss-120b

    evaluate:
      detection_validity_judge: gpt-oss-120b
      replace_type_fidelity_judge: gpt-oss-120b
      replace_relational_consistency_judge: gpt-oss-120b
      replace_attribute_fidelity_judge: gpt-oss-120b
  ```

That would make validation cleaner too: `preview()` / `run()` validate the roles needed to produce anonymized output, while `evaluate(...)` validates the roles needed to score an existing output. It also gives us a natural place to add rewrite-evaluation roles later without making `replace.yaml` or `rewrite.yaml` own evaluation config.

lipikaramaswamy (commented):
Thanks, this is a useful direction. The separation between anonymization and evaluation makes sense, as does the single DataDesigner workflow for the judges.

A few things to address 
1. Model-alias validation now needs to distinguish anonymization-time roles from evaluation-time roles.
2. `evaluate_replace()` / soon to be `evaluate()` should preserve all input rows even if the optional judge workflow drops a subset of rows.
3. The display currently shows an empty “Detection Judge” section before evaluation has been run.
4. The demo notebook shouldn't ship. 

Left more specific comments inline.

src/anonymizer/config/default_model_configs/replace.yaml
memadi-nv: The schema you drew is what user-supplied YAML sees in this implementation — `selected_models: { replace: { replacement_generator: ... }, evaluate: { detection_validity_judge: ... } }` works today. For internal defaults, I followed the existing per-workflow file convention (one yaml per workflow), so the four judge aliases live in a new evaluate.yaml — just wanted to flag the file split is separate from the schema you described."

src/anonymizer/engine/evaluation/replace/attribute_fidelity_judge.py
```diff
@@ -0,0 +1,343 @@
...
+
+
+class AttributeCheck(BaseModel):
+    original: str = Field(description="Original value taken verbatim from the replacement map.")
+    label: str = Field(description="Entity label assigned to the original value.")
+    synthetic: str = Field(description="Synthetic value that replaced the original.")
+    attributes_checked: list[str] = Field(
+        default_factory=list,
+        description=(
+            "Salient within-entity attributes inspected for this triple (e.g. ['gender', 'cultural_origin'])."
```
lipikaramaswamy: Small schema/prompt mismatch: the field description gives `['gender', 'cultural_origin']` as an example, but the prompt example checks gender and age bucket.

Can we remove `cultural_origin` from the schema description so there's consistency? Or just remove examples altogether.

src/anonymizer/engine/evaluation/replace/attribute_fidelity_judge.py
```diff
@@ -0,0 +1,343 @@
...
+   - Check only when the original name CLEARLY implies a gender (e.g. "Valentina", \
+     "Michael"). If the original is gender-neutral or ambiguous (e.g. "Alex", "Taylor", \
+     "J.", a surname-only token whose gender you can't reliably tell) -> SKIP this triple.
+   - The check: does the synthetic name carry the SAME implied gender as the original?
+
+2. AGE BUCKET — applies to labels: age, date_of_birth.
+   - Buckets: child (0-12), teen (13-19), young adult (20-29), adult (30-44), \
+     middle-aged (45-64), senior (65+). It's fine if the years are +/- 1 year of the original.
+   - For `age`, the bucket comes from the numeric value directly. For `date_of_birth`, \
+     compute age = current_year - dob_year, then map to a bucket.
```
lipikaramaswamy: I'm not sure current year works for all endpoints, e.g. build and inference inject the current calendar date into every request, but a custom brev endpoint i stand up may not.

src/anonymizer/engine/evaluation/detection_judge.py
```diff
@@ -0,0 +1,336 @@
...
+Return ONLY the JSON object that matches the required schema. Do NOT wrap your output in \
+``` or ```json markdown fences. Do NOT include any commentary, reasoning, preamble, or text \
+outside the JSON object. Your entire response must be a single valid JSON object.
+</output_format>
+"""
+    return substitute_placeholders(
+        prompt,
+        {
+            "<<COL_TEXT>>": _jinja(COL_TEXT),
+            "<<ENTITIES_COLUMN>>": _ENTITIES_FOR_JUDGE_COL,
```
lipikaramaswamy, line 158: The task is defined at `(value, label)` granularity, and the output schema asks for one invalid `value` + `label`, but the rendered input groups labels by value:

`"Alice" -> first_name, doctor_name`

Can we render one row per pair instead?

  ```text
  - value="Alice" | label=first_name
  - value="Alice" | label=doctor_name
```
That makes the judge task, invalid entries, and display denominator line up more cleanly.

File: src/anonymizer/engine/evaluation/replace/attribute_fidelity_judge.py [hunk 9]
```diff
@@ -0,0 +1,343 @@
...
+    COL_ATTRIBUTE_FIDELITY_VALID,
+    COL_REPLACEMENT_MAP,
+)
+from anonymizer.engine.ndd.adapter import FailedRecord, NddAdapter
+from anonymizer.engine.ndd.model_loader import resolve_model_alias
+from anonymizer.engine.prompt_utils import substitute_placeholders
+from anonymizer.engine.row_partitioning import merge_and_reorder, split_rows
+from anonymizer.engine.schemas import EntityReplacementMapSchema
+
+logger = logging.getLogger("anonymizer.replace.attribute_fidelity_judge")
```
lipikaramaswamy: Small cleanup from the package move: these modules now live under `engine/evaluation`, but the logger names still use `anonymizer.replace.*`.

Can we update these to `anonymizer.evaluation.*` / `anonymizer.evaluation.replace.*` so logs match the new module ownership?

src/anonymizer/interface/anonymizer.py
```diff
@@ -218,6 +243,83 @@ def preview(
...
+        Args:
+            output: An :class:`AnonymizerResult` or :class:`PreviewResult` from
+                a prior ``preview()`` / ``run()``. Carries the trace dataframe,
+                the resolved text-column name, and the replace strategy.
+            config: Optional :class:`EvaluateConfig` for evaluation-specific
+                knobs (placeholder today; reserved for metric selection,
+                per-judge model/prompt overrides, etc.).
+        """
+        _ = config  # placeholder; no knobs to read yet
+        replace_method = output.replace_method
```
lipikaramaswamy: Small robustness issue: this direct attribute access can raise `AttributeError` for genuinely legacy loaded results, even though the error message below is trying to give a clearer `ValueError`.

Can we use `getattr(output, "replace_method", None)` here instead?

That preserves current behavior for new `run()` / `preview()` results, and gives hand-built or old pickled results the intended actionable error.

src/anonymizer/engine/replace/replace_runner.py
```diff
@@ -63,7 +88,129 @@ def run(
...
+            # postprocess() converts that to *_valid=None ("Unavailable").
+            new_cols = [c for c in run_result.dataframe.columns if c not in prepared.columns]
+            if new_cols:
+                judged_df = prepared.merge(
+                    run_result.dataframe[[RECORD_ID_COLUMN, *new_cols]],
+                    on=RECORD_ID_COLUMN,
+                    how="left",
+                )
+            else:
+                judged_df = prepared
```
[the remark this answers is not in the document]
memadi-nv: Input must come from `preview()/run()`, not from `evaluate()`.

src/anonymizer/interface/anonymizer.py
```diff
@@ -218,6 +243,83 @@ def preview(
...
+            model_configs=self._model_configs,
+            selected_models=self._selected_models.evaluate,
+        )
+        renamed_trace = _rename_output_columns(replace_result.dataframe, resolved_text_column=text_column)
+        return AnonymizerResult(
+            dataframe=_build_user_dataframe(renamed_trace, resolved_text_column=text_column),
+            trace_dataframe=renamed_trace,
+            resolved_text_column=text_column,
+            failed_records=replace_result.failed_records,
+            replace_method=replace_method,
```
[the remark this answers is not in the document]
memadi-nv: I think it's better to pass on this idea as  records that fail during anonymization are already absent from `preview() / run()` output — they don't appear in `result.dataframe`, only in `result.failed_records`. By the time `evaluate()` runs, those records were never part of its input. Re-surfacing them in `evaluated.failed_records` doesn't restore any user-visible row in `evaluated.dataframe`, and the user can still inspect them on the original result they passed in. wyt @lipikaramaswamy ?

lipikaramaswamy (approved):
lgtm 🚢  refactoring in a separate pr makes sense

PR #160, opened 2026-05-18
chore: upgrade data-designer to 0.6.0

lipikaramaswamy:
## Summary

Upgrades `data-designer` from `0.5.7` → `0.6.0` to pick up the native async engine and bundles a few cleanups that surfaced while verifying the upgrade by re-running the tutorial notebooks end-to-end.

### Changes

- **`pyproject.toml` + `uv.lock`**: bump `data-designer` to `0.6.0`. No code changes
  were required — the async engine is enabled by default in 0.6, so the existing
  `NddAdapter` paths get the benefit transparently.
- **`models.yaml`**: raise `gpt-oss-120b` `max_parallel_requests` from `2` → `16`
  to actually exercise the new async concurrency. Previously gated the rest of
  the pipeline at 2 in-flight LLM calls; bumping cut tutorial notebook runtime
  roughly in half on `gpt-oss-120b`-heavy stages.
- **`Makefile`**: `make convert-notebooks` now passes `--set-kernel anonymizer-venv`
  to `jupytext`. Previously `jupytext` picked an arbitrary registered kernel
  (often a stale one from another project), causing notebooks to silently run
  against the wrong Python environment.
- **`llm_replace_workflow.py`**: replacement-map debug logs no longer emit raw
  PII. The previous `[DEBUG] Replacement map record …` line dumped the full
  `original → synthetic` mapping for every record, leaking the source values
  into log streams whenever DEBUG was enabled. Logs now summarize counts per
  label plus any anomalies (entities unrequested by the prompt, requested
  entities the LLM didn't fill). The warning path for empty/missing maps was
  also tightened to always fire (it previously sat under an `elif` that swallowed it).
- **`docs/notebook_source/*.py`**: consolidate `from anonymizer.config.rewrite import PrivacyGoal`
  into the top-level `from anonymizer import …` line. `PrivacyGoal` is already
  re-exported from the package root, so the deeper import was unnecessary and
  inconsistent with the bundled skill template.

### What's NOT in this PR

- Re-rendered notebook outputs (`docs/notebooks/*.ipynb`) — kept as a separate
  step so the dependency change is reviewable on its own.
- A separate sensitivity-disposition refactor (schema simplification + prompt
  fixes) that surfaced during this verification. Tracking that as a follow-up
  branch since it's a larger change with its own design discussion.

## Test plan

- [x] `make test` — unit tests pass on `data-designer 0.6.0`
- [x] `make convert-notebooks` — all five tutorial notebooks execute end-to-end
      against `gpt-oss-120b` and `nemotron-30b-thinking`
- [x] Verified detection + replace + rewrite pipelines all run under the async engine
- [x] Confirmed no raw PII in `[DEBUG] Replacement map record …` log lines
- [ ] CI green


src/anonymizer/engine/replace/llm_replace_workflow.py
```diff
@@ -169,19 +170,27 @@ def _filter_replacement_map_to_input_entities(
         filtered.append(replacement.model_dump())
 
     if logger.isEnabledFor(logging.DEBUG):
```
andreatnvidia, line 172: suggestion: since this is now a privacy-sensitive logging contract, could add a small `caplog` test that runs this at DEBUG and asserts raw entity values / synthetic values do not show up? The smoke I ran passes, but a committed regression test would keep this from drifting later.

andreatnvidia (approved):
Reviewed the dependency bump, replacement-map logging change, notebook conversion update, and import cleanup. Smoke tests and targeted unit tests pass. Only left one small suggestion to add a committed regression test for the PII-safe debug logging behavior. Otherwise looks good to me!

src/anonymizer/engine/replace/llm_replace_workflow.py
lipikaramaswamy: Added in [bd30625](https://github.com/NVIDIA-NeMo/Anonymizer/pull/160/commits/bd30625dfe65758e71125d7206be54346492cb94)

PR #163, opened 2026-05-19
refactor: derive needs_protection from prot method and fix latent entity

asteier2026:
Changes include:
- Remove `needs_protection` as a stored LLM-emitted field; derive it as a computed property
  (`protection_method_suggestion != leave_as_is`) to eliminate the correlated redundant field that    
  caused LLM consistency failures                                                                     
  - Simplify protection validator to enforce `combined_risk_level` → `protection_method_suggestion` 
  directly (low → leave_as_is, high → not leave_as_is)                                                  
  - Collapse `StrictEntityDispositionSchema` and `StrictSensitivityDispositionSchema` to thin
  subclasses, removing duplicated field definitions                                                     
  - Fix sensitivity disposition prompt: clarify that latent entities with `combined_risk_level='low'`
  must use `leave_as_is`, not `suppress_inference`


tests/engine/test_schemas.py
```diff
@@ -432,3 +428,56 @@ def test_quality_answers_reject_extra_ids() -> None:
...
+def test_strict_sensitivity_disposition_inherits_id_normalization() -> None:
+    schema = StrictSensitivityDispositionSchema.model_validate(
+        {
+            "sensitivity_disposition": [
+                _make_strict_entity(id=1),
+                _make_strict_entity(id=5, entity_label="last_name", entity_value="Smith"),
+            ]
+        }
+    )
+    assert [e.id for e in schema.sensitivity_disposition] == [1, 2]
```
lipikaramaswamy, line 487: One small addition — add an `isinstance` check to lock in the IS-A relationship:

```suggestion
    assert [e.id for e in schema.sensitivity_disposition] == [1, 2]
    assert isinstance(schema, SensitivityDispositionSchema)
```

This specifically defends the `# catches StrictSensitivityDispositionSchema too` comment in `parsers.py`. If someone ever reverts the inheritance, the ID-renumbering assertion alone would still pass (you could redefine `_normalize_ids` on the subclass), but the parser would silently start coercing strict dict payloads through the lenient validator. The `isinstance` line catches that drift.

src/anonymizer/engine/rewrite/rewrite_generation.py
lipikaramaswamy: Greptile flagged this too — the docstring still references the removed `needs_protection=True` terminology:

```suggestion
    """Pre-filter and serialize protected entities (protection_method_suggestion != "leave_as_is") for the rewrite prompt."""
```

(or similar phrasing). The function body is correct; just the docstring is stale.

lipikaramaswamy: I dont know why the suggestion is showing up weirdly though. probably worth doing locally and pushing, versus `Apply suggestion` from this interface

src/anonymizer/engine/rewrite/sensitivity_disposition.py
lipikaramaswamy: Unrelated to the schema collapse but worth flagging since it's in the same file: this example teaches a tag format the LLM will never actually see.

```python
The text below has explicit entities tagged inline using this format: ⟦entity_value|entity_label⟧

Example:
  "A 29-year-old male from Portland" becomes:
  "A ⟦29-year-old|age⟧ ⟦male|gender⟧ from ⟦Portland|city⟧"
```

The real tag notation is chosen per-document in `_choose_tag_notation` (`engine/detection/postprocess.py`) and can be `xml`, `bracket`, `paren`, or `sentinel` — never `⟦…|…⟧`. Every other prompt that references tagged text Jinja-dispatches on `<<TAG_NOTATION>>` (e.g. `rewrite_generation.py:65-68`); this one is the outlier.

Effect on the LLM: it has to guess the bracket style from `<<TAGGED_TEXT>>` and infer that the example's `entity_value` rule (substring before `|`) maps onto whatever the real notation is. 

I can do this in a quick follow up PR too

lipikaramaswamy: So it would be something replacing L74-L90 with
```python
<input_tagged_text>
The text below contains inline entity tags marking identified entities.
{% if <<TAG_NOTATION>> == 'bracket' %}Tags use the format [[entity_value|entity_label]] (e.g. [[Portland|city]]). The substring before "|" is the entity_value — record it EXACTLY, without the [[ ]] brackets.
{% elif <<TAG_NOTATION>> == 'xml' %}Tags use the format <entity_label>entity_value</entity_label> (e.g. <city>Portland</city>). The inner text is the entity_value — record it EXACTLY, without the surrounding tags.
{% elif <<TAG_NOTATION>> == 'paren' %}Tags use the format ((SENSITIVE:entity_label|entity_value)) (e.g. ((SENSITIVE:city|Portland))). The substring after "|" is the entity_value — record it EXACTLY, without the ((SENSITIVE:...)) wrapper.
{% elif <<TAG_NOTATION>> == 'sentinel' %}Tags use the format <<SENSITIVE:entity_label>>entity_value<</SENSITIVE:entity_label>> (e.g. <<SENSITIVE:city>>Portland<</SENSITIVE:city>>). The text between the sentinels is the entity_value — record it EXACTLY, without any <<SENSITIVE:...>> markers.
{% endif %}
Tagged Text:
---
<<TAGGED_TEXT>>
---
</input_tagged_text>
```

And then two small wiring changes are needed in the same file to make the placeholder resolve:

1. Add `COL_TAG_NOTATION` to the imports from `anonymizer.engine.constants` at the top.
2. Add `"<<TAG_NOTATION>>": COL_TAG_NOTATION,` to the `substitute_placeholders(...)` dict at the bottom of `_get_sensitivity_disposition_prompt`.

asteier2026: Made these changes.

lipikaramaswamy (approved):
LGTM, I ran a previously failing workflow as well and didn't see any errors 🤞

PR #168, opened 2026-05-26
feat: Make anonymizer evaluation mode-independent

memadi-nv:
## Summary
- **Public method renamed:** `Anonymizer.evaluate_replace` → `Anonymizer.evaluate`. Verb is no longer mode-dependent.

- **New call shape (single-argument default):**
  ```python
  anonymizer.evaluate(preview)                              # default path
  anonymizer.evaluate(preview, config=EvaluateConfig(...))  # extensible path
  output is positional, config is keyword-only and optional.

- Strategy travels on the result, not in a config. `AnonymizerResult` and `PreviewResult` gained a `replace_method: ReplaceMethod | None = None` field, populated by `run() / preview()`. `evaluate(`) reads it directly — eliminates the "result says Substitute, config says Redact" mismatch risk.
- `EvaluateConfig` is now an empty placeholder. Dropped its replace= / rewrite= fields and the validate_at_least_one_mode validator. Reserved for future evaluation-specific knobs (metric selection, per-judge model/prompt overrides, thresholds). Still exported from the top-level package so a future migration is just adding fields.
- Parameter renamed `result` → `output` in `evaluate()` for clearer naming.
- Clear error for hand-built / pre-tracking outputs. evaluate() raises a descriptive `ValueError` when `output.replace_method` is None, with guidance on how to set it.
- Demo notebook updated (`docs/notebooks/evaluate-replace-demo.ipynb`): all three evaluate(config=..., result=...) sites became single-arg evaluate(preview); `EvaluateConfig` removed from imports; intro paragraph rewritten to explain "strategy travels on the result."
- No test churn. The lower-level `ReplacementWorkflow.evaluate(dataframe, replace_method=...)` is unchanged, so existing tests pass without edits. 818 / 818 green.


## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing

## Documentation

## Related Issues
<!-- Closes #123 -->

### DIFF
--- a/docs/notebooks/evaluate-replace-demo.ipynb
+++ b/docs/notebooks/evaluate-replace-demo.ipynb
@@ -4,20 +4,22 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# evaluate_replace demo\n",
+    "# evaluate demo\n",
     "\n",
     "A clean walkthrough of the new two-step anonymization workflow:\n",
     "\n",
-    "1. `anonymizer.preview(...)` (or `.run(...)`) does **detection + replacement only** — no LLM judges, no evaluation cost.\n",
-    "2. `anonymizer.evaluate_replace(...)` is a **separate, opt-in** call that scores the result with four LLM-as-judge metrics:\n",
-    "   - **Detection Validity** — were the detected entities correct?\n",
-    "   - **Type Fidelity** — does each synthetic preserve the entity class/format? *(Substitute only)*\n",
-    "   - **Relational Consistency** — do cross-entity relationships stay coherent? *(Substitute only)*\n",
-    "   - **Attribute Fidelity** — do salient within-entity attributes (gender, age bucket) survive? *(Substitute only)*\n",
+    "1. `anonymizer.preview(...)` (or `.run(...)`) does **detection + replacement only** \u2014 no LLM judges, no evaluation cost.\n",
+    "2. `anonymizer.evaluate(...)` is a **separate, opt-in** call that scores the result with four LLM-as-judge metrics:\n",
+    "   - **Detection Validity** \u2014 were the detected entities correct?\n",
+    "   - **Type Fidelity** \u2014 does each synthetic preserve the entity class/format? *(Substitute only)*\n",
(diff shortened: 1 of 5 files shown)


memadi-nv:
Since `EvaluateConfig` mirrors the `AnonymizerConfig` in this branch, I don't really see a benefit to it. I think it makes more sense if we just stash strategy on the result; and drop `EvaluateConfig`. 

Keeping `EvaluateConfig` makes sense if we enable custom prompt judges, metric subset or even add the LLM judges in this config rather than the replace yaml file IMO.

Let me know your thoughts on this. @lipikaramaswamy

lipikaramaswamy:
I like moving the public API to `evaluate(...)`, but I agree, we can make the call shape cleaner by not asking users to restate the anonymization mode in `EvaluateConfig`.

Instead of 
  ```python
  evaluated = anonymizer.evaluate(
      config=EvaluateConfig(replace=cfg.replace),
      result=preview,
  )
  ```

we can make the default path this:
  ```python
  evaluated = anonymizer.evaluate(preview)
  ```

and the extensible path can be:

  ```python
  evaluated = anonymizer.evaluate(
      preview,
      config=EvaluateConfig(...),
  )
  ```

with a signature like:

  ```python
  def evaluate(
      self,
      output: AnonymizerResult | PreviewResult,
      *,
      config: EvaluateConfig | None = None,
  ) -> AnonymizerResult:
      ...
  ```
Asking users to pass `EvaluateConfig(replace=cfg.replace)` creates a mismatch risk if the output came from one strategy but the evaluate config says another.

I think `EvaluateConfig` can still be useful, but I’d reserve it for evaluation-specific knobs: selected metrics, judge model/prompt config, thresholds when applicable, etc. If `evaluate()` needs dispatch info now, we can store lightweight mode/strategy metadata on `AnonymizerResult` / `PreviewResult` when `run()` / `preview()` creates them, then have `evaluate()` dispatch from that.

More relevant to #158 but mentioning here - `result` sounds generic as the parameter name; suggest using something like `output` - reads more clearly as the anonymization output to evaluate.

PR #170, opened 2026-05-26
chore: Use uv and pip ecosystem for dependabot

kendrickb-nvidia:
## Summary

Enable dependabot to work with pyproject.toml and uv.lock for scanning and updates. See https://github.com/dependabot/dependabot-core/issues/12609

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing

## Documentation

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,9 +1,32 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# This project is uv-managed (pyproject.toml + uv.lock). The `pip` ecosystem
+# alone parses pyproject.toml but cannot update uv.lock, leaving transitive
+# dependencies invisible to both version updates and security alerts.
+# Pair `pip` and `uv` in a multi-ecosystem group so a single PR updates
+# pyproject.toml and uv.lock together.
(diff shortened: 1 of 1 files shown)


lipikaramaswamy (approved):
worth a shot!

.github/dependabot.yml
```diff
@@ -1,9 +1,32 @@
...
+
 version: 2
+
+multi-ecosystem-groups:
+  python:
+    schedule:
+      interval: "weekly"
+
 updates:
   - package-ecosystem: "pip"
```
mckornfield, line 20: why keep the pip AND uv bits working?

kendrickb-nvidia: From https://github.com/dependabot/dependabot-core/issues/12609#issuecomment-3407284321 (comment 2025-10-15 so relatively recent), says this config will update both pyproject.toml and uv.lock. Not really clear why both are needed to but that's stated there.

More generally though, seems dependabot isn't going to update lower bounds in pyproject.toml which seems to remove a lot of it's value as later in the thread there's:

> To clarify how the updater works: if Dependabot is attempting to bump your dependency and the new version still matches the constraints in your pyproject.toml, we only update the uv.lock file. We won't modify your pyproject.toml unless the constraints need to change.

mckornfield: so sad... can try it though, can't hurt? lol

mckornfield (approved):
give it a tryyy

PR #174, opened 2026-06-01
feat: tolerate small-model drift on detection + replace schemas

lipikaramaswamy:
## Summary

First of two stacked PRs reworking the LLM-facing schemas so small models (gemma4-e2b/e4b, nemotron-3-nano:4b, qwen3.5:4b) no longer drop records when their structured output drifts from the strict contract. This PR covers the **detection** schemas plus the **replace** (Substitute) wire schema; the **rewrite** schemas + server-side disposition follow in stacked PR #175.

This supersedes the schema work in #130, rebased onto current `main` and split for reviewability.

### Detection
- **`RawValidationDecisionSchema` / `ValidationDecisionSchema`**: coerce free-form `decision` prose into `keep`/`reclass`/`drop`, preserve `None` as "no answer" (required by the chunked-merge logic), and drop `value`/`label` from the wire (re-filled downstream from the trusted candidate lookup by `enrich_validation_decisions`).
- **`LatentEntitySchema`**: loosen `category`/`confidence`/`evidence`/`rationale` with before-validators; verbose rationales truncate instead of failing.
- **`LatentEntitiesSchema` + `detection_workflow`**: pad empty latent cells with a sentinel so PyArrow can write the parquet column. *(Latent entities are produced here but only consumed in rewrite mode — they land as inert-but-correct columns until #175 reads them.)*
- **`AugmentedEntitySchema`**: drop `value`/`label` `min_length` — a single empty entry no longer fails the whole augmentation batch (empties are already skipped by `apply_augmented_entities`).

### Replace
- **`EntityReplacementSchema`**: drop `min_length` on `original`/`label`/`synthetic`. This is the LLM `output_format` for Substitute mode, so a single drifted entry used to fail-validate the whole replacement map and drop the record. Empty `original`/`label` are filtered downstream (cannot match a requested entity); an empty `synthetic` is applied as a deletion — privacy-safe (no leak), if utility-poor.

## Test plan
- [x] `make test` green (837 passing on this branch)
- [x] `make format-check` clean
- [x] New `test_small_model_drift.py` (detection drift classes); `test_chunked_validation` updated for the slimmed wire shape; new replace drift tests in `test_llm_replace_workflow`
- [ ] Live small-model run (validation/augmentation/latent + substitute paths) before merge

### DIFF
--- a/src/anonymizer/engine/detection/detection_workflow.py
+++ b/src/anonymizer/engine/detection/detection_workflow.py
@@ -707,6 +707,11 @@ def _pad_empty_latent_column(df: pd.DataFrame) -> pd.DataFrame:
         # in the normal DD path (dict branch), but tolerate a bare-list cell from
         # alternate paths. The downstream reader (_coerce_entity_list) accepts
         # either shape, so we never mix dict and list within one column.
+        # None / NaN (e.g. a row absent from a partial-failure fallback merge,
+        # reintroduced by a pandas reindex) normalizes to the canonical struct
+        # so the column stays parquet-writable.
+        if cell is None or (not isinstance(cell, (dict, list)) and pd.isna(cell)):
+            return {"latent_entities": sentinel}
         if isinstance(cell, dict):
             if not cell.get("latent_entities"):
                 return {**cell, "latent_entities": sentinel}
--- a/src/anonymizer/engine/detection/detection_workflow.py
+++ b/src/anonymizer/engine/detection/detection_workflow.py
@@ -702,6 +702,11 @@ def _pad_empty_latent_column(df: pd.DataFrame) -> pd.DataFrame:
     sentinel = [LatentEntitySchema().model_dump()]
 
     def _fix(cell):
+        # Preserve each cell's existing shape: the column is uniformly the
+        # struct/dict shape ({"latent_entities": [...]}) from LatentEntitiesSchema
+        # in the normal DD path (dict branch), but tolerate a bare-list cell from
+        # alternate paths. The downstream reader (_coerce_entity_list) accepts
+        # either shape, so we never mix dict and list within one column.
         if isinstance(cell, dict):
             if not cell.get("latent_entities"):
                 return {**cell, "latent_entities": sentinel}
--- a/src/anonymizer/engine/detection/detection_workflow.py
+++ b/src/anonymizer/engine/detection/detection_workflow.py
@@ -58,6 +58,7 @@ from anonymizer.engine.schemas import (
     EntitiesByValueSchema,
     EntitiesSchema,
     LatentEntitiesSchema,
+    LatentEntitySchema,
 )
 
 logger = logging.getLogger("anonymizer.detection")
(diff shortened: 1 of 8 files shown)


lipikaramaswamy:
## Live small-model validation

These changes make detection + replacement **tolerate small-model output drift without dropping records**. Validated live against real small models on `build.nvidia.com`. 

### Setup
- **Pipeline:** GLiNER (`nvidia/gliner-pii`) detection → LLM validation + augmentation → `Substitute` replacement → `anonymizer.evaluate()` judges
- **Worker roles** (validator / augmenter / replacement_generator) pointed at a small model; **judge** fixed at `gpt-oss-120b` so verdicts are comparable across runs
- **Dataset:** 4 synthetic PII-dense records (clinical, financial/PII, employment, sensitive-attribute/latent)
- **Pass criterion:** zero dropped rows, zero validation failures (the P1-class bug these schema changes fix)

### Resilience results — the core goal

| Worker model | Size | Detection | Replacement | Rows in→out | Failures | Dropped |
|---|---|---|---|---|---|---|
| `gemma-3n-e2b-it` | ~2B | 0 failed | 0 failed | 4→4 | 0 | **0** |
| `nvidia-nemotron-nano-9b-v2` | ~9B | 0 failed | 0 failed | 4→4 | 0 | **0** |
| `gpt-oss-120b` (default) | ~120B | 0 failed | 0 failed | 4→4 | 0 | **0** |

**Zero record loss at every model size**, even though the 2B drifted heavily. Pre-change, a single drifted field could fail-validate an entire batch. Drifted output was always **privacy-safe** — original PII was replaced, never leaked back.

### Quality (LLM-as-judge, graded by gpt-oss-120b)

| Worker model | Judge passes /16 | Character of failures |
|---|---|---|
| `gemma-3n-e2b` (~2B) | 11 | detection mislabels; one geo mismatch |
| `nemotron-nano-9b` (~9B) | 9 | gender flips, one geo mismatch, mislabels |
| `gpt-oss-120b` (~120B) | 14 | residual gender flip + one detection false-positive |

Failures split into two classes:
- **Model-size-dependent** (geo inconsistency, grammar) — improve with bigger models.
- **Model-independent** (gender preservation on name swaps, occasional detection false-positives like `"meatpacking plant"→company_name`) — present even at 120B, so **not regressions from this PR**.

## Replacement-prompt clarification (included here)

`engine/replace/llm_replace_workflow.py`: per-label hints now render as `such as {examples}` and instruct the model to *generate a new realistic value of that kind, not reuse the examples or copy the reference text*. This keeps small models from emitting the reference text instead of a real value. Confirmed on the 2B run — `type_fidelity` and `attribute_fidelity` both score 4/4.=

lipikaramaswamy:
Addressed the remaining Greptile note (non-blocking) from the review summary in c15896c: `_clamp_evidence` previously returned `[]` for a bare-string `evidence` value, silently dropping the quote when a small model emits a single string instead of a one-element list. It now wraps a bare string into a one-element list — consistent with the drift-tolerance strategy — with a regression test (`test_bare_string_evidence_wrapped_not_dropped`).

The three inline comments (P0 `latent_sensitive_attribute`, P1 `_fix` None/NaN guard, P2 `proposed_label` docs) were already handled in 926bf8c / b287340 / 81ecba0.

PR #177, opened 2026-06-03
feat: anonymizer measurement instrumentation and benchmark tooling

binaryaaron:
## Summary

- Adds in-repo measurement sessions, stage timers, DataDesigner workflow metrics, direct model workflow metrics, and sanitized per-record safety metrics.
- Records sanitized `evaluation_record` rows when benchmark replace configs set `evaluate: true`, preserving judge verdict booleans and invalid-item counts without persisting evaluated dataframes, raw judge traces, prompts, entity values, or replacement strings.
- Refactors measurement internals into the `anonymizer.measurement` package while preserving the public `anonymizer.measurement` import surface.
- Adds the core benchmark runner for repeatable workloads: suite preflight validation, row slicing, case retries, per-case raw measurement shards, combined `measurements.jsonl`, table export, detection-artifact sidecars, raw DataDesigner message trace capture, and sanitized DataDesigner scheduler task traces.
- Uses DataDesigner native LLM column tracing for standard LLM columns, adds a temporary Anonymizer private model-registry/facade shim for model-backed `CustomColumnConfig` traces, and emits safe `dd_trace_coverage` records that show native, private-facade, and unsupported coverage.
- Adds the front-door measurement table exporter plus first-order benchmark and detection-artifact analysis tools, including case/group rollups for sanitized replace judge evaluation rows.
- Factors shared measurement-tool support into `tools/measurement/measurement_tools/` for CLI logging, export formats, table writing, manifests, and small aggregation helpers. Scripts keep their row models and metric semantics local.
- Documents the in-repo observability system and puts the `measurements.jsonl` to Parquet/CSV/JSONL workflow front and center in the measurement tool README.
- Refactors benchmark-output tests to use checked-in measurement fixtures, and uses the repo synthetic biography sample for valid benchmark workload tests.

This PR intentionally does not carry the larger derivative strategy/probe/comparison tools. Those are split to a stacked follow-up so this PR stays focused on measurement capture, export, and the basic benchmark harness.

## Stack

- #177: core measurement instrumentation, benchmark runner, export, and basic analysis
- Follow-up stacked PR: derivative measurement probes, strategy comparisons, trace analysis, and signature-delta tools
- #184: local structured substitute replacement-map strategy
- #183: regex-backed detection benchmark strategies

## Alignment

Distributed DataDesigner execution is outside this PR. Detection export APIs, such as the work in #182, should build configs for external runtimes; the measurement tools here should consume the resulting measurement JSONL, detection artifacts, and trace sidecars.

## Validation

Latest checks after benchmark analysis evaluation rollups:

- `uv run --frozen pytest tests/tools/test_measurement_tools.py tests/tools/test_benchmark_output_analysis.py -q`
  - Result: 39 passed, 6 existing DataDesigner model-config deprecation warnings
- `uv run --frozen pytest tests/tools/test_benchmark_output_analysis.py -q`
  - Result: 10 passed
- `uv run --frozen ruff check tools/measurement/analyze_benchmark_output.py tests/tools/test_benchmark_output_analysis.py`
- `uv run --frozen ruff format --check tools/measurement/analyze_benchmark_output.py tests/tools/test_benchmark_output_analysis.py`
- `uv run tools/codestyle/format.sh --check`
- `tynav --ty-bin /root/.local/share/uv/tools/ty/bin/ty diagnostics tools/measurement/analyze_benchmark_output.py`
  - Result: no diagnostics for the edited file; existing `pyproject.toml` deprecated `tool.ty.src.root` warning remains
- `git diff --check`

Earlier checks after sanitized evaluation metrics:

- `uv run --frozen pytest tests/test_measurement.py tests/tools/test_measurement_tools.py -q`
  - Result: 75 passed, 9 existing DataDesigner model-config deprecation warnings
- `uv run --frozen ruff check src/anonymizer/measurement/records/row.py src/anonymizer/measurement/__init__.py tools/measurement/run_benchmarks.py tests/tools/test_measurement_tools.py`
- `uv run --frozen ruff format --check src/anonymizer/measurement/records/row.py src/anonymizer/measurement/__init__.py tools/measurement/run_benchmarks.py tests/tools/test_measurement_tools.py`
- `tynav --ty-bin /root/.local/share/uv/tools/ty/bin/ty diagnostics src/anonymizer/measurement/records/row.py`
  - Result: no diagnostics for the edited file; existing `pyproject.toml` deprecated `tool.ty.src.root` warning remains
- `git diff --check`

Earlier checks after the custom-column DD trace shim:

- `uv run --frozen pytest tests/test_measurement.py tests/tools/test_measurement_tools.py tests/tools/test_benchmark_output_analysis.py -q`
  - Result: 71 passed, 6 existing DataDesigner model-config deprecation warnings
- `uv run --frozen ruff check src/anonymizer/engine/ndd/adapter.py src/anonymizer/measurement/sinks.py tests/test_measurement.py`
- `uv run --frozen ruff format --check src/anonymizer/engine/ndd/adapter.py src/anonymizer/measurement/sinks.py tests/test_measurement.py`
- `uv run tools/codestyle/format.sh --check`
- `git diff --check`

Earlier checks after the split:

- `uv run --frozen pytest tests/test_measurement.py tests/engine/test_ndd_adapter.py tests/tools/test_measurement_tools.py tests/tools/test_benchmark_output_analysis.py tests/tools/test_detection_artifact_analysis.py -q`
  - Result: 82 passed, 9 existing DataDesigner model-config deprecation warnings
- `uv run --frozen ruff check tools/measurement/run_benchmarks.py tools/measurement/export_measurements.py tools/measurement/analyze_benchmark_output.py tools/measurement/analyze_detection_artifacts.py tools/measurement/measurement_tools tests/tools/test_measurement_tools.py tests/tools/test_benchmark_output_analysis.py tests/tools/test_detection_artifact_analysis.py`
- `uv run tools/codestyle/format.sh --check`
- `git diff --cached --check`
- CLI smoke:
  - `uv run python tools/measurement/run_benchmarks.py --help`
  - `uv run python tools/measurement/export_measurements.py --help`
  - `uv run python tools/measurement/analyze_benchmark_output.py --help`
  - `uv run python tools/measurement/analyze_detection_artifacts.py --help`

Earlier branch checks also covered docs build, benchmark dry-run validation for `tools/measurement/examples/repo-data-smoke.yaml`, shell syntax validation for the DD-trace smoke script, and `tynav` diagnostics on the measurement module.

## Dogfood

Dogfood with local vLLM endpoint:

- Endpoint: `http://nemotron-3-super-h100-svc.aagonzales-dev.svc.cluster.local:8000/v1`
- Model: `nvidia/nemotron-3-super`
- Output: `/tmp/anonymizer-dogfood-mini-h100`
- Result: 2/2 cases completed, 0 errors
- `biographies__biographies-redact-default__r000`: ~12.0s, 21 final entities
- `legal__legal-hash-agent-labels__r000`: ~9.1s, 14 final entities
- `measurements.jsonl`: 14 records
- task traces: 10 records per case
- table export: 5 tables from 14 records (`run`, `dd_trace_coverage`, `ndd_workflow`, `stage`, `record`)

## Notes

- Checked-in benchmark suites use repository-relative paths and env-backed runtime config; they avoid machine-specific endpoints and absolute local paths.
- DD message traces, raw case shards, and DataDesigner artifacts are sensitive run artifacts and may contain prompts, model outputs, secrets, or PII.
- DataDesigner scheduler task traces are sanitized timing sidecars. They include queue/execution/total durations and error presence, but intentionally omit raw DD error strings.
- Optional replace judge evaluation metrics are sanitized into `evaluation_record` rows. They include verdict booleans and invalid-item counts, but not original text, entity values, replacement values, raw judge outputs, prompts, or model responses.
- Standard DD message tracing uses native DataDesigner trace side effects for `LLMTextColumnConfig` and `LLMStructuredColumnConfig`.
- Model-backed `CustomColumnConfig` traces currently use a temporary Anonymizer shim that instruments the per-run private DataDesigner model registry and returned model facades. This is intentionally documented as brittle and should be replaced by a public DataDesigner model-call trace sink. No DataDesigner issue or PR has been opened from this PR.
- Runtime timings, retry counts, token counts, and local endpoint names are environment-dependent and should not be treated as portable fixtures.


src/anonymizer/engine/ndd/adapter.py
```diff
@@ -225,3 +262,326 @@ def _detect_missing_records(
...
+                messages=messages,
+                response=response,
+                elapsed_sec=time.perf_counter() - started,
+                status=status,
+                error_type=error_type,
+                is_async=True,
+            )
+
+
+def _active_trace_collector(expected_collector: Any) -> Any | None:
```
andreatnvidia: Might be worth seeing if we can lean a bit more on DataDesigner’s built-in trace/observability hooks here.

For the DD LLM columns, `with_trace=TraceType.LAST_MESSAGE/ALL_MESSAGES` already gives `{column}__trace` side-effect columns in DD’s native format. That might be a nicer default because it keeps us closer to DD’s supported surface and avoids wrapping `ModelFacade` directly.

The custom path here may still be needed for things DD does not trace directly, but I think there’s a possible gap around calls that happen in plain worker threads, like chunked validation’s `ThreadPoolExecutor` path. Since `current_collector()` is a `ContextVar`, those worker calls may not see the active collector and could miss some trace records.

Maybe a good split is:
- DD `with_trace` for normal DD LLM-column message traces
- explicit handling for custom generator calls like chunked validation
- DD async scheduler traces / newer scheduler and request-admission observability events as optional timing sidecars

[Eric’s DD async scheduling / observability work](https://github.com/NVIDIA-NeMo/DataDesigner/pull/661) might be useful for the benchmark side too. The scheduler/request-admission events could help explain whether a run is scheduler-bound, request-admission-bound, provider-bound, or model-bound. I don’t think those replace message traces, but they could be a nice complement if there’s a stable way to surface them from DD.

tools/measurement/run_benchmarks.py
```diff
@@ -0,0 +1,1491 @@
...
+
+def build_cases(spec: BenchmarkSpec) -> list[BenchmarkCase]:
+    matrix = spec.matrix or _cross_product_matrix(spec)
+    return [
+        BenchmarkCase(
+            suite_id=spec.suite_id,
+            workload_id=entry.workload,
+            config_id=entry.config,
+            repetition=repetition,
+            case_id=f"{entry.workload}__{entry.config}__r{repetition:03d}",
```
andreatnvidia, line 250: Small edge case I noticed: duplicate matrix entries seem like they could generate the same `case_id`, since the ID only includes workload, config, and repetition. For example, two identical entries would both become something like `w__c__r000`, so their raw measurements, traces, and artifact paths could collide.

Might be worth either rejecting duplicate generated case IDs during validation, or adding a stable matrix-entry index to the ID.

tools/measurement/run_benchmarks.py
```diff
@@ -0,0 +1,1491 @@
...
+    config: ConfigSpec,
+    execution: _CaseExecution,
+) -> Path | None:
+    detection_artifact_path = _export_case_detection_artifacts_if_requested(
+        contexts,
+        paths.artifact_output_path,
+        case=case,
+        artifact_snapshot=paths.artifact_snapshot,
+    )
+    detection_artifact_path = _trace_final_artifact_path_if_requested(
```
andreatnvidia: Small export semantics thing: I think `--export=false` can still write a per-case detection artifact here. `_case_run_paths()` sets `artifact_snapshot=None`, but final-trace strategies can still call `_trace_final_artifact_path_if_requested()` and write `paths.artifact_output_path` when a trace dataframe is present.

Maybe `--no-export` should suppress this sidecar too, so export behavior stays predictable.

src/anonymizer/measurement.py
```diff
@@ -0,0 +1,1505 @@
...
+            "entity_relaxed_label_compatible_detected_tp_count": None,
+            "entity_relaxed_precision": None,
+            "entity_relaxed_recall": None,
+            "entity_relaxed_f1": None,
+            "entity_relaxed_label_compatible_precision": None,
+            "entity_relaxed_label_compatible_recall": None,
+            "entity_relaxed_label_compatible_f1": None,
+        }
+
+    predicted = _entity_identity_set(final_entities)
```
andreatnvidia: One metrics edge case: exact ground-truth matching seems to collapse duplicate entities because `_entity_identity_set()` turns `(value, label)` pairs into a set. If ground truth has two identical `Alice/person` occurrences and the output detects only one, I think this would report no false negative and recall would stay at 1.0.

A `Counter`/multiset match, or span-aware identity when offsets are available, might make this more accurate for repeated values.

andreatnvidia:
This overlaps with my benchmark CI PR #162 enough that I’m happy to close mine and let this be the main benchmark tooling direction.

The one thing I’d want to preserve from #162 is the CI/workflow shape: a manual GitHub Actions workflow, `NVIDIA_API_KEY` setup, benchmark artifact upload, and a step summary. This PR has the more complete runner/measurement stack, so it probably makes sense to retire my bespoke `scripts/benchmark_ci.py` path and have the workflow call `tools/measurement/run_benchmarks.py` instead.

src/anonymizer/engine/ndd/adapter.py
andreatnvidia: For timing in particular, async traces can be used doing something like:

```python
run_config = RunConfig(async_trace=True)
# pass through the DD config/builder path

result = data_designer.preview(...)

task_traces = result.task_traces
```

Those traces give per-task timing/status, including when each task is dispatched, when it actually starts running, how long it waits in scheduler/admission queues, and when it completes. That seems useful for benchmark runs because it gives a much more precise view of where time is going than aggregate model-call totals.

So message traces could stay focused on prompt/response debugging, while async traces help explain queueing, scheduling, and task-level latency.

tools/measurement/run_benchmarks.py
binaryaaron: addresesd

binaryaaron: addresesd

src/anonymizer/engine/ndd/adapter.py
```diff
@@ -225,3 +302,352 @@ def _detect_missing_records(
...
+    if isinstance(tool_calls, list):
+        return [getattr(tool_call, "__dict__", tool_call) for tool_call in tool_calls]
+    return []
+
+
+def _record_dd_task_traces(*, workflow_name: str, collector: Any | None, task_traces: list[Any]) -> None:
+    if collector is None or not collector.dd_task_trace_enabled:
+        return
+    for task_trace in task_traces:
+        collector.record_dd_task_trace(
```
andreatnvidia: For the offline replace-mode async profile, the DD task traces were almost enough to generate the Gantt chart, but the current records only have intra-task durations (`queue_wait_sec`, `execution_sec`, `total_sec`). They do not say where each task sat in the workflow timeline.

Could we add sanitized relative offsets here, computed from the first task dispatch in this trace batch? For example: `dispatched_offset_sec`, `slot_acquired_offset_sec`, and `completed_offset_sec`, each as `timestamp - min(dispatched_at)`. That would let downstream analysis reconstruct overlap/concurrency without storing prompts, text, outputs, or raw errors.

src/anonymizer/engine/ndd/adapter.py
```diff
@@ -225,3 +302,352 @@ def _detect_missing_records(
...
+
+
+def _native_dd_trace_type() -> TraceType:
+    # Preserve Anonymizer's existing dd_trace=last_message semantics: the trace
+    # sink records the final prompt message and response separately, while DD's
+    # native LAST_MESSAGE side effect only keeps the final assistant message.
+    return TraceType.ALL_MESSAGES
+
+
+def _column_has_untraced_model_calls(column: ColumnConfigT) -> bool:
```
andreatnvidia: This coverage record is useful, and the offline profile showed why the gap matters: the dominant cost was `_validation_decisions`, which is a `CustomColumnConfig` calling a model internally.

The coverage record can tell us that a custom model-backed column was untraced, but not what happened inside it. For a follow-up, it would be useful to add sanitized per-call records for custom-column model calls from the generator or a lower-level DD hook: workflow, column/purpose, alias/model/provider, elapsed or relative start/end, ok/error_type, and prompt size. No prompt or response content needed.

Without that, benchmark output can say entity detection used GPT tokens overall, but cannot split validation vs augmentation or explain why validation dominated.

andreatnvidia:
One more related follow-up, since GitHub does not let me anchor this on the unchanged `chunked_validation._dispatch_chunk()` call site in this PR diff:

`_dispatch_chunk()` was the main blind spot in the offline replace-mode async profile. On the biographies sample, each measured row made one validation call through that path, and those calls dominated pipeline wall time.

As an incremental path before a general DD model-call hook exists, it would be useful to record a sanitized `validator_chunk_model_call` event around the `facade.generate()` call there. Useful fields would be `alias`, `chunk_index`, `attempt_index`, `elapsed_sec`, `ok`, `error_type`, and `prompt_char_count` for the final prompt after `PydanticResponseRecipe` is applied. Token usage does not appear to be available from `facade.generate()` directly, so the workflow-level aggregate usage probe can keep covering that part.

src/anonymizer/engine/ndd/adapter.py
```diff
@@ -225,3 +262,308 @@ def _detect_missing_records(
...
+    return list(dict.fromkeys(alias for alias in aliases if alias))
+
+
+def _as_alias_list(raw: Any) -> list[str]:
+    if raw is None:
+        return []
+    if isinstance(raw, str):
+        return [raw]
+    if isinstance(raw, (list, tuple, set)):
+        return [str(item) for item in raw if str(item)]
```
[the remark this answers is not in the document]
binaryaaron: Outdated after the tracing refactor.

tools/measurement/run_benchmarks.py
```diff
@@ -0,0 +1,1226 @@
...
+        return result.model_dump_json(indent=2)
+    completed = sum(case.status == CaseStatus.completed for case in result.cases)
+    errored = sum(case.status == CaseStatus.error for case in result.cases)
+    planned = sum(case.status == CaseStatus.planned for case in result.cases)
+    if planned and completed == 0 and errored == 0:
+        return f"Planned {planned} case(s); output={result.output_dir}"
+    return f"Ran {completed}/{len(result.cases)} case(s); errors={errored}; output={result.output_dir}"
+
+
+def _run_tags(case: BenchmarkCase, spec: BenchmarkSpec) -> dict[str, Any]:
```
lipikaramaswamy, line 1089: The benchmark runner currently hardcodes `run_tags` to `suite_id`, `workload_id`, `config_id`, `repetition`, and `case_id`. Since the measurement core already supports generic `run_tags`, can we add a runner-level way to pass arbitrary tags?

I’m thinking about running benchmarks from our internal GitLab repo, where we’ll want to stamp metadata like `anonymizer_ref`, `commit_sha`, `benchmark_suite_ref`, `benchmark_suite_commit_sha`, and `pipeline_id` into every measurement row.

src/anonymizer/engine/ndd/adapter.py
```diff
@@ -86,7 +116,34 @@ def run_workflow(
...
+            collector=collector,
+        )
+        usage_probe = _DataDesignerUsageProbe(
+            self._data_designer,
+            enabled=collector is not None,
+            collector=collector,
+            workflow_name=workflow_name,
+            private_trace_columns=private_trace_columns,
+        )
+        _record_dd_trace_coverage(
```
lipikaramaswamy: Trace recording happens inside the DataDesigner create/preview try block. Since the benchmark runner uses `fail_on_write_error=True`,  a failure while writing the optional trace output file could be caught here and re-raised as `AnonymizerWorkflowError`, making it look like the model workflow failed.

Should we move trace recording outside the DataDesigner execution failure boundary, or somehow ensure trace-write failures are reported as trace/measurement failures rather than workflow failures?

tools/measurement/run_benchmarks.py
```diff
@@ -0,0 +1,1226 @@
...
+        run_id=case.case_id,
+        run_tags=_run_tags(case, spec),
+        streaming=True,
+        keep_records=False,
+        dd_trace=dd_trace.value,
+        dd_trace_path=trace_path,
+        dd_task_trace_path=task_trace_path,
+        fail_on_write_error=True,
+    )
+    with configured_measurement_session(measurement):
```
lipikaramaswamy, line 815: The runner currently only calls `anonymizer.run(...)`. That means replace mode (and soon to be rewrite mode) LLM judge quality from `Anonymizer.evaluate()` is not benchmarked. Can we add an explicit optional evaluate step to the benchmark spec?

tools/measurement/examples/repo-data-smoke.yaml
```diff
@@ -0,0 +1,27 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+suite_id: repo-data-smoke
```
lipikaramaswamy, line 4: The smoke suite omits `model_providers` and `model_configs`, so benchmark runs depend on Anonymizer's default provider config (which is Data Designer's default - build.nvidia.com). Can we make provider selection explicit in the benchmark docs/suite examples, and recommend that benchmark suites set model_providers for reproducibility?

lipikaramaswamy: I also just opened a related PR #189

lipikaramaswamy:
This is awesome! Thank you @binaryaaron for setting up measurement 🤩

Took me a while, but I reviewed the measurement core, benchmark runner, exporters/analyzers, benchmark CI workflow, and the Anonymizer/NDD instrumentation. I also tried running the smoke suite locally. Some notes from that below - 

The dry-run passed and planned 2 cases. Running the default smoke against build/integrate completed the biographies/redact case and produced the expected measurement outputs: run metadata, stage timings, NDD workflow request/token usage, tokens/sec, and per-record entity/replacement counts.

The legal/hash case initially failed on a transient `openai/gpt-oss-120b` health-check rate limit. I reran that case with explicit build provider/model config, `skip_health_check: true`, and lower model parallelism, and it completed successfully:

  - 5/5 rows completed
  - ~69s elapsed
  - `observed_total_tokens`: 38168
  - `observed_tokens_per_sec`: ~553
  - outputs included `measurements.jsonl` and normalized parquet tables for `run`, `ndd_workflow`, `stage`, and `record`

A few things I think would be good to have before merge:

  1. Can we add runner-level support for arbitrary `run_tags`? I left an inline comment on this, but for our GitLab flow we’ll want to stamp metadata like `anonymizer_ref`, `commit_sha`, `benchmark_suite_ref`, `benchmark_suite_commit_sha`, and `pipeline_id` into every measurement row.

  2. Can we make provider/model selection explicit in the benchmark docs/examples? Model IDs are provider-specific: build/integrate uses names like `openai/gpt-oss-120b`, while internal inference uses names like `nvidia/openai/gpt-oss-120b`, and other providers can use entirely different names. I think benchmark suites should treat `model_configs` and `model_providers` as a matched pair for reproducibility.

  3. Can we clarify the recommended health-check behavior for benchmark/smoke runs? The smoke can fail before producing useful measurements if a provider health check hits rate limits. It would be helpful to document when benchmark suites should use `skip_health_check`, retries/backoff, and lower parallelism, perhaps in AGENTS.md.

  4. Can we add an optional evaluate step to the runner if quality benchmarking is intended to be in scope? Right now the runner measures `anonymizer.run(...)`, but does not call `Anonymizer.evaluate(...)`, so LLM judge quality metrics are not produced or stored by the benchmark runner.

  5. Can we document the emitted record types and key fields? The outputs are useful, but downstream consumers will need a stable contract for `run`, `ndd_workflow`, `stage`, and `record` rows.

src/anonymizer/engine/ndd/adapter.py
binaryaaron: Addressed in `3afb44c`. Native DD trace recording and task trace recording now happen outside the DataDesigner create/preview failure boundary, so sidecar write errors surface as trace/measurement write errors rather than `AnonymizerWorkflowError`. I also kept private facade traces buffered during the DD call and flush them after success or after workflow failure without letting trace sidecar failures mask the original workflow error.

tools/measurement/run_benchmarks.py
binaryaaron: Addressed in `3afb44c`. `ConfigSpec` now has an explicit `evaluate: true` option for replace-mode configs, and the runner calls `anonymizer.evaluate(result)` after `anonymizer.run(...)` when enabled. The spec rejects `evaluate` on rewrite configs for now so the behavior stays explicit.

tools/measurement/examples/repo-data-smoke.yaml
binaryaaron: Addressed in `3afb44c`. The smoke benchmark example now references explicit `model_configs` and `model_providers` files, and the README/docs call out treating those as a matched pair for reproducibility. I also added preflight validation for provider config shape so bad provider YAML fails before an expensive run starts.

tools/measurement/run_benchmarks.py
binaryaaron: Addressed in `3afb44c`. The benchmark spec now supports top-level `run_tags`, validates that callers cannot override reserved benchmark tags, and merges those tags into every per-case `MeasurementConfig.run_tags` row alongside `suite_id`, `workload_id`, `config_id`, `repetition`, and `case_id`.

binaryaaron:
Thanks for the detailed run notes. I addressed the merge-blocking items in `3afb44c`: runner-level `run_tags`, explicit provider/model examples and docs, optional replace-mode `evaluate`, and documentation for emitted record types/key fields. The runner already has retries/backoff, and the example provider config shows `skip_health_check` so smoke runs can avoid provider health-check rate-limit noise when appropriate.

binaryaaron:
Agreed that `_dispatch_chunk()` remains an important observability gap. I did not add a dedicated `validator_chunk_model_call` event in this PR because the current branch now leans on DD/native/private model-call tracing where possible, and the validator chunk path deserves a small focused follow-up. I would keep the proposed sanitized fields: alias, chunk index, attempt index, elapsed time, ok/error type, and final prompt character count.

src/anonymizer/engine/ndd/adapter.py
```diff
@@ -225,3 +358,711 @@ def _detect_missing_records(
...
+                    usage=None,
+                )
+
+    existing_trace_columns = [column_name for column_name in trace_column_names if column_name in output_df.columns]
+    if not existing_trace_columns:
+        return output_df
+    return output_df.drop(columns=existing_trace_columns)
+
+
+def _native_trace_messages(value: Any) -> list[dict[str, Any]]:
```
lipikaramaswamy, line 979: I tried `run-repo-data-smoke-with-dd-traces.sh` with explicit provider/model config. The benchmark completed and wrote `trace_path` values in `summary.json`, and the measurement output included `dd_trace_coverage`, but the trace JSONL sidecars were empty.

The DataDesigner artifact parquet did contain trace columns like `_raw_detected_entities__trace` and `_augmented_entities__trace`. When read back, those trace values were `numpy.ndarray`, but `_native_trace_messages()` only accepts `list`, so real trace values get treated as empty.

Should we make this accept list-like trace containers and perhaps add a regression test? Otherwise `--dd-trace` can look enabled while silently writing empty sidecar files.

src/anonymizer/engine/ndd/adapter.py
binaryaaron: Outdated after the current branch revisions; resolving.

src/anonymizer/engine/ndd/adapter.py
binaryaaron: Outdated after the current branch revisions; resolving.

src/anonymizer/engine/ndd/adapter.py
binaryaaron: will take a look

src/anonymizer/engine/ndd/adapter.py
```diff
@@ -225,3 +358,710 @@ def _detect_missing_records(
...
+            continue
+
+        if _column_has_private_facade_model_calls(column):
+            configured_columns.append(_custom_column_with_trace_context(column))
+            private_trace_columns.append(_PrivateFacadeTraceColumn(column_name=column.name))
+            continue
+
+        configured_columns.append(column)
+
+    return configured_columns, native_trace_columns, private_trace_columns, unsupported_columns
```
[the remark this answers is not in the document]
binaryaaron: Fixed in 1d88389. The trace plan now appends non-native/non-private-facade columns to , and  covers the public  counters.

.github/workflows/benchmark-ci.yml
```diff
@@ -0,0 +1,174 @@
...
+        description: "Capture DataDesigner message traces"
+        required: true
+        type: choice
+        options:
+          - "none"
+          - "last-message"
+          - "all-messages"
+        default: "none"
+      dd_task_trace:
+        description: "Capture sanitized DataDesigner scheduler task traces"
```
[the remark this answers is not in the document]
binaryaaron: Fixed in 00c7293. The workflow, helper script, and README now use the runner enum values ( / ), and  asserts the workflow choices stay aligned with .

src/anonymizer/engine/ndd/adapter.py
binaryaaron: Fixed in 00c7293.  now accepts list-like trace containers such as NumPy object arrays while still rejecting mappings/strings, and the native DD trace regression uses a NumPy array to mirror the parquet readback shape you saw.

tools/measurement/run_benchmarks.py
```diff
@@ -0,0 +1,1246 @@
...
+        result = anonymizer.run(
+            config=anonymizer_config,
+            data=input_data,
+        )
+        if config.evaluate:
```
lipikaramaswamy, line 820: I think `evaluate: true` currently captures the judge workflow execution for timing/token accounting, but not the structured judge-quality results. Here we call `anonymizer.evaluate(result)` and discard the returned evaluated result, while record-level metrics are emitted during `run()` before evaluation. Can we persist these evals?

binaryaaron: Addressed in 2a49d18. When a replace benchmark config sets evaluate: true, the runner now records sanitized evaluation_record rows from the evaluated trace dataframe. Those rows include judge verdict booleans and invalid-item counts, and the new regression test covers that original text, entity values, replacement values, raw judge prompts/responses, and invalid finding payloads do not land in either the measurement JSONL or exported parquet.

tools/measurement/analyze_benchmark_output.py
```diff
@@ -0,0 +1,1447 @@
...
+def _case_ids(*frames: pd.DataFrame) -> list[str]:
+    values: set[str] = set()
+    for dataframe in frames:
+        for column in ("run_tags.case_id", "case_id", "run_id"):
+            if column in dataframe.columns:
+                values.update(str(value) for value in dataframe[column].dropna().tolist())
+    return sorted(values)
+
+
+def _build_case_row(
```
lipikaramaswamy, line 449: Ok, now that `evaluate: true` emits `evaluation_record` rows, the analyzer can also roll those into `case_analysis`/`group_analysis`, in addition to exporting `evaluation_record.parquet`.

I mocked this up locally against the latest PR head for my runs on slurm. The implementation was pretty small:

  - add `_EVALUATION_ROLLUPS = ((prefix, valid_col, invalid_count_col), ...)`
  - in `_build_case_row()`, add `evaluation_rows = _records_of_type(measurement_rows, "evaluation_record")`
  - add case fields like `detection_judged_record_count`, `detection_valid_record_count`, `detection_valid_rate`, `detection_invalid_entity_count`
  - add group fields like `sum_detection_judged_record_count`, `sum_detection_valid_record_count`, `micro_detection_valid_rate`, `sum_detection_invalid_entity_count`
  - compute group rates as micro rates from summed counts, not medians of case rates

Codex added a focused test with two cases in one group to verify per-case rates and grouped micro rates. Useful.

Anyway, this keeps `evaluation_record.parquet` as the source of truth, but makes the built-in analysis tables immediately useful for benchmark reports without reading raw judge outputs or evaluated trace dataframes!

lipikaramaswamy (approved):
Looks great, just small comment on the analysis tables :) Thanks!!

tools/measurement/run_benchmarks.py
binaryaaron: Addressed in 2a49d18. The benchmark runner now records sanitized evaluation_record rows when evaluate: true is set, preserving judge verdicts and invalid-item counts while excluding evaluated dataframe content, raw judge traces, prompts, entity values, and replacement strings.

tools/measurement/analyze_benchmark_output.py
binaryaaron: Addressed in 6f843af. evaluation_record rows now roll into case_analysis and group_analysis: case rows include judged/valid counts, valid rates, and invalid counts; group rows use summed counts for micro valid rates. Added a focused two-case regression test and README coverage for the new fields.

PR #178, opened 2026-06-03
chore: bump data-designer

lipikaramaswamy:
## Summary
- Relax the `data-designer` pin from `==0.6.0` to `>=0.6.1,<0.7` so installs pick up 0.6.1 patch releases on the same minor line.
- Refresh `uv.lock` (resolves `data-designer`, `data-designer-config`, and `data-designer-engine` at 0.6.1).

No Anonymizer source changes; chunked validation stays on the existing sync `ThreadPoolExecutor` + `facade.generate()` path. @andreatgretel could you take a look at this in `chunked_validation.py` when back to see if we still need it? 

## Test plan
- [x] `make test` (822 passed locally)
- [x] Smoke-run a notebook or preview against `build.nvidia.com` after merge - actually it works before too so we should be fine.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ requires-python = ">=3.11"
 license = "Apache-2.0"
 
 dependencies = [
-    "data-designer==0.6.0",
+    "data-designer>=0.6.1,<0.7",
     "pydantic>=2.9,<3",
     "cyclopts>=3",
     "pygments>=2.20.0",
--- a/uv.lock
+++ b/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
+revision = 2
 requires-python = ">=3.11"
 resolution-markers = [
     "python_full_version >= '3.14'",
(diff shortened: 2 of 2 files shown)


memadi-nv (approved):
- Pulled this down and ran the suite against 0.6.1 — **822 passed**, so the bump doesn't break anything on our end. One gotcha: my venv was still on 0.6.0 despite the updated lock, so I had to `uv sync` first — worth confirming the `make test` in the plan was run *after* syncing, otherwise it's just re-testing 0.6.0.

A couple of non-blocking flags:

- 0.6.1 spams a deprecation warning: `ModelConfig.provider=None is deprecated and will be required in a future release (#589)`. We create `ModelConfig` without `provider` in several default configs/tests, so **0.7 will likely break us**. The `<0.7` cap saves us for now — worth a follow-up to set `provider` explicitly before that bump.
- Unit tests stub the NDD adapter, so none of this hits DD's real LLM/HTTP path. Still want the smoke `preview()` against build.nvidia.com (already in the plan — just the part that actually matters here).


Otherwise LGTM 👍

PR #179, opened 2026-06-03
feat: add support for long-context documents

eurekayuan:
## Summary
Several stages embedded the whole document in a single prompt and hit DataDesigner's 512K (`MAX_RENDERED_LEN`) render cap, failing outright on long inputs. Every such stage is now windowed: each chunked generator renders its own per-window prompt and calls the model directly, bypassing the cap. Stages keep a single-call **fast path** when the rendered prompt already fits, so short-document behavior is unchanged.

### Per-stage windowing
- **Detection** (`chunked_detection.py`, new): Overlapping fixed-size character windows; each window is a raw text slice sent to the detector. Per-window offsets are rebased to global, boundary-touching spans are dropped, and overlaps are resolved (`resolve_overlaps`).
- **Validation** (`chunked_validation.py`): Not a text window — batches candidate entities (≤100 per call), each with a ±500-character excerpt. Calls run in parallel across the validator pool with round-robin + failover. Decisions are merged per row; the row is dropped only if every pool member fails.
- **Augmentation** (`chunked_augmentation.py`): Overlapping character windows over tagged text plus seed JSON. A window dynamically shrinks if its rendered prompt exceeds the cap. Outputs are unioned and deduped by `(value, label)`.
- **Latent** (`chunked_latent.py`): Same mechanism as augmentation (rewrite mode only); deduped by `(label, value)`.
- **Substitution map** (`chunked_replace.py`): Abutting newline-aligned windows, no overlap. Each chunk carries the accumulated replacement map and a rolling summary, proposing replacements only for new entities so mappings stay consistent across chunks.
- **Rewrite generation** (`chunked_rewrite.py`): Abutting newline-aligned windows, no overlap. Runs sequentially, passing a continuity preamble and rolling summary between chunks; rewritten parts are stitched.
- **Final judge** (`chunked_final_judge.py`, new): Splits original and rewritten text into N positionally-paired slices, scores each, and aggregates per-dimension by minimum. Rubric scales are embedded in the prompt with structured output. Replaces the non-windowed`LLMJudgeColumnConfig`.

### Parallel processing
- **Stateless** stages (detection, validation, augmentation, latent, judge) dispatch windows in parallel (bounded `ThreadPoolExecutor`; the per-alias rate limit still governs real in-flight calls) and merge afterward.
- **Stateful** stages (substitution-map, rewrite generation) stay sequential to thread the map / rolling summary across seams for consistency.

### Window sizing
- `detection_window_max_render_chars` (default **128 KiB**, clamped ≤ NDD's render cap) is the single knob; it is threaded into detection, augmentation, latent, substitute-map, rewrite, and judge.
- `detection_window_safety_margin_chars` (8K) leaves headroom for prompt scaffolding; `detection_window_overlap_chars` (1K) sets the overlap for the overlapping stages; a 4K floor prevents pathological shrinking.

### Fault tolerance & failure tracking
- Augmentation, latent, and the final judge are resilient to a single bad window: a window whose call fails is logged and skipped rather than dropping the whole record. Skipped-window counts are surfaced in `trace_dataframe` (`COL_AUGMENTATION_FAILED_WINDOWS` / `COL_LATENT_FAILED_WINDOWS`); the judge degrades to defaults if all windows fail.
- Detection windows re-raise (the record fails) to preserve detection completeness, and validation relies on pool failover.

### Observability
Per-window debug logging across all chunked stages: window ranges/sizes, rendered length vs cap, shrink events, rolling-summary contents, and per-stage entity/replacement/window counts.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing

## Documentation

### DIFF
--- a/src/anonymizer/config/anonymizer_config.py
+++ b/src/anonymizer/config/anonymizer_config.py
@@ -29,7 +29,7 @@ except Exception:  # pragma: no cover - fall back if NDD internals move
 # NDD's hard render cap so each window stays small enough to map/rewrite within a
 # single LLM request — large windows on entity-dense documents otherwise time out.
 # Clamped so it never exceeds NDD's cap if that is ever lowered.
-_DEFAULT_WINDOW_MAX_RENDER_CHARS = min(128 * 1024, _NDD_MAX_RENDERED_LEN)
+_DEFAULT_WINDOW_MAX_RENDER_CHARS = min(128_000, _NDD_MAX_RENDERED_LEN)
 
 
 def is_remote_input_source(value: str) -> bool:
--- a/src/anonymizer/config/anonymizer_config.py
+++ b/src/anonymizer/config/anonymizer_config.py
@@ -20,6 +20,17 @@ from anonymizer.config.rewrite import (
 
 logger = logging.getLogger(__name__)
 
+try:
+    from data_designer.engine.processing.ginja.environment import MAX_RENDERED_LEN as _NDD_MAX_RENDERED_LEN
+except Exception:  # pragma: no cover - fall back if NDD internals move
+    _NDD_MAX_RENDERED_LEN = 512_000
+
+# Default per-call render cap for the windowed long-context stages. Kept well below
+# NDD's hard render cap so each window stays small enough to map/rewrite within a
+# single LLM request — large windows on entity-dense documents otherwise time out.
+# Clamped so it never exceeds NDD's cap if that is ever lowered.
+_DEFAULT_WINDOW_MAX_RENDER_CHARS = min(128 * 1024, _NDD_MAX_RENDERED_LEN)
+
 
 def is_remote_input_source(value: str) -> bool:
     """Return True when the input source is an HTTP(S) URL."""
@@ -98,6 +109,33 @@ class Detect(BaseModel):
             "Number of characters to include before and after a chunk's entity span when "
             "building the text excerpt sent to the validator. Bounds the prompt context the "
             "validator sees per chunk; it is NOT the LLM's context window limit."
+        ),
+    )
+    detection_window_max_render_chars: int = Field(
+        default=_DEFAULT_WINDOW_MAX_RENDER_CHARS,
+        gt=0,
+        description=(
+            "Upper bound on a single rendered prompt (characters) for the windowed "
+            "augmentation, latent, substitute-map, and rewrite stages. Documents whose "
+            "rendered prompt would exceed this are processed in windows. Defaults to 128 KiB "
+            "(131072), kept below NDD's MAX_RENDERED_LEN render cap so each window maps or "
+            "rewrites within a single LLM request without timing out on long documents."
+        ),
+    )
+    detection_window_safety_margin_chars: int = Field(
+        default=8_000,
+        ge=0,
+        description=(
+            "Headroom subtracted from detection_window_max_render_chars to leave room for "
+            "prompt scaffolding and tags when sizing augmentation/latent windows."
+        ),
+    )
+    detection_window_overlap_chars: int = Field(
+        default=1_000,
+        ge=0,
+        description=(
+            "Overlap between adjacent augmentation/latent windows so an entity straddling a "
+            "window boundary is fully visible in at least one window."
         ),
     )
 
--- a/src/anonymizer/config/anonymizer_config.py
+++ b/src/anonymizer/config/anonymizer_config.py
@@ -31,6 +31,12 @@ except Exception:  # pragma: no cover - fall back if NDD internals move
 # Clamped so it never exceeds NDD's cap if that is ever lowered.
 _DEFAULT_WINDOW_MAX_RENDER_CHARS = min(128_000, _NDD_MAX_RENDERED_LEN)
 
+# Floor on the per-window character budget; mirrors ``_MIN_WINDOW_CHARS`` in the
+# engine's chunked-detection planners. Defined here (rather than imported) to keep
+# the user-facing config free of an engine import cycle. Used only to compute the
+# effective window size for overlap validation.
+_MIN_DETECTION_WINDOW_CHARS = 4_000
+
 
 def is_remote_input_source(value: str) -> bool:
     """Return True when the input source is an HTTP(S) URL."""
@@ -135,7 +141,9 @@ class Detect(BaseModel):
         ge=0,
         description=(
             "Overlap between adjacent augmentation/latent windows so an entity straddling a "
-            "window boundary is fully visible in at least one window."
+            "window boundary is fully visible in at least one window. Must be smaller than the "
+            "effective window size (max_render_chars - safety_margin_chars, floored at 4000); a "
+            "larger overlap stalls the planners to one character per step and is rejected at config time."
         ),
     )
 
@@ -151,6 +159,31 @@ class Detect(BaseModel):
         if len(deduped) != len(cleaned):
             logger.warning("entity_labels contained duplicates, removed automatically.")
         return deduped
+
+    @model_validator(mode="after")
+    def validate_window_overlap(self) -> Detect:
+        """Reject overlaps that would stall the windowed detection planners.
+
+        The augmentation/latent planners advance by ``window - overlap`` characters
+        per step. The effective window mirrors the engine's sizing:
+        ``max(_MIN_DETECTION_WINDOW_CHARS, max_render - safety_margin)``. When the
+        overlap meets or exceeds that, the stride collapses to a single character
+        and one long row explodes into tens of thousands of model calls (a 20k-char
+        row became 16,001 windows in testing), so require it to be strictly smaller.
+        """
+        effective_window = max(
+            _MIN_DETECTION_WINDOW_CHARS,
+            self.detection_window_max_render_chars - self.detection_window_safety_margin_chars,
+        )
+        if self.detection_window_overlap_chars >= effective_window:
+            raise ValueError(
+                f"detection_window_overlap_chars ({self.detection_window_overlap_chars}) must be smaller than "
+                f"the effective window size ({effective_window} chars = max({_MIN_DETECTION_WINDOW_CHARS}, "
+                "detection_window_max_render_chars - detection_window_safety_margin_chars)). A larger overlap "
+                "makes the windowed planners advance one character at a time, exploding a single row into "
+                "thousands of model calls."
+            )
+        return self
 
 
 class Rewrite(BaseModel):
--- a/src/anonymizer/engine/detection/chunked_detection.py
+++ b/src/anonymizer/engine/detection/chunked_detection.py
@@ -41,6 +41,7 @@ from __future__ import annotations
 
 import json
 import logging
+from concurrent.futures import ThreadPoolExecutor
 from typing import Any
 
 from data_designer.config import custom_column_generator
@@ -55,6 +56,12 @@ logger = logging.getLogger("anonymizer.detection.chunked_detection")
 
 # Floor on window size so a pathologically small cap still makes progress.
 _MIN_WINDOW_CHARS = 4000
+
+# Upper bound on detector windows dispatched concurrently for one record. Windows
+# are independent LLM calls, so they run in parallel; the per-alias rate limit
+# (``max_parallel_requests`` on the facade's ThrottledModelClient) still caps the
+# real in-flight count, so this just bounds thread creation on very long inputs.
+_MAX_PARALLEL_WINDOWS = 16
 
 
 class WindowedDetectionParams(BaseModel):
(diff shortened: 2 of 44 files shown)


eurekayuan:

src/anonymizer/engine/rewrite/rewrite_workflow.py
```diff
@@ -227,6 +231,8 @@ def run(
             ),
             *self._qa_wf.columns(selected_models=selected_models),
             *self._rewrite_gen_wf.columns(
+                window_max_render_chars=window_max_render_chars,
```
andreatnvidia, line 283: this only threads the user-supplied window cap into rewrite generation. domain classification, sensitivity disposition, QA generation, and final judge still build their window params from module defaults, so a user who lowers `Detect.detection_window_max_render_chars` still gets ~128k prompts in those stages. Could pass the same kwargs through those `columns()` calls and `_run_final_judge` too?

src/anonymizer/engine/rewrite/chunked_steps.py
```diff
@@ -0,0 +1,137 @@
...
+    if len(full_rendered) <= cap:
+        row[params.output_column] = merge_fn([_call(full_rendered, purpose_prefix)])
+        return row
+
+    text = str(row.get(params.text_column, ""))
+    windows = iter_boundary_windows(text, initial_window, delimiter=params.delimiter)
+    if params.first_only:
+        windows = windows[:1]
+    outputs = []
+    for start, end in windows:
```
andreatnvidia, line 108: Claude Code caught this one: once this takes the windowed path, a single transient model error or a chunk that legitimately has no meaning units can drop the whole record. Could wrap each window call, skip/log failed windows, and handle the all-failed case explicitly?

src/anonymizer/config/anonymizer_config.py
```diff
@@ -100,6 +111,33 @@ class Detect(BaseModel):
...
+    )
+    detection_window_safety_margin_chars: int = Field(
+        default=8_000,
+        ge=0,
+        description=(
+            "Headroom subtracted from detection_window_max_render_chars to leave room for "
+            "prompt scaffolding and tags when sizing augmentation/latent windows."
+        ),
+    )
+    detection_window_overlap_chars: int = Field(
```
andreatnvidia, line 139: suggestion: can we validate that `detection_window_overlap_chars` is smaller than the effective window size? Right now `overlap == window` is accepted and the planners advance one character at a time. My smoke test turned a 20k-char row into 16,001 windows.

src/anonymizer/engine/rewrite/chunked_rewrite.py
```diff
@@ -0,0 +1,291 @@
...
+            )
+            logger.debug(
+                "rewrite window %d/%d: continuity summary updated -> %d chars: %s",
+                i + 1,
+                len(windows),
+                len(summary),
+                _clip(summary),
+            )
+
+    stitched = "\n".join(part for part in rewritten_parts if part)
```
andreatnvidia, line 274: separate from the newline-stitching comment already here: filtering with `if part` also hides an empty rewrite chunk. If one window returns `{"rewritten_text": ""}`, that section disappears with no failed-window count or review signal. Maybe count/flag empty chunks instead of treating them as successful output?

andreatnvidia (commented):
Thanks for taking this on. This is a substantial first PR, and the overall direction makes sense: split long records into bounded windows, carry forward the state needed for consistency, and keep the replacement map explicit.

I left a few comments on edge cases I think are worth tightening before merge. The main themes are:

- thread the user-supplied window sizing through every windowed stage
- make per-window failures local where possible, instead of dropping the whole record
- validate overlap settings early so a bad config cannot explode into thousands of model calls
- avoid silently accepting empty rewrite chunks as successful output

The tests and docs coverage are in good shape, and I think the feature is close. These changes should make it more reliable on real long documents.

src/anonymizer/engine/rewrite/chunked_rewrite.py
```diff
@@ -0,0 +1,316 @@
...
+
+    # Chunked path: rewrite each boundary window with continuity carry-over, then stitch.
+    tagged = str(row.get(COL_TAGGED_TEXT, ""))
+    notation = TagNotation(str(row.get(COL_TAG_NOTATION) or TagNotation.sentinel.value))
+    disposition_block = row.get(COL_REWRITE_DISPOSITION_BLOCK, [])
+    replacement_map = row.get(COL_REPLACEMENT_MAP_FOR_PROMPT, {"replacements": []})
+
+    # Chunk on the already-tagged text so tag boundaries stay intact; the delimiter
+    # default ("\n") keeps cuts on line breaks.
+    windows = iter_boundary_windows(tagged, initial_window, delimiter=params.delimiter)
```
binaryaaron, line 195: > **agent** (`review-pr`): Make rewrite boundaries tag- and span-aware

`iter_boundary_windows()` can hard-cut at `max_chars` when no delimiter is available. Here it operates on `tagged_text`, so the cut can split a tag or entity value. `_filter_disposition_to_chunk()` then drops the entity’s disposition because neither fragment contains the complete value.

Please plan windows in raw-text/span coordinates and render complete tags afterward, or move intersecting boundaries to safe span edges. Add a delimiter-free test with a protected entity crossing the hard boundary.

src/anonymizer/engine/detection/chunked_detection.py
```diff
@@ -0,0 +1,309 @@
...
+    dropped only when this window does not start at the document start
+    (``window_start > 0``); likewise the right edge only when it is not the
+    document end (``window_end < text_len``). The dropped span's full form is
+    recovered from the overlapping neighbour window, where it sits interior.
+    """
+    left_artificial = window_start > 0
+    right_artificial = window_end < text_len
+    kept: list[EntitySpan] = []
+    for s in spans:
+        if left_artificial and s.start_position <= window_start:
```
binaryaaron, line 127: > **agent** (`review-pr`): Do not drop crossing detections unless recovery is proven

`drop_boundary_spans()` assumes every edge-touching span is recovered intact from an overlapping neighbor, but `iter_windows()` allows zero overlap and overlap shorter than the entity. In those cases every observation can be dropped before `resolve_overlaps()`.

Please retain boundary provenance and suppress a fragment only after confirming a complete neighboring observation, or enforce a sufficient-overlap contract. Tests should cover overlap `0`, overlap shorter than the entity, and entities touching both artificial edges.

src/anonymizer/engine/rewrite/chunked_steps.py
```diff
@@ -0,0 +1,161 @@
...
+    # the model cannot parse into the schema) should drop only that window, not
+    # the whole record. Failures are logged and skipped; the all-failed case is
+    # handled explicitly below so we never merge an empty output set silently.
+    outputs = []
+    failed = 0
+    for start, end in windows:
+        rendered = _compile_template(params.prompt_template).render(**{**row, params.text_column: text[start:end]})
+        try:
+            outputs.append(_call(rendered, f"{purpose_prefix}-{start}"))
+        except Exception:
```
binaryaaron, line 112: > **agent** (`review-pr`): Reject incomplete sensitivity coverage

A failed window is skipped whenever another window succeeds, but the merged disposition validates only that the result is nonempty. It does not require every input tagged or latent entity to appear. Missing entities are silently excluded from the rewrite instructions, while the surviving row prevents `NddAdapter` from reporting a `FailedRecord`.

Please validate the merged keys against the canonical entity set and retry, fail, or explicitly flag incomplete rows.

src/anonymizer/engine/rewrite/chunked_rewrite.py
```diff
@@ -0,0 +1,316 @@
...
+                i + 1,
+                len(windows),
+                len(summary),
+                _clip(summary),
+            )
+
+    # ``if part`` drops empty chunks from the join; ``empty_windows`` records how
+    # many so a section silently lost to an empty model response is visible as a
+    # review signal rather than passing as successful output.
+    stitched = "\n".join(part for part in rewritten_parts if part)
```
binaryaaron, line 274: > **agent** (`review-pr`): Make an empty rewrite window fail or force review

An empty window is omitted from `stitched`, while `_rewrite_empty_windows` is only counted and never affects `needs_human_review`. A document can therefore pass after losing an entire section.

Please fail the row, preserve a marked fallback section, or propagate `empty_windows > 0` into human review and the public failure surface.

src/anonymizer/engine/detection/detection_workflow.py
```diff
@@ -175,10 +189,15 @@ def _build_detection_spec(
             )
 
         columns: list[ColumnConfigT] = [
-            LLMTextColumnConfig(
+            # Windowed detection: long documents are tiled into overlapping windows
+            # so each detector call stays under the render cap. Serializable plugin
+            # config so the exported workflow stays runnable by external executors.
+            WindowedDetectionConfig(
```
binaryaaron, line 195: > **agent** (`review-pr`): Resolve the DataDesigner 0.9 plugin typing boundary

With current main’s DataDesigner and blocking ty configuration, the new `Windowed*Config` objects are rejected by the static `ColumnConfigT` union. The new model-registry generators also have incompatible `generate` and `agenerate` overrides.

Runtime plugin injection does not make that mutation visible to ty. Please introduce a typed project-owned plugin-column boundary or narrow adapter-boundary casts, and make the generator methods satisfy DataDesigner’s current protocol.

src/anonymizer/engine/replace/chunked_replace.py
```diff
@@ -0,0 +1,398 @@
...
+) -> dict[str, Any]:
+    """Build ``COL_REPLACEMENT_MAP`` for one row, chunking long documents with context carry-over."""
+    if params.alias not in models:
+        raise KeyError(
+            f"Replacement alias {params.alias!r} not present in models dict. Ensure "
+            "make_windowed_replace_generator was invoked with the same alias."
+        )
+    facade = models[params.alias]
+    cap = params.max_render_chars
+    initial_window = max(_MIN_WINDOW_CHARS, cap - params.safety_margin_chars)
```
binaryaaron, line 274: > **agent** (`review-pr`): Enforce the cap on the final submitted prompt

The planner sizes raw text before adding tags, replacement state, examples, summaries, continuity text, and response-recipe instructions. The 4,000-character floor can also exceed the configured cap by itself.

Please measure the fully rendered prompt and shrink or replan until the actual input to `facade.generate()` is within budget. Add captured-prompt assertions for every windowed stage.

src/anonymizer/config/anonymizer_config.py
```diff
@@ -100,6 +117,35 @@ class Detect(BaseModel):
             "validator sees per chunk; it is NOT the LLM's context window limit."
         ),
     )
+    detection_window_max_render_chars: int = Field(
```
binaryaaron, line 120: > **agent** (`review-pr`): Give long-context policy a shared owner

These settings are declared on `Detect` but control replace, rewrite, metadata, and judge stages. Both exported detection builders omit them, and standalone evaluation loses the originating cap.

Please introduce one cross-pipeline long-context policy, serialize it through every builder, and retain enough provenance for `evaluate()` to apply the same policy.

src/anonymizer/engine/rewrite/chunked_final_judge.py
```diff
@@ -0,0 +1,236 @@
...
+    The window count is chosen so each window's combined text fits the render
+    budget; both texts are split into that many near-equal slices. No LLM calls.
+    """
+    overhead = len(render_judge_prompt(template=template, original="", rewritten=""))
+    budget = max(_MIN_BUDGET_CHARS, cap - safety_margin_chars - overhead)
+    total = len(original) + len(rewritten)
+    n = max(1, math.ceil(total / budget))
+    o_slices = slice_evenly(original, n)
+    r_slices = slice_evenly(rewritten, n)
+    return [render_judge_prompt(template=template, original=o, rewritten=r) for o, r in zip(o_slices, r_slices)]
```
binaryaaron, line 125: > **agent** (`review-pr`): Align judge windows by content

Independent equal-count character slicing stops corresponding after the rewrite inserts, removes, or reorders content. Sensitive source text and a leaked rewritten copy can land in different judge prompts.

Please align windows using anchors or diff segments, or supply overlapping original context for each rewritten window. Add an early-deletion case with a later leaked identifier.

src/anonymizer/engine/replace/chunked_replace.py
```diff
@@ -0,0 +1,398 @@
...
+        len(windows),
+        initial_window,
+        params.delimiter,
+        params.summary_max_chars,
+    )
+    accumulated: list[dict[str, str]] = []
+    summary = ""
+    for i, (start, end) in enumerate(windows):
+        already = {(r["original"], r["label"]) for r in accumulated}
+        chunk_entities = new_chunk_entities(spans, start, end, already)
```
binaryaaron, line 312: > **agent** (`review-pr`): Assign Substitute entities only to complete-span windows

Entity assignment uses the start offset, while tag rendering requires full containment. A boundary-crossing entity is therefore listed in one prompt without its complete tagged value and omitted from the next.

Please move boundaries around spans or carry and reassemble crossing spans before assigning each entity to a prompt.

binaryaaron (changes requested):
> **agent** (`review-pr`): Long-context correctness and integration review

Reviewed head `051f41609773c22f4a17aeffb388e8c9a8645185`.

The long-context implementation has four blocking correctness problems:

1. Rewrite windows can split rendered entity tags, causing the affected entity to disappear from the protection disposition.
2. Detection drops boundary-touching spans without proving that an overlapping window contains a complete copy. Zero or short overlap can therefore erase PII.
3. Partial sensitivity-window success is accepted without validating coverage of every input entity. Those failures do not reach `FailedRecord` or human-review state.
4. Empty rewrite windows are omitted from the stitched document without failing the row or requiring review.

I also found several material defects: the configured prompt cap is not enforced on the final rendered prompt; long-context settings are lost by exported builders and standalone evaluation; final-judge windows become misaligned after length-changing rewrites; and Substitute assigns crossing entities to windows that cannot render their complete spans.

The repeated failure mode is the absence of a shared window contract. Raw, tagged, original, and rewritten offsets are treated as interchangeable, while partial-window outcomes have no typed failure channel. A shared window-plan model, explicit outcome policy, bounded executor, and final-render prompt budget would address most of these defects together.

This branch also needs a rebase before merge. Current main adds `CombinedRewriteWorkflow`, DataDesigner 0.9, and blocking ty checks. The new plugin column configurations are outside the statically accepted `ColumnConfigT`, and several generator overrides do not satisfy DataDesigner’s current signatures.

Validation completed:

- Focused tests: 196 passed
- Full test suite: 1,175 passed
- Format, lint, lockfile, SPDX, docs, and diff checks passed
- DCO remains the sole hosted check failure

NDD execution-boundary compliance, no-entity behavior, and bundled-skill compatibility look sound.

PR #189, opened 2026-06-10
feat: load bundled model providers by default

lipikaramaswamy:
## Summary
- Plain `Anonymizer()` now loads bundled `providers.yaml` and passes it to DataDesigner, so notebooks, CI, and benchmarks no longer depend on `~/.data-designer/model_providers.yaml`.
- User-supplied model configs must declare explicit `provider:` fields; init validates provider names against the resolved provider list and rejects empty provider lists with `InvalidConfigError`.
- Docs, agent skill, and tests updated to reflect bundled defaults and hardened provider/model resolution.

## Test plan
- [x] `make test` (833 passed)
- [x] `make format-check`
- [x] Verified bundled `models.yaml` provider names match `providers.yaml`
- [x] Verified `Anonymizer()` passes bundled providers to DataDesigner
- [x] Verified custom `model_providers=` override and empty-list rejection
- [x] Verified missing/unknown provider config raises `InvalidConfigError`


docs/concepts/models.md
```diff
@@ -31,7 +33,7 @@ Each pipeline stage has a **role** mapped to one of these aliases. See the full
 
 ## Custom providers
 
-Use `model_providers` to define named API endpoints for hosted models such as OpenAI or OpenRouter.
+Pass `model_providers` when you need a non-default endpoint — for example OpenAI, OpenRouter, a local GLiNER server, or an internal inference deployment. Plain `Anonymizer()` already uses bundled [build.nvidia.com](https://build.nvidia.com) settings; override only when your models point at a different provider name or URL.
```
andreatnvidia, line 36: Could we make the provider-name alignment explicit here? The Python example below currently builds only `openai` and `openrouter` providers, then calls `Anonymizer(model_providers=providers)`. With this PR, bundled model configs all reference `provider: nvidia`, so that construction raises `InvalidConfigError` unless the user also passes matching custom `model_configs`. Maybe either include the `nvidia` provider in that Python list, or make the example instantiate with `model_configs="my_models.yaml"` so the provider names line up.

src/anonymizer/interface/anonymizer.py
```diff
@@ -128,10 +134,14 @@ def __init__(
...
-        parsed = parse_model_configs(model_configs)
-        self._model_configs = parsed.model_configs
-        self._selected_models = parsed.selected_models
-        self._resolved_providers: list[ModelProvider] | None = _resolve_model_providers(model_providers)
+        try:
+            parsed = parse_model_configs(model_configs)
+            self._model_configs = parsed.model_configs
+            self._selected_models = parsed.selected_models
+            self._resolved_providers = _resolve_model_providers(model_providers)
+            validate_model_configs_reference_providers(self._model_configs, self._resolved_providers)
```
andreatnvidia: This validation also runs when the caller supplies a preconfigured `data_designer` and leaves `model_providers=None`. In that case `_resolved_providers` is the bundled `nvidia` list, so custom `model_configs` that reference providers already registered on the supplied DataDesigner now raise `InvalidConfigError` before that DataDesigner is used. Should this path skip provider-list validation, or validate against the supplied DataDesigner providers instead?

docs/concepts/models.md
lipikaramaswamy: Updated the Python example to include the `nvidia` provider alongside `openai/openrouter`, so it lines up with the YAML example above and with the bundled provider: `nvidia` model configs (no `InvalidConfigError`). I also added a short note explaining that every model config's provider name must resolve to a configured provider, so it's clear why `nvidia` needs to stay in the list.

src/anonymizer/interface/anonymizer.py
lipikaramaswamy: You're right. This was wrong for the BYO-DataDesigner path. Our resolved providers (bundled defaults when model_providers=None) are never passed to a caller-supplied DataDesigner; they only feed telemetry host classification. So I now skip validate_model_configs_reference_providers when data_designer is not None, since provider registration is owned by that instance. The default path (data_designer=None) is unchanged and still validates. The provider-field-presence check (`_validate_raw_model_configs_have_provider`) still runs unconditionally, so missing-provider configs are rejected regardless. Added `test_anonymizer_skips_provider_validation_with_supplied_data_designer` to cover it.

andreatnvidia (approved):

PR #191, opened 2026-06-16
feat(measurement): add W&B benchmark support

binaryaaron:
## Summary

- Add opt-in W&B publication, reports, workspaces, and sweep orchestration for benchmark tooling only.
- Replace loose dictionary and Pandas egress with strict measurement ingress and typed W&B config, history, summary, table, and report models.
- Add atomic completion seals and a strict, retry-safe importer for externally produced Slurm benchmark cases.
- Preserve historical v1 report reads while publishing v2 metadata with explicit native-suite, sweep-arm, and imported-config comparison axes.
- Centralize scalar measurement fields in the product package and derive W&B aggregation, validation, and report metric paths from shared immutable registries.

## Security and Privacy Boundary

- Capture measurement JSONL once through pinned descriptors with owner, regular-file, hard-link, byte, line, record, nesting, mutation, and terminal-state checks.
- Reject trace types, unknown records, coercions, non-finite values, negative counts, mixed runs, symlinks, FIFOs, sockets, devices, parent redirection, growth, and truncation before SDK import.
- Build complete outbound payloads through `extra="forbid"` models. Raw text, prompts, model responses, replacement maps, entity payloads, local paths, provider secrets, and trace data cannot enter W&B payloads.
- Isolate W&B in a guarded process environment. `WANDB_API_KEY` and ambient routing variables are removed; online runs use the local W&B credential file under the preserved home directory.
- Keep native publication best-effort so observability failures do not change benchmark or sweep outcomes. Keep sealed Slurm imports strict across init, config, log, summary, and finish.
- Keep table exposure as an explicit fail-closed allowlist while deriving operational scalar policy from package-owned field groups.

## Slurm Import Contract

- `write_completion_seal.py` atomically binds case identity, Slurm provenance, producer commit, terminal status, byte and record counts, and measurement SHA-256.
- `import_wandb_run.py` verifies the seal and snapshot before initialization, derives a destination-scoped stable run ID, and publishes imported history at deterministic step 0 with `resume="allow"`.
- Identical imports become remote no-ops. Changed sealed content receives a different run ID.
- The companion `anonymizer-experiments` launcher patch writes seals after report completion, requires a seal for `SKIP_COMPLETED`, and removes stale seals before forced reruns.

## Reports and Sweeps

- Parse historical v1 and current v2 runs into a closed `WandbRunView`.
- Escape Markdown by rendering context, reconstruct run URLs from validated identifiers, and reject mixed or ambiguous comparison groups.
- Preserve completed sweep-arm results when report or workspace creation fails, while returning a redacted view error and nonzero CLI status.
- Compose report metric families from namespace helpers and derive complete report coverage from the panel registry.

## Validation

- [x] `make test` (`1025 passed`)
- [x] Measurement tool suite (`145 passed`, council verifier)
- [x] Product measurement plus tool suites (`189 passed`; two sandbox-only integration tests excluded locally)
- [x] `make format-check`
- [x] Scoped `ty check` over changed measurement modules
- [x] `git diff --check`
- [x] Real sequential offline SDK publication and strict sealed retry
- [x] Online strict retry using local-file authentication with `WANDB_API_KEY` removed
- [x] Remote API and file readback: one history row at step 0, matching publication markers and seal digest, no configured privacy canaries
- [x] Four-member correctness, architecture, and privacy council review with no remaining actionable findings

## Notes

- W&B remains disabled by default.
- Product CLI and SDK runs do not initialize W&B.
- Workspaces and reports remain noncanonical presentation layers over sanitized benchmark data.
- The package-owned scalar field catalog is a precursor to the broader typed measurement-event migration. Closed event models, shared codecs, typed emitters, and producer migration remain deferred to separate changes.


tools/measurement/measurement_tools/wandb_setup.py
```diff
@@ -0,0 +1,506 @@
...
+    parts = {part for part in normalized.split("_") if part}
+    if "api" in parts and "key" in parts:
+        return True
+    return bool(parts & _SENSITIVE_RUN_TAG_PARTS)
+
+
+def _config_key_is_path_like(key: str) -> bool:
+    normalized = key.lower().replace("-", "_").replace(".", "_")
+    parts = {part for part in normalized.split("_") if part}
+    return any(part in {"path", "url"} or part.endswith(("_path", "_url")) for part in parts)
```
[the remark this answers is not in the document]
binaryaaron: Fixed in `11431ba`. `_config_key_is_path_like` now only checks normalized key parts against `path` and `url`, so the unreachable `endswith` branch is removed.

tools/measurement/sweep_benchmarks.py
```diff
@@ -0,0 +1,510 @@
...
+    lines = [f"Ran {result.completed_arms}/{len(result.arms)} sweep arm(s); errors={result.errored_arms}"]
+    lines.append(f"Output: {result.output_root}")
+    if result.report_url:
+        lines.append(f"Report: {result.report_url}")
+    if result.workspace_url:
+        lines.append(f"Workspace: {result.workspace_url}")
+    for arm in result.arms:
+        lines.append(
+            f"- {arm.arm_id}: {arm.status}, cases={arm.completed_cases}/{arm.completed_cases + arm.errored_cases}"
+        )
```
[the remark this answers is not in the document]
binaryaaron: Fixed in `11431ba`. Sweep arm results now carry `total_cases`, `_arm_error` populates it when possible, and rendered output uses `completed_cases/total_cases` so pre-run and partial failures no longer show as `0/0`.

tools/measurement/sweep_benchmarks.py
```diff
@@ -0,0 +1,510 @@
...
+        raise SystemExit(125) from exc
+    sys.stdout.write(render_result(result, json_output=json_output) + "\n")
+    if result.errored_arms:
+        raise SystemExit(1)
+
+
+def _run_sweep_from_cli(options: dict[str, Any]) -> SweepResult:
+    wandb_settings = _resolve_cli_wandb_settings(
+        wandb_mode=options["wandb_mode"],
+        wandb_entity=options["wandb_entity"],
```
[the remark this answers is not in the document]
binaryaaron: Fixed in `11431ba`. The CLI now constructs a `SweepCliOptions` dataclass and passes that into `_run_sweep_from_cli` instead of passing `locals()`.

tools/measurement/sweep_benchmarks.py
```diff
@@ -0,0 +1,574 @@
...
+    except Exception as exc:  # noqa: BLE001 -- keep sweeping other arms and report failure
+        return _arm_error(
+            arm,
+            suite_path=suite_path,
+            output_dir=output_dir,
+            run_name=run_name,
+            total_cases=total_cases,
+            error=str(exc),
+        )
+    return _arm_result(arm, suite_path=suite_path, output_dir=output_dir, run_name=run_name, result=result)
```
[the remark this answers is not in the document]
binaryaaron: Fixed in `d1b813f`. `materialize_arm_suite`, planned case counting, and `run_or_plan` now run inside the per-arm `try`, so materialization failures are recorded through `_arm_error` and the sweep continues.

tools/measurement/measurement_tools/wandb_setup.py
```diff
@@ -0,0 +1,506 @@
...
+    *,
+    suite_id: str,
+    output_dir: Path,
+    metadata: dict[str, Any] | None,
+) -> dict[str, Any]:
+    init_kwargs = {
+        "project": settings.effective_wandb_project,
+        "name": settings.wandb_run_name or _default_run_name(suite_id, metadata),
+        "mode": settings.wandb_mode.value,
+        "settings": wandb.Settings(console="wrap"),
```
andreatnvidia: `console="wrap"` creates an unsanitized upload channel outside the structured-payload guards. W&B captures process stdout and stderr, and the benchmark path logs full input paths, so an online run can send local paths or exception content to W&B. Please use `console="off"`, or make console capture a separate explicit opt-in, so only sanitized structured data is uploaded.

tools/measurement/measurement_tools/wandb_logging.py
```diff
@@ -0,0 +1,297 @@
...
+    dataframe = pd.DataFrame(records)
+    if "record_type" not in dataframe.columns:
+        return
+    for record_type, rows in dataframe.groupby("record_type", sort=False):
+        if record_type in PRIVACY_BLOCKED_RECORD_TYPES:
+            continue
+        if record_type not in PRIVACY_ALLOWED_RECORD_TYPES:
+            continue
+        table = normalize_table(rows)
+        table = _drop_blocked_columns(table)
```
andreatnvidia: This drops columns based only on their names, but every measurement record contains user-supplied `run_tags`, which `normalize_table` flattens. Values such as `run_tags.pipeline_id="sk-secret-token"` or `run_tags.label="Alice"` survive and are uploaded when table logging is enabled. Please drop or allowlist `run_tags.*` table columns, or apply the same value-aware sanitization before creating the table.

tools/measurement/measurement_tools/wandb_logging.py
```diff
@@ -0,0 +1,297 @@
...
+        aggregated[key] = total
+    for key, values in numeric_means.items():
+        aggregated[key] = sum(values) / len(values)
+    aggregated[f"{_SCALAR_METRIC_PREFIX}/record_count"] = len(records)
+    return aggregated
+
+
+def _metric_uses_mean_aggregation(metric_name: str) -> bool:
+    field_name = metric_name.rsplit("/", maxsplit=1)[-1]
+    return field_name.endswith(_MEAN_METRIC_FIELD_SUFFIXES)
```
andreatnvidia: The suffix heuristic misclassifies production numeric fields. `entity_precision`, `entity_recall`, `entity_f1`, and `leakage_mass` are summed across records, while `schema_version` and `timestamp_unix_sec` are also summed; the report then requests a `leakage_mass_mean` key that is never emitted. Please use an explicit per-field aggregation policy: sum counts, average bounded per-record metrics, and exclude metadata.

tools/measurement/create_wandb_report.py
```diff
@@ -0,0 +1,785 @@
...
+    sweep_param_columns: list[str] | None = None,
+) -> Any:
+    """Build an unsaved report object for a W&B run group."""
+    _wandb, wr, expr = require_wandb_report_sdk()
+    runset = wr.Runset(
+        entity=project_path.entity,
+        project=project_path.project,
+        name="Benchmark sweep",
+        filters=[expr.Metric("Group") == group],
+        groupby=["config:sweep_arm_id"],
```
andreatnvidia: The locked `wandb-workspaces==0.4.3` parser expects `section.metric` for `Runset.groupby`. With the colon form, it serializes this as a run field named `config:sweep_arm_id`; the `BarPlot` path similarly emits `config:sweep_arm_id.value`, so sweep reports do not group by the config value. Please use `config.sweep_arm_id` for the runset and `wr.Config("sweep_arm_id")` for panel grouping.

tools/measurement/create_wandb_report.py
```diff
@@ -0,0 +1,785 @@
...
+        _bar_panel(wr, "Stage Throughput", _STAGE_THROUGHPUT_METRICS, groupby=groupby),
+    ]
+
+
+def _comparison_workspace_panels(wr: Any) -> list[Any]:
+    return [
+        wr.RunComparer(diff_only=True),
+        wr.ParallelCoordinatesPlot(
+            title="Sweep Parameter Tradeoffs",
+            columns=[wr.ParallelCoordinatesPlotColumn(metric=column) for column in _WORKSPACE_COMPARISON_COLUMNS],
```
andreatnvidia: `ParallelCoordinatesPlotColumn` treats plain strings as summary metrics. Passing these already-prefixed values serializes them as `summary:config:*` and `summary:summary:*`, so every tradeoff axis points at a nonexistent field. Please build the list with typed `wr.Config(...)` and `wr.SummaryMetric(...)` accessors.

tools/measurement/sweep_benchmarks.py
```diff
@@ -0,0 +1,584 @@
...
+
+def _safe_param_name(path: str) -> str:
+    return path.replace("*", "all").replace(".", "_").replace("-", "_")
+
+
+def _rebase_suite_paths(suite: dict[str, Any], base_dir: Path) -> dict[str, Any]:
+    rebased = copy.deepcopy(suite)
+    for key in ("model_configs", "model_providers", "artifact_path"):
+        if isinstance(rebased.get(key), str):
+            rebased[key] = str(_resolve_path(rebased[key], base_dir))
```
andreatnvidia, line 504: This treats every string as a file path, but `model_configs` and `model_providers` also accept multiline inline YAML. Rebasing an inline providers block prefixes its top-level `providers:` key, so preflight rejects it; rebasing inline model configs renames `selected_models`, causing defaults to be used and custom aliases to fail validation. Preserve strings containing newlines and only rebase actual `.yaml` or `.yml` references.

tools/measurement/measurement_tools/wandb_setup.py
binaryaaron: Fixed in `8da858f`. W&B now uses `console="off"` and also disables code, Git, host, machine metadata, system statistics, and requirements capture. Regression coverage verifies the SDK settings and a real offline artifact probe confirms those values are absent.

tools/measurement/measurement_tools/wandb_logging.py
binaryaaron: Fixed in `8da858f`. Measurement tables now use per-record-type field allowlists instead of a name denylist. Suite `run_tags`, unknown columns, unsafe dimensions, and exact per-record text lengths are excluded. Table construction is also completed before the single upload call.

tools/measurement/measurement_tools/wandb_logging.py
binaryaaron: Fixed in `8da858f`. Aggregation now uses explicit sum and mean field sets. Precision, recall, F1, leakage mass, rates, and throughput use mean keys; counts use sums; schema and timestamp metadata are excluded. Report metric names and regression tests were updated together.

tools/measurement/create_wandb_report.py
binaryaaron: Fixed in `8da858f`. The report runset now uses `config.sweep_arm_id`, and grouped panels use `wr.Config("sweep_arm_id")`. The installed `wandb-workspaces==0.4.3` serialization path is covered by a regression test.

tools/measurement/create_wandb_report.py
binaryaaron: Fixed in `8da858f`. Parallel-coordinate columns now use typed `wr.Config(...)` and `wr.SummaryMetric(...)` accessors. A serialization regression test verifies the generated model references the intended config and summary fields.

tools/measurement/sweep_benchmarks.py
binaryaaron: Fixed in `8da858f`. Sweep materialization now rebases only newline-free `.yaml` or `.yml` references and preserves multiline inline model and provider YAML. Parametrized coverage checks both inline/file combinations.

tools/measurement/measurement_tools/wandb_logging.py
```diff
@@ -0,0 +1,100 @@
...
+    )
+
+
+def extract_scalar_metrics(record: MeasurementRecord) -> dict[str, float | int | bool | str]:
+    """Return aggregate metrics from one already-validated measurement record."""
+    metrics: dict[str, float | int | bool | str] = {}
+    for field_name in SCALAR_AGGREGATION_BY_FIELD:
+        value = getattr(record, field_name, None)
+        if value is not None:
+            namespace = metric_path(_SCALAR_METRIC_PREFIX, record.record_type)
```
andreatnvidia, line 88: `record_type` is too coarse for stage metrics. A split run with a 100s terminal stage, 60s detection stage, and 30s replacement stage publishes `measurement/stage/elapsed_sec=190`, triples the row counts, and retains only the last stage label. Please either include the stage name in the metric namespace or restrict generic `measurement/stage/*` metrics to `Anonymizer._run_internal`.

tools/measurement/measurement_tools/wandb_completion.py
```diff
@@ -0,0 +1,230 @@
...
+class ImportedCaseIdentity(BaseModel):
+    case_id: VisibleIdentifier
+    workload_id: VisibleIdentifier
+    config_id: VisibleIdentifier
+    repetition: NonNegativeInt
+
+    model_config = ConfigDict(extra="forbid", frozen=True, strict=True, hide_input_in_errors=True)
+
+
+class SlurmCaseProvenance(BaseModel):
```
andreatnvidia: Could this represent multiple SLURM jobs for one logical imported case, minimally as a list of `{role, job_id}` entries? A split pipeline can have separate detection and replacement jobs under one measurement file and completion seal, so a single `job_id` loses important provenance. Richer scheduler timing and resource fields can be added later.

andreatnvidia (commented):
Thanks Aaron. I re-reviewed the current head, and all six of my previous comments are addressed.

I also looked at how the sealed SLURM import would fit RAT Bench and similar distributed runs. I do not think the full integration needs to be part of this PR. The exporter and launcher wiring, stable case identities, seal lifecycle, batch import, `sacct` collection, and RAT Bench-specific metrics can evolve separately in our internal experiments repo.

I left two focused comments: one correctness issue with stage aggregation and one request to support multiple SLURM jobs per logical imported case. Richer scheduler timing and resource metadata can be added later.

The existing thread about invalid W&B identifiers escaping the post-run error guard also still looks actionable.

tools/measurement/sweep_benchmarks.py
```diff
@@ -0,0 +1,666 @@
...
+    try:
+        result = create_benchmark_group_report(
+            project,
+            settings=wandb_settings,
+            group=wandb_settings.wandb_group or spec.sweep_id,
+            expected_run_kind="sweep_arm",
+        )
+    except Exception as exc:  # noqa: BLE001 -- remote SDK errors must not expose response contents
+        raise WandbViewCreationError(f"W&B report creation failed ({type(exc).__name__})") from None
+    return result.report_url
```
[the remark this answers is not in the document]
binaryaaron: Fixed in `9465b37`. `WandbProjectPath` construction now occurs inside the existing redaction boundary for both report and workspace creation. Completed arm results are preserved, and validation failures expose only the exception type. Added regression coverage for both paths.

tools/measurement/measurement_tools/wandb_logging.py
binaryaaron: Fixed in `9465b37`. Generic stage scalars now include only the terminal `Anonymizer._run_internal` record, while the sanitized stage table retains all workflow stages. The regression test verifies terminal elapsed time and row counts directly.

tools/measurement/measurement_tools/wandb_completion.py
binaryaaron: Addressed in `9465b37`. Completion seals now accept repeatable typed `--slurm-job role=job_id` entries, enforce bounded identifiers and unique roles, and carry the job list through strict import, W&B metadata policies, and report projection. Existing singular Slurm fields and old seal serialization remain compatible.

andreatnvidia (approved):

src/anonymizer/measurement/recorders.py
```diff
@@ -179,7 +179,7 @@ def _record_model_workflow(
         "column_count": column_count,
         "column_names": column_names or [],
         "model_usage": dict(model_usage or {}),
-        **dict(extra_fields or {}),
+        "local_fields": dict(extra_fields or {}),
```
andreatnvidia: `record_ndd_workflow()` reaches this shared builder with `record_type="ndd_workflow"`, but the builder always emits `"local_fields": {}`. `NddWorkflowMeasurement` rejects unknown fields and does not define `local_fields`, so native NDD measurements fail strict ingress.

Could we emit `local_fields` only for `model_workflow`, or explicitly accept and exclude it for NDD? An NDD recorder-to-ingress round-trip test would cover this.

tools/measurement/import_wandb_run.py
```diff
@@ -0,0 +1,305 @@
...
+        result = WandbPublisher().publish_payload(
+            settings,
+            payload=prepared.payload,
+            measurement_sha256=prepared.measurement_sha256,
+            record_count=prepared.record_count,
+        )
+    except Exception as exc:
+        logger.error("Strict W&B import failed (%s)", type(exc).__name__)
+        raise SystemExit(1) from None
+    output = result.model_dump(mode="json", exclude_none=True)
```
andreatnvidia, line 299: Could the import result include the W&B run URL, or at least its entity and project alongside the run ID? Currently callers must reconstruct the URL.

It would also help to report whether publication created a run, resumed one, or detected an already-complete idempotent import.

tools/measurement/README.md
```diff
@@ -240,6 +262,238 @@ record, such as source refs, commit SHAs, CI pipeline IDs, topology labels, or
...
+`ANONYMIZER_MEASUREMENT_WANDB_JOB_TYPE`,
+`ANONYMIZER_MEASUREMENT_WANDB_RUN_NAME`,
+`ANONYMIZER_MEASUREMENT_WANDB_TAGS`, and
+`ANONYMIZER_MEASUREMENT_WANDB_LOG_TABLES`. Generic W&B routing variables such
+as `WANDB_MODE`, `WANDB_PROJECT`, `WANDB_GROUP`, `WANDB_JOB_TYPE`, `WANDB_NAME`,
+and `WANDB_TAGS` are deliberately ignored. Precedence is an explicit CLI value,
+then the corresponding `ANONYMIZER_MEASUREMENT_WANDB_*` variable, then the
+publisher default. The W&B SDK still receives its audited timeout variables,
+including `WANDB_HTTP_TIMEOUT` and `WANDB_INIT_TIMEOUT`. Authentication comes
+from the SDK's local credential files under the preserved home directory; the
```
andreatnvidia, line 303: Since the publisher deliberately removes `WANDB_API_KEY`, online publication requires credentials in the preserved home directory. Could we document a short authentication and connectivity preflight for external launchers, so credential failures are detected before benchmark execution?

andreatnvidia:
Validated the sealed external-import flow end to end with a multi-stage SLURM case. The importer accepted seven strict measurement records, preserved multiple role-tagged SLURM job IDs, and an identical re-import was idempotent. Destination-scoped re-import also produced a distinct completed run as intended.

The only correctness issue found was the native NDD `local_fields` mismatch described inline.

src/anonymizer/measurement/recorders.py
binaryaaron: > **agent** (`review-pr`): Fixed native NDD strict ingress

Fixed in `28c145a`. `record_ndd_workflow()` no longer emits `local_fields`; `model_workflow` records retain the excluded local-only field. Added a real collector-to-strict-ingress round-trip test for native NDD records.

tools/measurement/import_wandb_run.py
binaryaaron: > **agent** (`review-pr`): Added importer destination and publication state

Fixed in `28c145a`. `WandbPublishResult` now includes entity, project, a validated run URL, and a typed `created`, `resumed`, or `already_complete` publication state. JSON and plain importer output expose the destination and state, with regression coverage for all three lifecycle outcomes.

tools/measurement/README.md
binaryaaron: > **agent** (`review-pr`): Documented credential-safe launcher preflight

Fixed in `28c145a`. The Slurm import documentation now shows public-cloud and self-hosted `wandb login --verify` preflights using the preserved local credential store with `WANDB_API_KEY` unset. It also warns against logging `wandb status` and notes that publisher connectivity cannot depend on ambient proxy or custom-CA variables.

andreatnvidia:
This seems to work well overall. I validated the sealed import flow with three 100-record benchmark runs, including multi-job SLURM provenance and W&B publication.

The main follow-up I noticed is that some of the generic metrics are difficult to interpret. For example, per-record mean F1 can use a different denominator for each workload and differ substantially from micro F1.

I think we could address this in a focused follow-up PR without blocking this one, unless you would prefer to include it now. Some things I'm thinking of:

- Publish strict and relaxed micro precision/recall/F1 with their denominators.
- Separate pipeline completion from record failures and coverage.
- Distinguish unavailable telemetry from measured zero.
- Reduce duplicate or low-signal summary fields.
- Leave benchmark-specific dashboards and headline metrics to the experiment tooling.

The underlying typed records already preserve the required counts, so this should be additive and should not require rerunning benchmarks.

tools/measurement/import_wandb_run.py
binaryaaron: Implemented in 28c145a. The import result now includes the W&B run URL and publication state (created, resumed, or already completed); the non-JSON CLI output reports both. This avoids requiring callers to reconstruct the destination URL.

tools/measurement/README.md
binaryaaron: Implemented in 28c145a. The measurement README now documents an external-launcher authentication and connectivity preflight using `wandb login --verify`, including the self-hosted endpoint form. It also explains that the publisher uses credentials from the preserved W&B home directory and removes `WANDB_API_KEY` from its process environment.

tools/measurement/import_wandb_run.py
binaryaaron: > **agent** (`review-pr`): Closed the default-entity destination gap

Follow-up in `feb3226`. Strict online imports now require `--wandb-entity` (or `ANONYMIZER_MEASUREMENT_WANDB_ENTITY`), so every successful online import result includes the configured entity, project, and W&B run URL. Offline imports may omit the entity. Added a regression test for the rejected ambiguous online configuration.

tools/measurement/measurement_tools/wandb_setup.py
```diff
@@ -0,0 +1,523 @@
...
+    flags = os.O_RDONLY | getattr(os, "O_CLOEXEC", 0) | getattr(os, "O_DIRECTORY", 0)
+    no_follow = getattr(os, "O_NOFOLLOW", 0)
+    try:
+        descriptor = os.open(absolute.anchor, flags)
+        for part in absolute.parts[1:]:
+            child = os.open(part, flags | no_follow, dir_fd=descriptor)
+            os.close(descriptor)
+            descriptor = child
+            _validate_directory_metadata(os.fstat(descriptor), final=part == absolute.name)
+        return descriptor
```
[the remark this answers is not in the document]
binaryaaron: > **agent** (`review-pr`): Fixed final-directory detection

Fixed in `5404abc`. `_open_directory_no_follow` now determines the final component by its index, matching the completion-seal traversal. The regression test simulates a foreign-owned intermediate directory whose basename repeats at the final output directory and verifies staging still succeeds.

andreatnvidia (commented):
Thanks Aaron. I re-reviewed the current head and resolved my previous threads. The fixes for strict NDD ingress, importer destination/publication state, and credential preflight look good. The integration remains opt-in and CI is green.

One issue before approval: `test_wandb_staging_allows_repeated_basename_in_foreign_intermediate_directory` resolves descriptors through `/proc/self/fd`, which is unavailable on macOS and causes `make test` to fail. Could we make this portable by comparing `st_dev` and `st_ino` with `temp_root.stat()`?

Non-blocking: the PR is still quite large. Reports, workspaces, and sweeps look like the cleanest candidates for a follow-up if reducing the current diff is still practical. The metric-semantics improvements discussed earlier can also remain a focused follow-up.

andreatnvidia (approved):

PR #195, opened 2026-06-26
feat: add entity coverage as the new Nemo Anonymizer feature

### LINKED ISSUE #193: Replace detection_validity score with leakage_judge in NA replace
### Priority Level

Medium

### Task Summary

As a follow up on [this thread](https://nvidia.slack.com/archives/C096DGNMAGN/p1782167491161449?thread_ts=1782142704.740929&cid=C096DGNMAGN), detection validity score mainly includes precision complaints rather than making sure any real PII slip through after entity detection. The latter question will be answered by judge_leakage, which is briefly explored in [here](https://gitlab-master.nvidia.com/sdg-research/anonymizer-experiments/-/blob/lramaswamy/latency-benchmark/experiments/latency-benchmark/scripts/leakage_eval.py?ref_type=heads).

This issue is related to #176 as we have enough evidence to replace the `detection_validity` score with `leakage_judge`.

_No response_

### Dependencies

_No response_

memadi-nv:
### Entity Coverage Judge

Adds a per-record, judge-anchored recall metric to `Anonymizer.evaluate()` for both Replace and Rewrite modes.

Entity coverage runs in two phases:

- **LLM candidate extraction** — independently extracts all in-scope PII and sensitive values from the original text using structured output. The prompt respects the configured `entity_labels` and uses `data_summary` as optional interpretive context.
- **Deterministic postprocessing** — filters out-of-scope and non-literal candidates, deduplicates them by normalized value, and compares them with the anonymizer’s `final_entities` using value-based exact, subspan, and composite matching.

**Formula:**

`entity_coverage = n_covered / n_candidates`

Only judge candidates enter the denominator. Anonymizer detections do not independently increase it, but a detection matching a judge candidate classifies that candidate as covered. Labels remain diagnostic context; Detection Validity separately evaluates value-label correctness.

**Output columns:**

- `entity_coverage` (`float | None`) — fraction of unique judge candidate values detected by the anonymizer. Returns `1.0` when no candidates were missed or the judge found no candidates, and `None` when judging was unavailable.
- `missed_entities` (`list`) — unmatched candidate values with their judge-assigned label and reasoning.

The metric measures detection recall independently of `strict_entity_protection`.

### Evaluation Behavior

- Entity Coverage runs for all Replace and Rewrite evaluations.
- Substitute additionally runs Type Fidelity, Attribute Fidelity, and Relational Consistency.
- Rewrite additionally runs the holistic Privacy, Quality, and Style judge.
- Detection Validity is opt-in through:

```python
EvaluateConfig(compute_detection_validity=True)
```

- Model-alias validation checks only the judge roles active for the requested evaluation.
- Rewrite evaluation validates its holistic judge without requiring the complete rewrite-generation model set.
- Active Replace judges are submitted through a shared DataDesigner workflow.

### Model Configuration

- Adds the `entity_coverage_judge` evaluation model role.
- Configures `nemotron-super` as the default entity-coverage judge.
- Adds model-role validation to evaluation and benchmark preflight checks.

### Results and Display

- Persists `entity_labels`, `data_summary`, and rewrite metadata on preview and run results for later evaluation.
- Includes `entity_coverage` and `missed_entities` in public evaluated results.
- Displays the covered/candidate fraction and an expandable table containing each missed value, label, and reason.
- Supports Entity Coverage display for both Replace and Rewrite records.
- Keeps Detection Validity columns absent unless the diagnostic is enabled.

### Reliability and Logging

- Preserves row order and successful scores when individual LLM records fail or are dropped.
- Marks affected scores as unavailable instead of removing or misaligning records.
- Retains evaluation failures in `AnonymizerResult.failed_records`.
- Adds INFO logs for evaluation stages, timing, and completion.
- Adds DEBUG logs for active configuration and failure details.
- Adds WARNING logs identifying unavailable active-judge scores.

### Documentation and Skill

- Documents the Entity Coverage formula, candidate extraction, deterministic matching, output columns, and special values.
- Documents Detection Validity as an opt-in diagnostic.
- Updates Replace and Rewrite evaluation guidance.
- Updates model configuration documentation, contributor guidance, and the bundled Anonymizer skill.

### Tests

Adds coverage for:

- Prompt construction and structured candidate extraction
- Entity scope and `data_summary` propagation
- Literal filtering, value deduplication, and matching
- Judge-anchored recall calculation
- Replace and Rewrite evaluation orchestration
- Optional Detection Validity
- Model-role validation
- Display rendering
- Evaluation logging and unavailable scores
- Row-order preservation and partial workflow failures

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update
- [ ] Refactoring

## Validation

- [x] Lint and format checks
- [x] Tests on Python 3.11, 3.12, and 3.13
- [x] Advisory type check
- [x] Documentation build

## Related Issues

Closes #193


docs/concepts/evaluation.md
```diff
@@ -55,7 +89,7 @@ Four LLM judges run per record: one that scores detection quality and three that
 
 > "Are the detected entities actually correct (value, label) pairs in context?"
 
-This judge runs regardless of which replace mode was used. It looks at each detected span and flags:
+Detection validity is **opt-in** — it runs only when you pass `Anonymizer.evaluate(..., config=EvaluateConfig(compute_detection_validity=True))`. It is disabled by default and is intended for internal model/threshold experiments rather than as a customer-facing metric. When enabled, it looks at each detected span and flags:
```
lipikaramaswamy:
```suggestion
Detection validity is disabled by default. To enable it, set `compute_detection_validity=True` in `EvaluateConfig`. This metric is intended for internal model and threshold evaluation. It examines each detected span and flags:
```

docs/concepts/evaluation.md
```diff
@@ -45,7 +45,41 @@ with open("/tmp/preview.pkl", "rb") as f:
 evaluated = anonymizer.evaluate(loaded)
 ```
 
-Four LLM judges run per record: one that scores detection quality and three that score replacement quality (Substitute mode only).
+Which judges run depends on the strategy and your `EvaluateConfig`:
```
lipikaramaswamy:
```suggestion
The active judges depend on the anonymization strategy and `EvaluateConfig`:
```

docs/concepts/evaluation.md
```diff
@@ -45,7 +45,41 @@ with open("/tmp/preview.pkl", "rb") as f:
...
+
+| Judge | When it runs |
+|-------|--------------|
+| **Entity Coverage** | Always. |
+| **Detection Validity** | Only when `EvaluateConfig(compute_detection_validity=True)` — off by default. |
+| **Type Fidelity, Attribute Fidelity, Relational Consistency** | Substitute mode only. |
```
lipikaramaswamy:
```suggestion
| Evaluation | Applies to |
|------------|------------|
| Entity Coverage | All anonymization strategies |
| Detection Validity | Optional; enable with EvaluateConfig(compute_detection_validity=True) |
| Type Fidelity, Attribute Fidelity, Relational Consistency** | Substitute mode only |
```

src/anonymizer/config/default_model_configs/models.yaml
```diff
@@ -28,3 +28,17 @@ model_configs:
       temperature: 0.4
       top_p: 1.0
       timeout: 300
+
+  - alias: nemotron-super
+    model: nvidia/nemotron-3-super-v3
+    provider: nvidia
+    inference_parameters:
+      max_parallel_requests: 16
+      max_tokens: 16384
```
lipikaramaswamy, line 37: Why did we select this value for max_tokens?

src/anonymizer/config/anonymizer_config.py
```diff
@@ -212,5 +212,11 @@ class EvaluateConfig(BaseModel):
     knobs are introduced.
     """
 
-    # Intentionally empty for now. New fields land here as evaluation
-    # configurability is introduced.
+    compute_detection_validity: bool = False
+    """Run the tag-precision judge (detection_valid / detection_invalid_entities).
+
+    Disabled by default — intended for internal use during model and threshold
+    experiments, not a customer-facing metric. When True, adds
```
lipikaramaswamy: Suggest not referencing "customers", see my suggested change [here](https://github.com/NVIDIA-NeMo/Anonymizer/pull/195/changes#r3669171660)

src/anonymizer/engine/evaluation/entity_coverage_judge.py
```diff
@@ -0,0 +1,621 @@
...
+logger = logging.getLogger("anonymizer.evaluation.entity_coverage_judge")
+
+_FINAL_ENTITIES_FOR_COVERAGE_COL = "_final_entities_for_coverage_judge"
+_WORD_RE = re.compile(r"\w+", re.UNICODE)
+# Grammatical stopwords only — function words that carry no PII and just absorb
+# article/preposition noise in a value (e.g. "the Nawabganj" matches "Nawabganj").
+# Deliberately does NOT include generic content descriptors (festival, summit, club,
+# conference, …): those turn a named event/org into a quasi-identifier, so ignoring
+# them would suppress real leaks (e.g. "Davos Summit" collapsing into "Davos").
+_COVERAGE_IGNORE_TOKENS = frozenset(
```
lipikaramaswamy: I understand the intent is to remove only grammatical noise, but can we safely assume these tokens never carry meaning within an entity value? For example, they are significant in names such as The Hague, Bank of America, and AT&T—and AT&T would reduce to the single token t. Could this normalization be limited to specific patterns, such as an optional leading article, instead of removing these tokens from arbitrary positions?

src/anonymizer/engine/evaluation/entity_coverage_judge.py
```diff
@@ -0,0 +1,621 @@
...
+
+    @classmethod
+    def _extract_invalid(cls, parsed: BaseModel) -> list[dict[str, object]]:
+        return [e.model_dump() for e in parsed.leaked_entities]
+
+    # ----------------------------------------------------------------- overrides
+
+    def column_config(self, selected_models: EvaluateModelSelection) -> LLMTextColumnConfig:
+        """Override to inject instance-specific entity_labels and strict_entity_protection."""
+        return LLMTextColumnConfig(
```
lipikaramaswamy: What is the reason for using LLMTextColumnConfig instead of LLMStructuredColumnConfig with output_format=EntityCoverageSchema? The current approach appears to require custom JSON extraction and means the model does not receive the schema’s field descriptions.

src/anonymizer/engine/evaluation/entity_coverage_judge.py
```diff
@@ -0,0 +1,621 @@
...
+
+In structured text, distinguish a generic category name from a literal category value. A short \
+literal that itself instantiates an in-scope type remains a candidate when it appears in a form \
+row, list item, or signature. Do not merge tokens from unrelated people or fields into one value.
+</reportability_check>
+
+<guidance>
+Do NOT flag:
+- Text whose in-scope semantic type is not supported by its original-text context.
+- Information that is inferable but not literally present in the text.
```
lipikaramaswamy, line 165: How is strict mode expected to evaluate inferable or indirect information when the prompt explicitly excludes information that is not literally present? Should strict mode change this instruction, or is it intended to cover only literal quasi-identifiers?

src/anonymizer/engine/evaluation/entity_coverage_judge.py
```diff
@@ -0,0 +1,621 @@
...
+
+    # Exact match against a single final value.
+    if any(leaked_tokens == final_tokens for final_tokens in final_token_lists):
+        return True
+
+    # Subspan: the leak's core tokens appear as a contiguous, in-order run within a single
+    # final entity (adjacency + order required, not merely a shared set of tokens).
+    leaked_core = _core_token_sequence(leaked_tokens)
+    if leaked_core and any(
+        _is_contiguous_sublist(leaked_core, _core_token_sequence(final_tokens)) for final_tokens in final_token_lists
```
lipikaramaswamy: Could value-only subspan matching suppress a genuine leak when the same token appears in a different entity or occurrence? For example, a missed surname "Green" would be considered covered by a detected street value "Bowling Green Road". Should matching consider compatible labels or source spans before removing the candidate?

src/anonymizer/engine/evaluation/entity_coverage_judge.py
```diff
@@ -0,0 +1,621 @@
...
+        leaked_lists: list[list[dict]] = []
+
+        for idx in out.index:
+            raw = out[self.RAW_COL].loc[idx] if self.RAW_COL in out.columns else None
+            leaked = _parse_leaked_entities(raw)
+            if leaked is None:
+                coverage_vals.append(None)
+                leaked_lists.append([])
+            else:
+                leaked = _filter_nonliteral_entities(leaked, out[COL_TEXT].loc[idx])
```
lipikaramaswamy: Should deterministic post processing also verify that each returned label belongs to the configured `entity_labels`? The prompt asks the model to remain within scope, but an out-of-scope or empty label currently appears to be accepted and would lower the coverage score.

src/anonymizer/engine/evaluation/entity_coverage_judge.py
```diff
@@ -0,0 +1,621 @@
...
+                leaked = _deduplicate_judge_entities(leaked)
+                final_entities = out[_FINAL_ENTITIES_FOR_COVERAGE_COL].loc[idx]
+                leaked = _filter_covered_leaked_entities(leaked, final_entities)
+                n_final = len(final_entities) if isinstance(final_entities, list) else 0
```
lipikaramaswamy: What is the intended unit for the coverage calculation: unique values, source occurrences, or `(value, label)` pairs? `entities_by_value` groups occurrences by surface value and accumulates their labels, after which `_final_entities_for_coverage()` expands each label into a separate entry. As a result, `n_final` currently counts `(value, label)` pairs. Is that the denominator we want the public metric to represent?

src/anonymizer/engine/evaluation/entity_coverage_judge.py
```diff
@@ -0,0 +1,621 @@
...
+    strict_guidance = (
+        "- In strict protection mode, flag every missed entity regardless of its perceived risk level."
+        if strict_entity_protection
+        else ""
+    )
+
+    active_labels = entity_labels if entity_labels is not None else DEFAULT_ENTITY_LABELS
+    labels_str = ", ".join(active_labels)
+
+    prompt = f"""You are an exhaustive privacy-entity span extractor. Extract every literal value \
```
lipikaramaswamy: Could we consolidate some of the repeated instructions in this prompt? For example, the requirements that values be literal, remain within `entity_labels`, and be scanned exhaustively each appear in several sections. Would one authoritative statement for each rule make the prompt shorter and reduce the chance of inconsistent wording?

src/anonymizer/interface/display.py
```diff
@@ -576,6 +582,74 @@ def _verdict_badge(valid: object, correct: int, total: int) -> tuple[str, str]:
...
+    Replace mode shows a Satisfied / Not Satisfied badge.
+    """
+    if COL_ENTITY_COVERAGE not in row.index:
+        return ""
+    coverage = row.get(COL_ENTITY_COVERAGE)
+    leaked_entries = _normalize_invalid_entities(row.get(COL_LEAKED_ENTITIES))
+    n_leaked = len(leaked_entries)
+    n_detected = _count_detected_entity_label_pairs(row)
+    total = n_detected + n_leaked
+
```
lipikaramaswamy, line 600: Could this check use `pd.isna(coverage)` rather than only `coverage is None`? In replace mode, a coverage column containing floats and unavailable values is represented by Pandas as `float64`, so `None` becomes `NaN`. The current code then reaches `round(float(coverage) * 100)` and raises `ValueError: cannot convert float NaN to integer` when displaying an unavailable row.

src/anonymizer/engine/evaluation/entity_coverage_judge.py
```diff
@@ -0,0 +1,621 @@
...
+            if self.INVALID_COL not in out.columns:
+                out[self.INVALID_COL] = [[] for _ in range(len(out))]
+            else:
+                out[self.INVALID_COL] = out[self.INVALID_COL].apply(
+                    lambda value: value if isinstance(value, list) else []
+                )
+            if not had_record_ids:
+                out = out.drop(columns=[RECORD_ID_COLUMN])
+            return out, result.failed_records
+        except Exception:
```
lipikaramaswamy: Could we include the exception reason in an INFO or WARNING log so it is visible without enabling debug logging?

src/anonymizer/config/default_model_configs/models.yaml
memadi-nv: Just updated the model name to match the one in build not inference. Regarding the `max_token`, this is the set value for nemotron-super in build. Should I bump this down for any limitation that I was not aware of?

src/anonymizer/engine/evaluation/entity_coverage_judge.py
memadi-nv: Good catch. Just restricted the normalization to leading articles only (a, an, the) and dropped all prepositions from the set — so "Bank of America" and "AT&T" are no longer at risk of being incorrectly normalized. 

The worst case with just leading articles is minor: a value like "The Home Depo" loses its leading "The", but that's a much smaller tradeoff than suppressing real leaks IMO.

src/anonymizer/engine/evaluation/entity_coverage_judge.py
memadi-nv: Did some A/B testing and did not observe much regression after updating the prompt. Removed the repetitive parts in the prompt.

src/anonymizer/engine/evaluation/entity_coverage_judge.py
memadi-nv: Strict mode is intended to cover only literal quasi-identifiers. Updating the instructions to be more specific.

src/anonymizer/engine/evaluation/entity_coverage_judge.py
memadi-nv: This is a known limitation documented in the function's docstring. Label-aware matching would be the natural fix, but judge labels are free-form and don't always share the same vocabulary as Anonymizer's detection labels — so comparing them directly would introduce false negatives of a different kind (mismatched labels on the same entity value). 

The tradeoff is accepted because the failure mode is bounded: it only affects single-token leaks where that token coincidentally appears inside a longer detected entity of a different type which seems like a very rare case. Also, I'm hoping on the anonymizer detector to catch the single token entity in the first place considering the context- with entity augmentor.

That said, happy to revisit if you feel the risk of suppressing real leaks outweighs the label incompatibility tradeoff.

src/anonymizer/engine/evaluation/entity_coverage_judge.py
memadi-nv: You're right! Not sure why `LLMTextColumnConfig` was added in first place.

Switched to LLMStructuredColumnConfig with output_format=EntityCoverageSchema — the model now receives the schema's field descriptions and DataDesigner handles structured output enforcement. Removed the custom JSON string fallback parser (_parse_json_object) since it's no longer reachable, and updated the tests accordingly. Verified with nemotron-super — no parsing issues observed.

src/anonymizer/engine/evaluation/entity_coverage_judge.py
memadi-nv: Addressed, Thanks! — added a deterministic label filter in postprocessing as a safety net. As apposed to the discussion in [R377](https://github.com/NVIDIA-NeMo/Anonymizer/pull/195/changes#r3678178491), label drift is unlikely in practice since the prompt explicitly instructs the model to use the configured entity_labels vocabulary, and `LLMStructuredColumnConfig` reinforces this via the schema field descriptions.

src/anonymizer/engine/evaluation/entity_coverage_judge.py
memadi-nv: The intended unit is (value, label) pairs — this keeps the numerator and denominator consistent since the judge also returns leaked entities as (value, label) pairs. An entity detected under multiple labels contributes one entry per label, which is intentional: each detection deserves credit in the score. Added a note to the docstring to make this explicit.

src/anonymizer/engine/evaluation/entity_coverage_judge.py
```diff
@@ -0,0 +1,619 @@
...
+
+
+class LeakedEntity(BaseModel):
+    value: str = Field(description="Exact text span from the original text that was missed by the anonymizer.")
+    label: str = Field(description="Entity type label (e.g. first_name, email, phone_number).")
+    reasoning: str = Field(description="One sentence explaining why this is PII that the anonymizer missed.")
+
+
+class EntityCoverageSchema(BaseModel):
+    leaked_entities: list[LeakedEntity] = Field(
```
lipikaramaswamy: Should `leaked_entities` be required rather than using `default_factory=list`? As written, `{}` is valid according to the generated schema, and `_parse_leaked_entities({})` returns an empty list. So an incomplete structured response can produce perfect coverage instead of an unavailable score. Removing the default and adding a regression test for `{}` would distinguish “the judge found no leaks” from “the judge omitted its result.”

docs/concepts/evaluation.md
```diff
@@ -45,7 +45,41 @@ with open("/tmp/preview.pkl", "rb") as f:
...
+|------------|------------|
+| Entity Coverage | All anonymization strategies |
+| Detection Validity | Optional; enable with EvaluateConfig(compute_detection_validity=True) |
+| Type Fidelity, Attribute Fidelity, Relational Consistency** | Substitute mode only |
+
+---
+
+### Entity Coverage
+
+> "Which sensitive values from the original text survived into the anonymized output?"
```
lipikaramaswamy: The documented meaning still does not match the implementation. This workflow identifies sensitive values in the original text that are absent from final_entities; it never receives the anonymized output. Therefore, it measures detection coverage rather than whether sensitive values survived anonymization. Could we either describe and name it consistently as a detection-recall metric?

docs/concepts/evaluation.md
```diff
@@ -45,7 +45,41 @@ with open("/tmp/preview.pkl", "rb") as f:
...
+
+Entity coverage is the primary residual-leakage metric and **always runs**. The judge scans the original text independently and reports any in-scope sensitive values that were **not** removed or replaced. A deterministic postprocessing step then drops candidates that are already accounted for by the Anonymizer's final entities (exact, sub-span, or composite matches), so only genuine leaks remain.
+
+The judge is scoped and contextualized by the same signals used during anonymization:
+
+- **`entity_labels`** — the detection taxonomy in scope; the judge only reports values whose type falls within it.
+- **`data_summary`** — used purely to interpret literal values and their semantic types, never to invent entities absent from the text.
+- **`strict_entity_protection`** — (rewrite only) when enabled, the judge also reports inferable/indirect sensitive values, not just literal identifiers.
```
lipikaramaswamy: Strict mode is now explicitly limited to literal quasi-identifiers in the prompt, but the documentation still says that it reports inferable or indirect sensitive values. Just thinking, is that what we want?

src/anonymizer/interface/anonymizer.py
lipikaramaswamy, line 440: Should evaluation model validation account for `compute_detection_validity`? Detection validity is disabled by default, but `check_evaluate=True` currently validates its model alias unconditionally. A configuration containing only the models needed for the enabled judges can therefore fail before evaluation begins.

docs/concepts/evaluation.md
memadi-nv: Updated the doc to detecting literal quasi-identifiers matching the entity coverage prompt. 

> Just thinking, is that what we want?

Yes, that's intentional. Strict mode in the evaluator is deliberately scoped to literal quasi-identifiers to mirror what strict_entity_protection does in the Anonymizer rewrite . Inferable/indirect values are the domain of `privacy_goal`, not `strict_entity_protection`, so including them in the evaluator's strict mode would create a mismatch between anonymizer and evaluator entity coverage scores."

docs/concepts/evaluation.md
```diff
@@ -45,7 +45,43 @@ with open("/tmp/preview.pkl", "rb") as f:
...
+Entity coverage is the primary detection-recall metric and **always runs**. The judge scans the original text independently and reports any in-scope sensitive values that were **not detected** by the anonymizer. A deterministic postprocessing step then drops candidates that are already accounted for by the Anonymizer's final entities (exact, sub-span, or composite matches), so only genuine misses remain.
+
+Note: the judge measures detection coverage, not output leakage. A value that was detected but replaced with a poor substitute would still score as covered.
+
+The judge is scoped and contextualized by the same signals used during anonymization:
+
+- **`entity_labels`** — the detection taxonomy in scope; the judge only reports values whose type falls within it.
+- **`data_summary`** — used purely to interpret literal values and their semantic types, never to invent entities absent from the text.
+- **`strict_entity_protection`** — (rewrite only) when enabled, the judge lowers the threshold for borderline literal spans — flagging quasi-identifiers that would normally receive benefit of the doubt.
+
```
asteier2026, line 70: I don't understand why strict_entity_protection should impact anything. What does it mean "no benefit-of-the-doubt for missed quasi-identifiers)"

src/anonymizer/engine/evaluation/entity_coverage_judge.py
```diff
@@ -0,0 +1,618 @@
...
+        "Flag ALL entities as leaked if they were not caught — including quasi-identifiers\n"
+        "and low-risk entities that would normally be given benefit of the doubt.\n"
+        "Do NOT apply MINIMUM NECESSARY CHANGE reasoning to excuse a missed entity.\n"
+        "Do NOT excuse a missed entity because its combined re-identification risk is low.\n"
+        "Any PII span not caught by the anonymizer is a miss in strict mode.\n"
+        "Strict mode lowers the threshold for borderline literal spans — it does NOT change\n"
+        "the requirement that every flagged value must be literally present in the original text.\n"
+        "</strict_entity_protection>"
+    )
+
```
asteier2026: Are we generally trying to both do the job of our LLM augment step as well as the contextual reasoning in the sensitivity disposition? I feel like that might be too much for one prompt. Thus, I'm wondering if worrying about strict_entity_protection is really worth it. It could backfire and give us lame, low risk entities. Where do we state the "MINIMUM NECESSARY CHANGE reasoning"? And are we comparing what entities this prompt gathers to the entities we gather from gliner and LLM augment ("final_entities") or are we taking the sensitivity disposition into account which might tag some as "leave_as_is"? I think worrying about the latter might be too much.  Just to be clear "strict_entity_protection" has no impact on the initial gliner/llm augment entity detection, it only impacts the sensitivty disposition

src/anonymizer/engine/evaluation/entity_coverage_judge.py
lipikaramaswamy: Thanks, the `(value, label)` unit makes sense. 

One follow-up: should every final detection receive credit if the metric represents detection recall? `n_final` currently includes false-positive detections. For example, if the coverage judge finds one genuinely missed entity and the anonymizer produced 99 false positives, the score would be `99 / 100 = 0.99`, even though none of those detections represents a sensitive entity that was successfully detected.

Would it be more accurate to calculate coverage from the judge’s independent candidate set, i.e. candidates matched by final detections divided by all candidates identified by the judge?

src/anonymizer/engine/evaluation/entity_coverage_judge.py
```diff
@@ -0,0 +1,618 @@
...
+
+class LeakedEntity(BaseModel):
+    value: str = Field(description="Exact text span from the original text that was missed by the anonymizer.")
+    label: str = Field(description="Entity type label (e.g. first_name, email, phone_number).")
+    reasoning: str = Field(description="One sentence explaining why this is PII that the anonymizer missed.")
+
+
+class EntityCoverageSchema(BaseModel):
+    leaked_entities: list[LeakedEntity] = Field(
```
lipikaramaswamy: Could we align this schema description with the independent-extraction prompt? The prompt tells the model to return every in-scope candidate without knowing what the anonymizer detected, while this description asks it to return only entities “the anonymizer failed to detect” and to return an empty list when “the anonymizer caught everything.”

Because `LLMStructuredColumnConfig` includes these field descriptions in the model’s structured-output instructions, the model receives contradictory guidance and cannot determine what the anonymizer caught. Should the raw schema describe independently extracted candidates, leaving the missed-versus-covered classification to deterministic postprocessing?

src/anonymizer/engine/constants.py
```diff
@@ -60,11 +60,16 @@
...
-# Replace evaluation: detection-validity judge
+# Replace evaluation: detection-validity judge (opt-in internal feature; default off)
 COL_DETECTION_JUDGE = "_detection_judge"  # raw judge output, internal
 COL_DETECTION_VALID = "detection_valid"  # user-facing bool (None if judge unavailable)
 COL_DETECTION_INVALID_ENTITIES = "detection_invalid_entities"  # user-facing list of {value, label, reasoning}
 
+# Replace / rewrite evaluation: entity coverage judge (customer-facing recall metric)
+COL_ENTITY_COVERAGE_JUDGE = "_entity_coverage_judge"  # raw judge output, internal
+COL_ENTITY_COVERAGE = "entity_coverage"  # user-facing float | None (0.0–1.0)
+COL_LEAKED_ENTITIES = "leaked_entities"  # user-facing list of {value, label, reasoning}
```
lipikaramaswamy: Now that the metric is explicitly defined as detection coverage rather than output leakage, should this public column be named `missed_entities` or `undetected_entities` instead?

`leaked_entities` suggests that these values were confirmed to survive in the anonymized output, but this workflow only compares independently identified values with the detector’s final entities and never examines the output. Since this is a new public column, should we align the name with what the metric actually establishes before the API is released?

tools/measurement/run_benchmarks.py
```diff
@@ -371,6 +371,7 @@ def _preflight_config_errors(spec: BenchmarkSpec, *, parsed_models: Any | None)
                 or anonymizer_config.rewrite is not None,
                 check_rewrite=anonymizer_config.rewrite is not None,
                 check_evaluate=config.evaluate,
+                check_detection_validity_judge=config.evaluate,
```
lipikaramaswamy: Does `config.evaluate` imply that detection validity is enabled here? Benchmark execution calls `anonymizer.evaluate(result)` without an `EvaluateConfig`, so `compute_detection_validity` remains false.

Passing `check_detection_validity_judge=config.evaluate` therefore requires a model alias for a judge that will not run, which appears to reintroduce the conditional-validation issue fixed in `Anonymizer.evaluate()`. Should this remain false unless `ConfigSpec` gains a separate option for enabling detection validity?

src/anonymizer/engine/constants.py
memadi-nv: Renamed `leaked_entities` → `missed_entities` for the user-facing output column. Also cleaned up the internal schema: the LLM raw output now uses `candidate_entities` (independent extractions, pre-postprocessing) and `missed_entities` is reserved for the postprocessed, confirmed misses.

src/anonymizer/engine/evaluation/entity_coverage_judge.py
memadi-nv: You're right that `strict_entity_protection` only affects sensitivity disposition, not GLiNER/LLM detection — and the judge compares against `final_entities` (detection output only), not disposition decisions. Entities tagged `leave_as_is` still count as covered since detection succeeded.

The judge's strict mode is intentionally minimal — it only lowers the threshold for borderline literal quasi-identifiers, **not replicating disposition logic**. That said, if you feel it adds more noise than signal, we can remove it from the judge entirely since the detection recall metric stands on its own without it.

docs/concepts/evaluation.md
memadi-nv: In the judge context, that translated to 'flag borderline quasi-identifiers even if they seem low-risk.' But as mentioned [in here](https://github.com/NVIDIA-NeMo/Anonymizer/pull/195#discussion_r3716776860), since wemight be removing it from the judge for this PR, this becomes moot — the coverage metric works purely off detection output regardless.

src/anonymizer/engine/evaluation/entity_coverage_judge.py
memadi-nv: Good point — but the intent of entity coverage is specifically to score the anonymizer's detection recall: did the detector catch the PII that was actually in the text? n_final representing the detector's output is intentional. False positives in n_final are a detection quality issue, but that's what detection validity is for. The two metrics are designed to be complementary (I know that detection validity is disabled now but the goal is to detect FP) — coverage tells you about recall, detection validity tells you about precision. Switching to the judge-anchored formula would conflate the two. This is an interesting discussion and I think is worth more inputs.

src/anonymizer/engine/evaluation/entity_coverage_judge.py
asteier2026: Thanks Marjan, I think my preference would be to remove it.

src/anonymizer/engine/evaluation/entity_coverage_judge.py
memadi-nv: Update: more discussed with the rest of the team, and agreed on entity coverage as judge-anchored recall.

src/anonymizer/engine/evaluation/entity_coverage_judge.py
memadi-nv: discussed with the team and agreed on switching to the judge-anchored formula.

memadi-nv:
Recent code changes:

- **Redefined entity coverage as judge-anchored recall:** The evaluation judge independently extracts candidate entities, and coverage is calculated as the fraction of judge candidates matched by Anonymizer’s final detections.
- **Removed rewrite disposition policy from entity coverage:** `strict_entity_protection` no longer affects the coverage prompt or score; it remains limited to Rewrite sensitivity disposition.Cleaned up the coverage prompt interface: Simplified the prompt function after removing the strict-protection argument.
**Aligned scoring with value-based matching:** Coverage now counts unique normalized candidate values, preventing labels from double-weighting a value; label correctness remains part of detection validity.

asteier2026 (approved):
Looks great!

src/anonymizer/engine/evaluation/entity_coverage_judge.py
```diff
@@ -0,0 +1,599 @@
...
+    entity_scope_block = _entity_type_scope_block(entity_labels)
+    data_context_section = f"\n\n{_data_summary_block(data_summary)}" if data_summary and data_summary.strip() else ""
+
+    entity_scope_guidance = (
+        "- Respect the entity_type_scope: do not report candidate values outside the configured scope."
+        if entity_labels is not None
+        else ""
+    )
+
+    active_labels = entity_labels if entity_labels is not None else DEFAULT_ENTITY_LABELS
```
lipikaramaswamy, line 105: How should the coverage judge interpret `entity_labels=None`? In this path, `<entity_type_scope>` says to evaluate all PII and sensitive types, but `active_labels` substitutes `DEFAULT_ENTITY_LABELS`, after which `<identifier_taxonomy>` says those listed types are the types in scope. The label instructions also say to return labels exactly as they appear in `entity_type_scope`, although that block contains no labels in this case.

Detection treats `entity_labels=None` as permissive: the augmenter may produce labels outside `DEFAULT_ENTITY_LABELS`, and those labels are retained in `final_entities`. Could these conflicting instructions cause the coverage judge to omit domain-specific candidates on the default path? Should the default labels be presented as non-exhaustive examples when `entity_labels` is `None`, with exact-label enforcement only when an explicit list was configured? This can get messy and go off rails really quickly though, so I'm ok flagging this as a follow up, but then we should be clear about which entities we are focusing on in this judge.

src/anonymizer/interface/anonymizer.py
```diff
@@ -426,29 +447,169 @@ def evaluate(
...
                 check_evaluate=True,
+                check_rewrite_judge=is_rewrite,
+                check_detection_validity_judge=evaluate_config.compute_detection_validity,
             )
         except ValueError as exc:
             raise InvalidConfigError(str(exc)) from exc
-        text_column = output.resolved_text_column
+
+        entity_labels = getattr(output, "entity_labels", None)
+        strict_entity_protection = getattr(output, "strict_entity_protection", False)
```
lipikaramaswamy: Now that `strict_entity_protection` no longer affects entity coverage, is this value still needed during evaluation? It appears to be read here but never used.

More broadly, the PR added `strict_entity_protection` to `AnonymizerResult` and `PreviewResult` to preserve it for the coverage judge. If there is no longer another consumer, could we remove that result-field plumbing and its tests rather than introduce an unused public result attribute?

docs/concepts/evaluation.md
```diff
@@ -45,7 +45,42 @@ with open("/tmp/preview.pkl", "rb") as f:
...
 ```
 
-Four LLM judges run per record: one that scores detection quality and three that score replacement quality (Substitute mode only).
+The active judges depend on the anonymization strategy and `EvaluateConfig`:
+
+| Evaluation | Applies to |
+|------------|------------|
+| Entity Coverage | All anonymization strategies |
+| Detection Validity | Optional; enable with EvaluateConfig(compute_detection_validity=True) |
+| Type Fidelity, Attribute Fidelity, Relational Consistency** | Substitute mode only |
```
lipikaramaswamy:
```suggestion
| Type Fidelity, Attribute Fidelity, Relational Consistency | Substitute mode only |
```

src/anonymizer/config/default_model_configs/models.yaml
lipikaramaswamy: It's a model limitation, usually we set higher for entity rich text. ok to leave as is!

lipikaramaswamy (approved):
A few minor nits, and one follow up worth thinking through in `active_labels`.

src/anonymizer/engine/evaluation/entity_coverage_judge.py
memadi-nv: The current `None` path contains conflicting prompt instructions, and I agree the prompt and postprocessing need one consistent scope.

I’m hesitant to make the default path non-exhaustive because open-ended label generation can introduce false positives and make the entity-coverage denominator less stable across runs. My current preference is for `entity_labels=None` to resolve to `DEFAULT_ENTITY_LABELS`, with domain-specific types supplied explicitly.

The augmenter is currently permissive when labels are None, so this decision affects more than the coverage prompt. I’d like to test both behaviors and align detection, the coverage prompt, and postprocessing in a follow-up PR. Explicitly configured labels remain strictly enforced in this PR.

PR #197, opened 2026-06-27
fix: validate augmented entities before finalization

fallintoplace:
## Summary
- keep the existing seed validation step used to prepare augmentation inputs
- add a merged validation step for the combined seed and augmented entities before finalization
- add regression coverage for merged validation decisions across chunked validation, custom columns, and the detection workflow

## Why
The detection workflow only validated seed detector entities before augmentation. After the merged entity set was built, augmented-only entities had no matching validation decisions, and finalization kept unmatched entities unchanged. That allowed hallucinated augmented entities to become final detections without validation.

## Impact
Augmented entities now need a validation decision from the merged candidate pass before they can reach the final detected entity set.

## Validation
- `uv run --group dev ruff check src/anonymizer/engine/constants.py src/anonymizer/engine/detection/chunked_validation.py src/anonymizer/engine/detection/custom_columns.py src/anonymizer/engine/detection/detection_workflow.py src/anonymizer/engine/workflow_columns/detection/config.py src/anonymizer/engine/workflow_columns/detection/impl.py tests/engine/test_chunked_validation.py tests/engine/test_detection_custom_columns.py tests/engine/test_detection_workflow.py`
- `uv run --group dev pytest tests/engine/test_chunked_validation.py tests/engine/test_detection_custom_columns.py tests/engine/test_detection_workflow.py -q`

### DIFF
--- a/src/anonymizer/engine/constants.py
+++ b/src/anonymizer/engine/constants.py
@@ -38,6 +38,8 @@ COL_VALIDATION_CANDIDATES = "_validation_candidates"
 COL_VALIDATION_SKELETON = "_validation_skeleton"
 COL_VALIDATION_DECISIONS = "_validation_decisions"
 COL_VALIDATED_ENTITIES = "_validated_entities"
+COL_MERGED_VALIDATION_DECISIONS = "_merged_validation_decisions"
+COL_MERGED_VALIDATED_ENTITIES = "_merged_validated_entities"
 
 # Step 6: apply_validation_and_finalize
 COL_DETECTED_ENTITIES = "_detected_entities"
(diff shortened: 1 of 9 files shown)


fallintoplace:

lipikaramaswamy:
Hi @fallintoplace, thanks for the PR and for describing the failure mode.

Because this adds another LLM validation pass to the detection workflow, we would need evidence before taking on the added cost and latency. Could you please open an issue with the datasets, models/configs, and examples where augmented entities are hallucinated or otherwise incorrectly finalized?

I’m going to close this PR for now under our [contributing guidelines](https://github.com/NVIDIA-NeMo/Anonymizer/blob/main/CONTRIBUTING.md), which ask external contributors to start non-trivial changes from a linked, triaged issue so we can confirm scope before implementation. We can study and discuss the hallucination cases there.

PR #186, opened 2026-06-09
feat: rewrite evaluation improvements 

memadi-nv:
# PR Summary — Rewrite Evaluation

## Overview

Separates rewrite evaluation from `run()` / `preview()` and exposes it as an explicit `Anonymizer.evaluate()` call, consistent with replace evaluation. Adds detection validity scoring to rewrite evaluation, replaces numeric judge scores with categorical rubrics, and renames the fluency dimension.

---

## Core Changes

- **`Anonymizer.evaluate()` now supports rewrite results** — `AnonymizerResult` and `PreviewResult` carry `rewrite_config` so `evaluate()` can dispatch without the user restating the mode. `evaluate()` on a rewrite result validates only the evaluate model aliases it actually uses (`check_rewrite=False`), not the full rewrite pipeline roles.

- **Final judge moved out of `run()` / `preview()`** — the repair loop (leakage, utility, `needs_human_review`) still runs as part of every `run()` call. The holistic quality judge runs only when `evaluate()` is called explicitly.

- **Detection validity added to rewrite evaluation** — `detection_valid` is a `float | None` in rewrite mode, representing the fraction of detected entities that passed the judge. Passthrough rows (no detected entities) receive `1.0` (trivially valid) with an `INFO` log. If the score cannot be computed, `detection_valid` is `None`, logged as a `WARNING
`, and displayed as **Unavailable**.

- **`rewrite_judge` model role moved to evaluate config** — no longer part of rewrite generation config; now under `selected_models.evaluate.rewrite_judge`, consistent with the other judge roles.

- **Categorical judge scores replace numeric `1–10` rubrics** — the final rewrite judge now returns `low` / `medium` / `high` per dimension, stored as structured JSON under `judge_evaluation`. Higher is better for all three dimensions, rendered as colour-coded badges in `display_record()`.

- **`naturalness` renamed to `style`** — more precisely reflects what the rubric measures (fluency, grammar, readability).

- **`judge_evaluation` column made public** — renamed from `_judge_evaluation`; appears in `display_record()` output and user-facing docs.

- **`_render_scores_section` receives `is_rewrite` as an explicit parameter** — replaces the previous column-name scan that could misfire on user-defined column names.

---
## Display

- `display_record()` for rewrite results renders `detection_valid` alongside utility and leakage scores.
- `detection_valid = None` / `NaN` renders as **Unavailable** in grey rather than a numeric score.
- Categorical judge rubrics (privacy, quality, style) render with `high` / `medium` / `low` badges.

[...]

PR #199, opened 2026-06-29
fix: evaluate rewrite_judge not checked

memadi-nv:
## Summary

- **Bug:** `evaluate.rewrite_judge` alias was never validated — `run()` skips `check_evaluate` entirely, and `evaluate()` passes `check_rewrite=False` but `rewrite_judge` was nested under `if check_rewrite`, so neither path caught a misconfigured alias at preflight
- **Fix:** Added `check_rewrite_judge: bool = False` to `validate_model_alias_references`; gates `evaluate.rewrite_judge` validation independently of `check_rewrite`, so full rewrite pipeline roles are not required just to call `evaluate()`
- **Call site:** Rewrite `evaluate()` path in `anonymizer.py` now passes `check_rewrite_judge=True` alongside `check_rewrite=False, check_evaluate=True`
- **Tests:** Two new tests in `test_model_loader.py` — bad alias raises with `check_rewrite_judge=True`; skipped without it

Follow-up to #186.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing

## Documentation

### DIFF
--- a/src/anonymizer/engine/ndd/model_loader.py
+++ b/src/anonymizer/engine/ndd/model_loader.py
@@ -260,13 +260,16 @@ def validate_model_alias_references(
     check_substitute: bool = False,
     check_rewrite: bool = False,
     check_evaluate: bool = False,
+    check_rewrite_judge: bool = False,
 ) -> None:
     """Validate that active workflow model aliases exist in the model pool.
 
-    ``check_evaluate`` validates the evaluation roles. ``detection_validity_judge``
-    is always checked when evaluation is on; the three ``replace_*_judge`` roles
-    are additionally checked when ``check_substitute`` is also True (those only
-    fire on the Substitute replace strategy).
+    ``check_evaluate`` validates the shared evaluation roles (``detection_validity_judge``
+    is always checked when on; the three ``replace_*_judge`` roles are additionally
+    checked when ``check_substitute`` is also True). ``check_rewrite_judge`` validates
+    ``evaluate.rewrite_judge`` and must be set independently when evaluating a rewrite
+    result — it is separate from ``check_rewrite`` so that full rewrite pipeline roles
+    are not required just to call ``evaluate()`` on an existing rewrite result.
     """
     known_aliases = {model_config.alias for model_config in model_configs}
     detection_roles = selected_models.detection.model_dump()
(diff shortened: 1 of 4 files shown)


src/anonymizer/engine/ndd/model_loader.py
```diff
@@ -284,15 +287,15 @@ def validate_model_alias_references(
     if check_evaluate:
         evaluate_roles = selected_models.evaluate.model_dump()
         _collect_role(roles_to_check, "evaluate.detection_validity_judge", evaluate_roles["detection_validity_judge"])
+        if check_rewrite_judge:
+            _collect_role(roles_to_check, "evaluate.rewrite_judge", evaluate_roles["rewrite_judge"])
```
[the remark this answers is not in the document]
memadi-nv: Currently `check_rewrite_judge=True` also sets `check_evaluate=True`. Since this is not part of the public API, I believe we can just leave it as is.

src/anonymizer/interface/anonymizer.py
```diff
@@ -383,6 +383,7 @@ def evaluate(
                     self._selected_models,
                     check_rewrite=False,
                     check_evaluate=True,
+                    check_rewrite_judge=True,
```
lipikaramaswamy, line 386: Could we add one interface-level regression test for this call path? The current tests cover `validate_model_alias_references(..., check_rewrite_judge=True)`, but the important behavior is that rewrite `Anonymizer.evaluate()` actually enables that validation. A test that either asserts the call includes `check_rewrite_judge=True`, or behaviorally verifies that a bad `evaluate.rewrite_judge` alias raises `InvalidConfigError`, would protect the integration path.

lipikaramaswamy (commented):
LGTM. The fix is narrow - rewrite evaluation now validates `evaluate.rewrite_judge` without requiring the full rewrite pipeline model roles 👍

lipikaramaswamy (approved):

src/anonymizer/interface/anonymizer.py
memadi-nv: Thanks for flagging this. Addressed.

PR #205, opened 2026-07-02
fix: increase latent entity rationale maxLength from 150 to 250

asteier2026:
## Summary
- Increases `LatentEntitySchema.rationale` `max_length` from 150 to 250 characters

## Motivation
Validation errors on longer LLM-generated rationale strings were causing record loss. The model regularly produces rationales in the 150–200 char range that were being silently dropped.


lipikaramaswamy (approved):
Reasonable change, esp for larger models.

PR #206, opened 2026-07-02
feat: replace MeaningUnitAspect enum with open-ended str

asteier2026:
## Summary
- Drops the `MeaningUnitAspect` enum from `EntityDispositionSchema` and `schemas/__init__.py`
- Changes `MeaningUnitSchema.aspect` from `MeaningUnitAspect` to `str`
- Updates tests to use string literals instead of enum values

## Motivation
The enum was causing record loss: the model returns open-ended aspect values like `"goal"`, `"symptom"`, `"diagnosis"` that can't be exhaustively enumerated. Validation errors were silently dropping records.


src/anonymizer/engine/schemas/rewrite.py
```diff
@@ -249,33 +249,14 @@ class StrictSensitivityDispositionSchema(SensitivityDispositionSchema):
...
-
 class MeaningUnitImportance(str, Enum):
     critical = "critical"
     important = "important"
 
 
 class MeaningUnitSchema(BaseModel):
     id: int = Field(ge=1)
-    aspect: MeaningUnitAspect
+    aspect: str
```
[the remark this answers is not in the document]
asteier2026: fixed

tests/engine/test_qa_generation.py
```diff
@@ -75,13 +74,13 @@
     units=[
         MeaningUnitSchema(
             id=1,
-            aspect=MeaningUnitAspect.ROLE,
+            aspect="role",
```
lipikaramaswamy, line 77: nit: could we add a regression test using an aspect outside the former enum, such as "diagnosis" or "goal"? The current "role" and "environment" fixtures were valid under the old enum, so they don’t directly verify the behavior this PR fixes.

lipikaramaswamy (approved):
LGTM on the code and tests (left a small nit), plus internal benchmark results support merging.

tests/engine/test_qa_generation.py
asteier2026: Update made

PR #207, opened 2026-07-02
feat: enhance augment prompt to detect disguised identifiers

asteier2026:
## Summary
- Adds an `Additional extraction requirements` block to the LLM entity augmentation prompt
- Requires verbatim span extraction (no normalization, reconstruction, or reformatting)
- Adds detection examples for identifiers written as digit words (phone/SSN/credit card) and letter-by-letter name spellings

## Motivation
The augmenter was missing entities that were obfuscated or spoken aloud, e.g.:
- `"nine o two, five five five, one two three four"` (phone number)
- `"J-O-H-N"` (name spelled out with hyphens)

### DIFF
--- a/src/anonymizer/engine/detection/detection_workflow.py
+++ b/src/anonymizer/engine/detection/detection_workflow.py
@@ -670,6 +670,24 @@ Other information:
 
 <<EXAMPLE_BLOCK>>
 
+Additional extraction requirements:
+- The "value" field must be the EXACT verbatim span from the input text.
+  Copy the text character-for-character exactly as it appears.
+  Do NOT normalize, correct spelling, expand abbreviations, decode encodings,
+  infer hidden values, translate text, reformat numbers, or otherwise modify
+  the extracted span.
+- Extract only text that is explicitly present in the input.
+  Never reconstruct, guess, or generate a value that does not appear verbatim.
+- Identifiers may be disguised, fragmented, hyphenated, misspelled, obfuscated,
+  spaced out, mixed with punctuation, or written in words instead of digits.
+  Detect the identifier and extract the exact text as written.
+  Examples of disguised identifiers to detect:
+  - Phone numbers, SSNs, and credit card numbers spoken as digit words,
+    including "o" or "oh" used in place of zero:
+    "nine o two, five five five, one two three four"
+  - Names spelled out letter by letter with hyphens or commas:
+    "J-O-H-N", "M, A, R, Y"
+
 ---
 Input text: <<TAGGED_TEXT>>
 Already-detected entities: <<SEED_ENTITIES>>
--- a/src/anonymizer/engine/detection/detection_workflow.py
+++ b/src/anonymizer/engine/detection/detection_workflow.py
@@ -652,7 +652,7 @@ Output: {"entities": [{"value": "Jane", "label": "first_name", "reason": "first 
         "  Examples of disguised identifiers to detect:\n"
         '  - Any identifier spoken as digit words, including "o" or "oh" used in place of zero:\n'
         '    "nine o two, five five five, one two three four"\n'
-        '  - Values spelled out letter by letter with hyphens or commas:\n'
+        "  - Values spelled out letter by letter with hyphens or commas:\n"
         '    "J-O-H-N", "M, A, R, Y"\n'
     )
 
--- a/src/anonymizer/engine/detection/detection_workflow.py
+++ b/src/anonymizer/engine/detection/detection_workflow.py
@@ -645,20 +645,6 @@ Input text: Jane Doe lives in <<SENSITIVE:city>>Santa Clara<</SENSITIVE:city>>. 
 Already-detected entities: [{"value": "Santa Clara", "label": "city"}]
 Output: {"entities": [{"value": "Jane", "label": "first_name", "reason": "first name"}, {"value": "Doe", "label": "last_name", "reason": "last name"}, {"value": "full-time", "label": "employment_status", "reason": "employment status"}]}"""
 
-    example_block += """
-
-Example (disguised identifier — digit words):
-Input text: Reach me at nine o two, five five five, one two three four.
-Already-detected entities: []
-Output: {"entities": [{"value": "nine o two, five five five, one two three four", "label": "phone_number", "reason": "phone number written as digit words with 'o' for zero"}]}"""
-
-    example_block += """
-
-Example (disguised identifier — letter spelling):
-Input text: My name is J-O-H-N D-O-E.
-Already-detected entities: []
-Output: {"entities": [{"value": "J-O-H-N", "label": "first_name", "reason": "first name spelled out letter by letter"}, {"value": "D-O-E", "label": "last_name", "reason": "last name spelled out letter by letter"}]}"""
-
     disguised_hints_block = (
         "- Identifiers may be disguised, fragmented, hyphenated, misspelled, obfuscated,\n"
         "  spaced out, mixed with punctuation, or written in words instead of digits.\n"
--- a/src/anonymizer/engine/detection/detection_workflow.py
+++ b/src/anonymizer/engine/detection/detection_workflow.py
@@ -604,7 +604,19 @@ Template: <<VALIDATION_SKELETON>>
     )
 
 
+_NUMBER_DISGUISE_LABELS: frozenset[str] = frozenset(
+    {"phone_number", "ssn", "social_security_number", "credit_card", "credit_card_number"}
+)
+_NAME_DISGUISE_LABELS: frozenset[str] = frozenset(
+    {"first_name", "last_name", "name", "full_name", "person_name"}
+)
+
+
 def _get_augment_prompt(*, data_summary: str | None, labels: list[str], strict_labels: bool = False) -> str:
+    label_set = set(labels)
+    include_number_hint = not strict_labels or bool(label_set & _NUMBER_DISGUISE_LABELS)
+    include_name_hint = not strict_labels or bool(label_set & _NAME_DISGUISE_LABELS)
+
     if strict_labels:
         label_block = (
             "Here are the allowed entity classes. Use ONLY labels from this list:\n"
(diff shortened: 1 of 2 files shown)


src/anonymizer/engine/detection/detection_workflow.py
```diff
@@ -670,6 +670,24 @@ def _get_augment_prompt(*, data_summary: str | None, labels: list[str], strict_l
...
+  Never reconstruct, guess, or generate a value that does not appear verbatim.
+- Identifiers may be disguised, fragmented, hyphenated, misspelled, obfuscated,
+  spaced out, mixed with punctuation, or written in words instead of digits.
+  Detect the identifier and extract the exact text as written.
+  Examples of disguised identifiers to detect:
+  - Phone numbers, SSNs, and credit card numbers spoken as digit words,
+    including "o" or "oh" used in place of zero:
+    "nine o two, five five five, one two three four"
+  - Names spelled out letter by letter with hyphens or commas:
+    "J-O-H-N", "M, A, R, Y"
```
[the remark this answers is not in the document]
asteier2026: I think it's good the way it is.

src/anonymizer/engine/detection/detection_workflow.py
```diff
@@ -670,6 +670,24 @@ def _get_augment_prompt(*, data_summary: str | None, labels: list[str], strict_l
...
+
```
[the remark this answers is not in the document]
asteier2026: I think it's good the way it is.

src/anonymizer/engine/detection/detection_workflow.py
```diff
@@ -670,6 +670,24 @@ def _get_augment_prompt(*, data_summary: str | None, labels: list[str], strict_l
 
 <<EXAMPLE_BLOCK>>
 
+Additional extraction requirements:
```
lipikaramaswamy, line 682: This block conflicts with strict custom-label mode. For example, when `entity_labels=["hostname", "ipv4"]`, the prompt first says to skip entities outside the allowed list, but this unconditional block then instructs the model to detect phone numbers, SSNs, credit cards, and names.

Can we generate this guidance conditionally? So we include each disguised-identifier example only when its canonical label is allowed, or when the default permissive label set is being used. A strict hostname/IP prompt should contain none of the phone, SSN, card, or name instructions.

src/anonymizer/engine/detection/detection_workflow.py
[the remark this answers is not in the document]
lipikaramaswamy: I agree that a complete worked example would make the new behavior less ambiguous, particularly by demonstrating the canonical label and confirming that the output value must preserve the exact disguised span.

Could we move the exact-span requirements before the demonstrations and add at least one labeled input/output example for a digit-word identifier?

```
Input: `Call me at nine o two, five five five, one two three four.`

Output:
```json
{
  "entities": [
    {
      "value": "nine o two, five five five, one two three four",
      "label": "phone_number",
      "reason": "phone number written as digit words"
    }
  ]
}
```

src/anonymizer/engine/detection/detection_workflow.py
asteier2026: Added change to make conditional.

src/anonymizer/engine/detection/detection_workflow.py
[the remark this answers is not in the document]
asteier2026: Added changes to 
- Requirements moved above <<EXAMPLE_BLOCK>> — the LLM reads verbatim-extraction rules before seeing the worked examples that demonstrate them
- Disguised-identifier hints are now conditional: in strict mode, the number hint (phone/SSN/CC) only appears when one of those labels is in scope; the name hint only appears when a name label is in scope; in permissive mode both always appear
- Worked examples added conditionally to example_block — digit-word phone example when include_number_hint, letter-spelling name example when include_name_hint — both showing exact verbatim output, the correct label, and a reason field

src/anonymizer/engine/detection/detection_workflow.py
```diff
@@ -645,7 +657,47 @@ def _get_augment_prompt(*, data_summary: str | None, labels: list[str], strict_l
...
+Already-detected entities: []
+Output: {"entities": [{"value": "nine o two, five five five, one two three four", "label": "phone_number", "reason": "phone number written as digit words with 'o' for zero"}]}"""
+
+    if include_name_hint:
+        example_block += """
+
+Example (disguised identifier — letter spelling):
+Input text: My name is J-O-H-N D-O-E.
+Already-detected entities: []
+Output: {"entities": [{"value": "J-O-H-N", "label": "first_name", "reason": "first name spelled out letter by letter"}, {"value": "D-O-E", "label": "last_name", "reason": "last name spelled out letter by letter"}]}"""
```
lipikaramaswamy: Thanks for making these hints conditional. I noticed one remaining edge case in strict mode: the number family can be enabled by `ssn` or a credit-card label, but the worked example always outputs `phone_number`. Similarly, the name family may be enabled by `name` or `full_name`, while the example outputs `first_name` and `last_name`.

Could we choose the worked example and output label from the labels that are actually allowed? It would also be helpful to include `credit_debit_card`, which is the canonical card label in `DEFAULT_ENTITY_LABELS`.

A small parameterized test checking that strict-mode example outputs only use allowed labels would make this especially robust.

asteier2026: I'm having second thoughts about making the hints conditional. It would be impossible to list all the entities where tricks like this could occur and just what the trick is.  So, up for discussion, but for now I took the conditional/label-gating out — disguised-identifier hints and examples now appear unconditionally in every augment prompt, with wording updated to say "any identifier" rather than calling out specific label types. One worry - is the prompt too obviously geared towards RAT-Bench? We have to show the prompt in the paper we're writing. But I definitely needed these couple examples to get diff2 to work well.

binaryaaron: > **agent** (`review-pr`): Keep universal guidance, but select label-safe examples

I agree that the general obfuscation guidance should be unconditional. A narrow allowlist cannot enumerate every way an in-scope identifier might be fragmented, punctuated, spelled out, or otherwise disguised.

The remaining issue is with the labeled few-shot outputs. In strict mode, the prompt can say to use only `hostname` and `ipv4`, then demonstrate outputs labeled `phone_number`, `first_name`, and `last_name`. The existing Jane/Doe/city example has the same problem.

We could keep the universal guidance everywhere while selecting each worked example only when every label it emits belongs to the active strict scope. A small private helper should be enough:

```python
@dataclass(frozen=True)
class _AugmentExample:
    text: str
    output_labels: frozenset[str]


_DISGUISED_NUMBER_EXAMPLE = _AugmentExample(
    text="""\
Example (disguised identifier: digit words):
Input text: Reach me at nine o two, five five five, one two three four.
Already-detected entities: []
Output: {"entities": [{"value": "nine o two, five five five, one two three four", "label": "phone_number", "reason": "phone number written as digit words"}]}""",
    output_labels=frozenset({"phone_number"}),
)

_DISGUISED_NAME_EXAMPLE = _AugmentExample(
    text="""\
Example (disguised identifier: letter spelling):
Input text: My name is J-O-H-N D-O-E.
Already-detected entities: []
Output: {"entities": [{"value": "J-O-H-N", "label": "first_name", "reason": "first name spelled out letter by letter"}, {"value": "D-O-E", "label": "last_name", "reason": "last name spelled out letter by letter"}]}""",
    output_labels=frozenset({"first_name", "last_name"}),
)


def _select_augment_examples(
    *,
    labels: list[str],
    strict_labels: bool,
) -> tuple[_AugmentExample, ...]:
    examples = (
        _DISGUISED_NUMBER_EXAMPLE,
        _DISGUISED_NAME_EXAMPLE,
    )
    if not strict_labels:
        return examples

    allowed_labels = frozenset(labels)
    return tuple(
        example
        for example in examples
        if example.output_labels <= allowed_labels
    )


def _render_augment_examples(
    *,
    labels: list[str],
    strict_labels: bool,
) -> str:
    examples = _select_augment_examples(
        labels=labels,
        strict_labels=strict_labels,
    )
    return "\n\n".join(example.text for example in examples)
```

The same selection rule should cover the existing base example so that every strict-mode demonstration is label-safe. The generic verbatim and obfuscation rules can remain in a separate unconditional block.

The rendered blocks could then use the existing prompt substitution convention:

```python
example_block = _render_augment_examples(
    labels=labels,
    strict_labels=strict_labels,
)

prompt = """\
...
<<DISGUISED_GUIDANCE_BLOCK>>
<<EXAMPLE_BLOCK>>
...
"""

return substitute_placeholders(
    prompt,
    {
        ...
        "<<DISGUISED_GUIDANCE_BLOCK>>": disguised_guidance_block,
        "<<EXAMPLE_BLOCK>>": example_block,
    },
)
```

A parameterized test can enforce the contract directly:

```python
@pytest.mark.parametrize(
    "labels",
    [
        ["hostname", "ipv4"],
        ["ssn"],
        ["full_name"],
        ["first_name", "last_name"],
        ["first_name", "last_name", "phone_number"],
    ],
)
def test_strict_augment_examples_only_use_allowed_labels(
    labels: list[str],
) -> None:
    examples = _select_augment_examples(
        labels=labels,
        strict_labels=True,
    )
    allowed_labels = frozenset(labels)
    assert all(
        example.output_labels <= allowed_labels
        for example in examples
    )
```

This keeps the change internal to `_get_augment_prompt()`, supports arbitrary custom labels honestly, and avoids adding a public prompt configuration system to this PR. If no compatible worked example exists for a strict custom taxonomy, omitting it is safer than giving the model a contradictory example.

asteier2026:
UPDATE: We removed the two full worked examples but kept the bullet point hints. Motivation - we didn't want it to seem like we were gaming the system. RAT-Bench was redone using this new query and we do well.

lipikaramaswamy:
I agree that the general guidance should stay broad. We probably can’t enumerate every identifier/disguise combination.

I think the strict-label concern is slightly different, though. e.g. a prompt limited to `hostname` and `ipv4` still includes a phone-shaped digit sequence and a spelled-out name. 

Would it make sense to keep the generic guidance unconditional, but include the phone/name examples only in permissive mode or when the relevant labels are allowed? That seems like it would preserve the broader detection guidance while keeping strict custom-label prompts focused.

asteier2026:
But the examples we are giving are supposed to be generic. They also apply to fields like hostname and ipv4. I don't think including them will make a custom list that didn't include phone/name go off target. I think it would be hard to deduce which of a custom list might have tricky spelling/digits.

lipikaramaswamy (approved):
Got it—makes sense that the examples are intended to demonstrate general spelling and digit patterns rather than particular entity types. I’m okay with keeping them unconditional.

PR #210, opened 2026-07-02
fix: coerce low combined_risk_level to medium instead of rejecting

asteier2026:
## Summary
 - The LLM occasionally outputs combined_risk_level='low' with a non-leave_as_is
  protection_method_suggestion (e.g. suppress_inference, generalize)
  - Previously EntityDispositionSchema._validate_protection_consistency raised a ValidationError, which
  caused the entire record to be skipped (UPDATE - sometimes the record isn't skipped, but the entity silently gets dropped from the sensitivity disposition)
  - Now combined_risk_level is promoted to medium — preserving the model's intent to protect the entity
  rather than suppressing it. This is more appropriate for weaker models that apply protection methods
  without fully accounting for contextual risk
  - Updated the corresponding test from asserting a ValidationError to asserting that
  combined_risk_level is promoted and the original protection_method_suggestion is retained

UPDATE - the sensitivity disposition prompt is overloaded and doesn't do as well on gpt-oss-120b. We will do another PR to fix that, but above is a quick fix for now

### DIFF
--- a/src/anonymizer/engine/schemas/rewrite.py
+++ b/src/anonymizer/engine/schemas/rewrite.py
@@ -150,8 +150,9 @@ class EntityDispositionSchema(BaseModel):
             self.combined_risk_level == CombinedRiskLevel.low
             and self.protection_method_suggestion != ProtectionMethod.leave_as_is
         ):
-            # LLM occasionally returns an inconsistent combination; coerce rather than reject.
-            self.protection_method_suggestion = ProtectionMethod.leave_as_is
+            # Trust the protection intent over the risk label; promote risk to medium
+            # rather than suppressing the protection.
+            self.combined_risk_level = CombinedRiskLevel.medium
         if (
             self.combined_risk_level == CombinedRiskLevel.high
             and self.protection_method_suggestion == ProtectionMethod.leave_as_is
--- a/src/anonymizer/engine/schemas/rewrite.py
+++ b/src/anonymizer/engine/schemas/rewrite.py
@@ -38,9 +38,12 @@ Supporting enums: Domain, EntitySource, EntityCategory, SensitivityLevel,
 
 from __future__ import annotations
 
+import logging
 from enum import Enum
 
 from pydantic import BaseModel, ConfigDict, Field, ValidationInfo, model_validator
+
+logger = logging.getLogger("anonymizer.schemas.rewrite")
 
 # ---------------------------------------------------------------------------
 # Domain
@@ -150,6 +153,13 @@ class EntityDispositionSchema(BaseModel):
             self.combined_risk_level == CombinedRiskLevel.low
             and self.protection_method_suggestion != ProtectionMethod.leave_as_is
         ):
+            logger.warning(
+                "Entity %d (%r): combined_risk_level='low' conflicts with "
+                "protection_method_suggestion=%r; promoting risk to 'medium'.",
+                self.id,
+                self.entity_value,
+                self.protection_method_suggestion,
+            )
             # Trust the protection intent over the risk label; promote risk to medium
             # rather than suppressing the protection.
             self.combined_risk_level = CombinedRiskLevel.medium
--- a/src/anonymizer/engine/schemas/rewrite.py
+++ b/src/anonymizer/engine/schemas/rewrite.py
@@ -150,10 +150,8 @@ class EntityDispositionSchema(BaseModel):
             self.combined_risk_level == CombinedRiskLevel.low
             and self.protection_method_suggestion != ProtectionMethod.leave_as_is
         ):
-            raise ValueError(
-                f"Entity {self.id}: combined_risk_level='low' requires protection_method_suggestion='leave_as_is', "
-                f"got '{self.protection_method_suggestion}'"
-            )
+            # LLM occasionally returns an inconsistent combination; coerce rather than reject.
+            self.protection_method_suggestion = ProtectionMethod.leave_as_is
         if (
             self.combined_risk_level == CombinedRiskLevel.high
             and self.protection_method_suggestion == ProtectionMethod.leave_as_is
--- a/src/anonymizer/engine/schemas/rewrite.py
+++ b/src/anonymizer/engine/schemas/rewrite.py
@@ -154,10 +154,10 @@ class EntityDispositionSchema(BaseModel):
             and self.protection_method_suggestion != ProtectionMethod.leave_as_is
         ):
             logger.warning(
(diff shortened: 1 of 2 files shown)


src/anonymizer/engine/schemas/rewrite.py
```diff
@@ -150,10 +150,8 @@ def _validate_protection_consistency(self) -> EntityDispositionSchema:
             self.combined_risk_level == CombinedRiskLevel.low
             and self.protection_method_suggestion != ProtectionMethod.leave_as_is
         ):
-            raise ValueError(
-                f"Entity {self.id}: combined_risk_level='low' requires protection_method_suggestion='leave_as_is', "
-                f"got '{self.protection_method_suggestion}'"
-            )
+            # LLM occasionally returns an inconsistent combination; coerce rather than reject.
+            self.protection_method_suggestion = ProtectionMethod.leave_as_is
```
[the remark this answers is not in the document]
asteier2026: fixed

src/anonymizer/engine/schemas/rewrite.py
```diff
@@ -150,10 +150,9 @@ def _validate_protection_consistency(self) -> EntityDispositionSchema:
             self.combined_risk_level == CombinedRiskLevel.low
             and self.protection_method_suggestion != ProtectionMethod.leave_as_is
         ):
-            raise ValueError(
-                f"Entity {self.id}: combined_risk_level='low' requires protection_method_suggestion='leave_as_is', "
-                f"got '{self.protection_method_suggestion}'"
-            )
+            # Trust the protection intent over the risk label; promote risk to medium
+            # rather than suppressing the protection.
+            self.combined_risk_level = CombinedRiskLevel.medium
```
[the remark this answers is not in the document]
asteier2026: Fixed

tests/engine/test_schemas.py
```diff
@@ -244,11 +244,11 @@ def mixed_disposition() -> SensitivityDispositionSchema:
...
-def test_entity_disposition_invalid_low_risk_but_not_leave_as_is() -> None:
-    with pytest.raises(ValidationError, match="combined_risk_level='low'"):
-        EntityDispositionSchema.model_validate(
-            _make_entity(combined_risk_level="low", protection_method_suggestion="replace")
-        )
+def test_entity_disposition_low_risk_non_leave_as_is_is_coerced() -> None:
+    entity = EntityDispositionSchema.model_validate(
+        _make_entity(combined_risk_level="low", protection_method_suggestion="replace")
+    )
+    assert entity.protection_method_suggestion == "leave_as_is"
```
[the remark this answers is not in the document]
asteier2026: Fixed

src/anonymizer/engine/schemas/rewrite.py
```diff
@@ -150,10 +153,16 @@ def _validate_protection_consistency(self) -> EntityDispositionSchema:
...
             and self.protection_method_suggestion != ProtectionMethod.leave_as_is
         ):
-            raise ValueError(
-                f"Entity {self.id}: combined_risk_level='low' requires protection_method_suggestion='leave_as_is', "
-                f"got '{self.protection_method_suggestion}'"
+            logger.warning(
+                "Entity %d (%r): combined_risk_level='low' conflicts with "
+                "protection_method_suggestion=%r; promoting risk to 'medium'.",
+                self.id,
+                self.entity_value,
```
binaryaaron: > **agent** (`review-pr`): Avoid logging raw entity values

`entity_value` is detected PII, so this warning can write a name or identifier to application logs whenever the model emits this inconsistent combination. Please log the entity ID, protection method, and coercion instead, without the raw value.

asteier2026: Fixed.

src/anonymizer/engine/schemas/rewrite.py
```diff
@@ -150,10 +153,16 @@ def _validate_protection_consistency(self) -> EntityDispositionSchema:
...
+            logger.warning(
+                "Entity %d (label=%r): combined_risk_level='low' conflicts with "
+                "protection_method_suggestion=%r; promoting risk to 'medium'.",
+                self.id,
+                self.entity_label,
+                self.protection_method_suggestion,
             )
+            # Trust the protection intent over the risk label; promote risk to medium
+            # rather than suppressing the protection.
+            self.combined_risk_level = CombinedRiskLevel.medium
```
lipikaramaswamy: Non-blocking: because `model_config` uses `use_enum_values=True`, normally validated `combined_risk_level` values are stored and returned by `model_dump()` as plain strings. This assignment happens after validation, so the coerced path instead stores a `CombinedRiskLevel` enum object (although it compares equal to `"medium"`). Could we assign `CombinedRiskLevel.medium.value` here and add a `model_dump()` assertion so coerced and non-coerced records preserve the same serialization contract?

lipikaramaswamy (commented):
One non-blocking serialization consistency suggestion.

asteier2026:
Note that this branch has a different version of rewrite_generation.py than what is being updated in PR 208 - it doesn't have the prereplace logic. How do we handle that?

src/anonymizer/engine/schemas/rewrite.py
asteier2026: Made suggested changs.

lipikaramaswamy: Thanks—the `.value` change looks good. CI is failing because `ty` requires `# ty: ignore[invalid-assignment]`. Also, since `CombinedRiskLevel` subclasses `str`, the `isinstance(..., str)` assertion passes for the old enum path too; could you change it to `type(...) is str` so the regression test actually distinguishes them?

lipikaramaswamy:
> Note that this branch has a different version of rewrite_generation.py than what is being updated in PR 208 - it doesn't have the prereplace logic. How do we handle that?

The difference is expected: #210 is a focused sensitivity-disposition fix and should not pull in prereplace changes. As you mentioned, #208 is now superseded by #246, which implements prereplacement through validated source spans. Since #210 and #246 touch separate files, we can merge #210 independently and let #246 rebase onto the resulting main. We should run an integrated rewrite check after both land, but #210 does not need to carry `rewrite_generation.py` changes itself.

src/anonymizer/engine/schemas/rewrite.py
asteier2026: Suggestions incorporated.

lipikaramaswamy (approved):
Looks good, you'll need to fix the DCO signature to merge tho.

PR #211, opened 2026-07-06
build: migrate developer tasks to mise

binaryaaron:
## Why

Developer commands and CI duplicated task definitions across the Makefile and workflow files, which allowed local and CI behavior to drift. Mise provides one versioned task graph and pins the tools that execute it.

No linked issue required: maintainer-owned developer infrastructure migration.

## Summary

- migrate developer tasks and CI invocation to Mise
- adopt the shared colon-delimited task vocabulary; public task names contain no hyphens or aliases
- compose prerequisites and ordered stages with native Mise task references
- preserve blocking ty 0.0.69 and DataDesigner 0.9 compatibility from current `main`
- require Mise 2026.7.6 or newer; pin uv 0.12.3 and Ruff 0.16.2
- add `setup all` for every locked dependency group and `test:all` for unit plus opt-in E2E suites
- keep benchmark and release workflow orchestration on the current revision while checking arbitrary historical targets into isolated directories
- install each isolated benchmark target's locked development group before running measurement tooling
- run wheel builds and documentation builds and deploys through repository-owned Mise tasks
- harden Mise installation with the fingerprint-pinned, versioned signed installer and a file-backed unsigned fallback download

## Command tree

| Intent | Command |
|---|---|
| Onboarding | `mise run setup [runtime\|dev\|docs\|notebooks\|all]` |
| Dependency sync | `mise run deps:sync [runtime\|dev\|docs\|notebooks\|all]` |
| Repository hooks | `mise run hooks:install` |
| Static checks | `mise run check` and `mise run check:*` |
| Default local gate | `mise run check ::: test` |
| Tests | `mise run test`, `mise run test:all`, `mise run test:coverage`, `mise run test:e2e` |
| Documentation | `mise run docs:build`, `mise run docs:serve`, `mise run docs:deploy VERSION` |
| Notebooks | `mise run notebooks:execute` |
| Build | `mise run build:wheel` |
| Lock update | `mise run lock:update` |

Task names follow `<domain>[:<action>[:<qualifier>...]]`. Arguments select profiles or values. Commands containing `check` preserve tracked files. Mutating tasks state their outputs in `mise tasks`.

The default local gate stays explicit because it runs only static checks and unit tests; affected specialized suites remain separate. No `validate` task or removed Make compatibility alias is restored.

## Task and environment boundaries

- Mise owns pinned tools, environment variables, virtual-environment activation, and task composition.
- uv owns Python dependency resolution and the contents of the project environment.
- `uv run --locked --group <profile>` makes leaf tasks independently runnable and synchronizes the requested locked profile.
- `deps:sync` remains the explicit command for preparing a profile without running another task.
- `setup` remains the shared local and CI entry point. CI uses `--no-hooks` to skip repository hook installation.
- Historical benchmark and release refs supply the target source only. Workflow YAML, local actions, and Mise tasks always come from the PR revision.

## Base and compatibility

- rebased onto `main` at `46bcd1e`
- preserve `data-designer>=0.9,<0.10` and ty 0.0.69 from `main`; no dependency downgrades were introduced while regenerating `uv.lock`
- retain only `make help`, `make install-mise`, and `make setup`
- use `make setup` as the bootstrap path for machines without Mise
- keep Ruff and ty versions aligned between Mise and the uv development environment
- reject unknown dependency profiles through Mise usage validation
- preserve the Python version requested by each CI matrix job

## Migration audits

- Ruff 0.16: enable `E9`, `RUF100`, `UP015`, `UP017`, `UP035`, and `UP037`; apply safe fixes; format and lint tracked Python files and rendered notebooks. The incidental library and tooling formatting edits are mechanical Ruff migrations with no intended behavior change.
- uv 0.12: use locked synchronization for every dependency profile and locked project execution where applicable.
- ty 0.0.69: preserve the blocking repository check and verify that the configured `docs` path discovers all rendered notebooks.
- pre-commit: run the blocking aggregate check, including read-only lock verification; lock regeneration remains the explicit `lock:update` operation.
- Make: confirm no repository automation or documentation depends on the removed forwarding aliases.

## Security

GitHub Actions disables the unsigned fallback. When Mise must be installed, the pinned GPG-verified installer is required and setup fails if verification cannot complete. The installer fetches the pinned release key through HTTPS, verifies its fingerprint, downloads the signed installer from the exact pinned Mise release, and avoids dirmngr keyserver hangs. Local installation retains its documented unsigned fallback for slim development environments, but downloads the script completely to a temporary file before execution so curl retries cannot concatenate partial responses.

The PyPI task validates credentials before building, requires an explicit version, keeps credentials out of process arguments, and supports an upload-free `--dry-run`.

## Validation

- `mise tasks validate`: all 26 tasks passed
- `mise run deps:sync all`: every locked dependency group synchronized successfully
- `mise run --dry-run test:all`: resolved to the unit suite followed by the opt-in E2E suite
- `mise run check ::: test`: 1,213 tests passed; Ruff, ty, lock, formatting, SPDX, and benchmark shell checks passed
- `mise run docs:build`: strict documentation build passed
- `mise run test -- tests/tools/test_mise_configuration.py -q`: 22 focused workflow and tooling tests passed
- current `build:wheel` and `docs:build` tasks built the wheel and strict docs from a separate checkout of historical tag `v0.1.0`
- benchmark shell regression coverage passed on Bash 5.2 and Bash 3.2.57
- `bash -n tools/install-mise.sh`: signed-installer script syntax passed


.github/workflows/benchmark-ci.yml
```diff
@@ -64,27 +64,19 @@ jobs:
 
     steps:
       - name: Checkout benchmark target
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
         with:
           ref: ${{ env.BENCHMARK_REF }}
```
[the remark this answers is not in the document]
binaryaaron: Fixed in 9844192. The workflow now checks out its own revision first, loads `./.github/actions/setup-python-env` from that revision, and only then checks `BENCHMARK_REF` out under `benchmark-target`. Benchmark execution, commit resolution, summaries, and artifact paths are rooted in that separate target directory, so arbitrary historical refs no longer select or need the setup action. Regression coverage is in `test_benchmark_workflow_keeps_setup_on_workflow_revision`; the focused suite passes (17 tests), the full local gate passes (1,208 tests), and the strict docs build passes.

.github/workflows/release.yml
```diff
@@ -51,20 +51,15 @@ jobs:
...
           fetch-tags: true
 
-      - name: Install uv
-        uses: astral-sh/setup-uv@v6
-        with:
-          enable-cache: true
-
-      - name: Set up Python
-        uses: actions/setup-python@v6
+      - uses: ./.github/actions/setup-python-env
```
andreatnvidia, line 50: Could we keep the workflow revision checked out while this local action runs, then check out `release-ref` into a separate directory like `benchmark-ci.yml` does? Right now the action comes from `release-ref`, so older tags use the old setup action: `checkout: "false"` is ignored, the workspace switches back to `github.ref`, and Mise never gets installed. Historical releases therefore fail at the build step, and the docs job has the same issue.

tools/install-mise.sh
```diff
@@ -0,0 +1,248 @@
...
+    fi
+}
+
+# ${VAR:?msg} aborts the script with the given message when VAR is unset or
+# empty. Paired with `:` (the no-op builtin) it becomes an assert-and-die
+# guard that surfaces a clearer error than `set -u`'s "unbound variable".
+MISE_VERSION="${MISE_VERSION:-$(read_pinned_mise_version)}"
+: "${MISE_GPG_KEY:?MISE_GPG_KEY is required (pass from bootstrap caller)}"
+
+readonly MISE_SIG_URL="https://mise.jdx.dev/install.sh.sig"
```
andreatnvidia: Could we fetch `install.sh.sig` from the exact `${MISE_VERSION}` release instead? We currently verify the latest installer and then ask it to install an older pinned version. That path trusts a checksum manifest fetched over HTTPS without verifying its publisher signature, so `MISE_REQUIRE_SIGNED_INSTALL=1` does not protect the binary that runs in release jobs with PyPI credentials. Using the version-pinned signed installer would preserve the signed chain.

.mise/tasks/benchmark
```diff
@@ -0,0 +1,33 @@
...
+    ;;
+  smoke-traces)
+    trace_args=(--dd-trace last_message --dd-task-trace)
+    ;;
+esac
+
+uv run --locked --group dev python tools/measurement/run_benchmarks.py \
+  "$suite" \
+  --output "$output_dir" \
+  "${trace_args[@]}" \
```
andreatnvidia: This breaks the documented smoke benchmark on stock macOS Bash 3.2. With `set -u`, an empty `${trace_args[@]}` is treated as unbound; `smoke` always leaves it empty, and `smoke-traces` hits the same problem when `runner_args` is empty. Could we use Bash 3.2-safe guarded expansions for both arrays and add a macOS shell test?

.mise/tasks/benchmark
```diff
@@ -0,0 +1,33 @@
...
+#USAGE arg "[runner_args]" var=#true double_dash="optional" help="Arguments forwarded to the benchmark runner"
+
+set -euo pipefail
+
+profile="${usage_profile?}"
+suite="tools/measurement/examples/repo-data-smoke.yaml"
+output_dir="${BENCHMARK_OUTPUT_DIR:-benchmark-runs/${profile}}"
+runner_args=()
+if [[ -n "${usage_runner_args:-}" ]]; then
+  read -r -a runner_args <<<"$usage_runner_args"
```
andreatnvidia: It looks like `read -r -a` loses Mise’s quoting here. For example, `--wandb-run-name 'test run'` reaches the runner as `'test` and `run'` instead of one value. Could we use Mise’s documented `eval "runner_args=($usage_runner_args)"` pattern and add a case with an argument containing spaces?

andreatnvidia (changes requested):
The Mise direction looks good. I found three blockers around historical releases, the signed bootstrap, and macOS benchmark execution. There’s also one non-blocking argument-forwarding bug in the same benchmark task.

tools/install-mise.sh
binaryaaron: Fixed in 7ab8add. The signed installer URL now points to the exact `${MISE_VERSION}` GitHub release asset, so the verified installer embeds the checksum for the binary being installed. The focused workflow/tooling suite passes (20 tests), the full local gate passes (1,211 tests), and I also verified that the v2026.7.6 signature decrypts under the pinned fingerprint to the installer with embedded v2026.7.6 checksums.

.mise/tasks/benchmark
binaryaaron: Fixed in 32c79b5. Both optional arrays now use Bash 3.2-safe guarded expansions. `tests/tools/test_benchmark_task.sh` exercises empty arguments for both `smoke` and `smoke-traces`; it passes under Bash 3.2.57 and Bash 5.2. CI now runs that shell regression on `macos-latest`, and the full local gate passes (1,211 tests).

.mise/tasks/benchmark
binaryaaron: Fixed in 32c79b5. The task now uses Mise's documented `eval "runner_args=($usage_runner_args)"` reconstruction for the shell-escaped variadic value. The regression passes `--wandb-run-name 'test run'` and asserts that `test run` reaches the runner as one argument. The focused shell suite and the full 1,211-test local gate pass.

.github/workflows/release.yml
binaryaaron: Addressed in stacked follow-up #249. Instead of adding a second release checkout, it follows Safe Synthesizer’s tag-push strategy: pushed v* tags run the full production release from the triggering revision, while workflow_dispatch remains available with dry-run and create-gh-release controls but no release-ref input. Every job uses its ordinary checkout ref, so workflow orchestration, setup, build, and docs stay on one revision. The focused workflow suite passes (21 tests), the full local gate passes (1,212 tests), and the strict docs build passes. I am leaving this thread open until the stacked follow-up lands.

binaryaaron: Fixed in 019efb4. The workflow now checks out its own revision first, runs the current setup action and Mise tasks from that checkout, and places the requested historical ref under release-target. Build, upload, and docs commands target that directory, while GitHub release creation passes the requested ref explicitly with --target. Regression tests cover checkout order and every target path. I also verified the current tasks by building the v0.1.0 wheel and strict docs from a separate historical checkout. I will resolve this thread once the pushed CI run is green.

.github/workflows/benchmark-ci.yml
```diff
@@ -114,7 +112,7 @@ jobs:
             FAIL_FAST_ARGS+=(--fail-fast)
           fi
 
-          uv run python tools/measurement/run_benchmarks.py \
+          uv run --locked python tools/measurement/run_benchmarks.py \
```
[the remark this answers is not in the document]
binaryaaron: Fixed in 3cea487. The isolated benchmark command now runs the target with uv run --locked --group dev, matching the repository benchmark task and installing the measurement integrations declared in the dev group. The workflow contract test now asserts that group explicitly. The focused suite passes (22 tests), and the full local gate passes (1,213 tests). I will resolve this thread once the pushed CI run is green.

Issue #164, opened 2026-05-19
feat: add a deny-list for entities in NA


memadi-nv:
### Priority Level

Medium

### Task Summary

Add a list of entities that are not desired to be detected in NA.

PR #243, opened 2026-08-11
feat: add excluded labels for entities in Anonymizer

memadi-nv:
## Summary

Adds `excluded_entity_labels` to the `Detect` configuration, allowing users to explicitly exclude specific entity label types from detection. Denied labels are removed from detector scope, filtered from final explicit and latent entity outputs, and excluded from entity-coverage scoring.

### Config
- Added `excluded_entity_labels: list[str] | None` field to `Detect` with the same normalisation as `entity_labels` (strip, lowercase, deduplicate, reject empty)
- Added model validator that warns at config construction time when `entity_labels` and `excluded_entity_labels` share labels that would never be detected.

### Detection Engine
- Subtracted denied labels from the effective label set before GLiNER and LLM prompts run (`_resolve_detection_labels`)
- Added safety-net filter in `_materialize_final_entities` to drop denied labels from final entity output regardless of LLM output
- Threaded `excluded_entity_labels` through all detection pipeline methods: `run`, `detect_and_validate_entities`, `_build_detection_spec`, `identify_latent_entities`
- Threaded through export paths: `build_detection_config` and `build_detection_builder_for_seed`
- Added warning when the denylist empties the effective label set (GLiNER returns no detections for an empty list, not defaults)
- Rewrite latent detection receives the denylist in its prompt and deterministically filters denied latent labels from its output.
- Final allowlist and denylist comparisons are case- and whitespace-insensitive.

### Evaluation
- Added `excluded_entity_labels` field to `AnonymizerResult` and `PreviewResult` so `evaluate()` has access to the full detection scope
- Propagated `excluded_entity_labels` through `evaluate()` to `EntityCoverageWorkflow` and `ReplacementWorkflow.evaluate()`
- Explicit allowlists have denied labels removed. When entity_labels=None, coverage remains permissive so novel augmenter-defined labels are evaluated, while denied labels are independently excluded in the prompt and postprocessing.

### Telemetry
- Added `excluded_entity_labels` to `_detect_config_metadata` so it is captured in measurement records alongside `entity_labels`

### Tests
- Config validation: field normalisation, empty list rejection, overlap warning
- Detection engine: label resolution, GLiNER injection, final entity filtering, combined allowlist + denylist, export paths
- Evaluation: `_effective_entity_labels` helper, coverage judge postprocess filter, coverage prompt scoping
- Telemetry: `excluded_entity_labels` captured correctly when set and absent
- Case-insensitive denylist enforcement.
- Latent-entity filtering.
- Preservation of novel non-denied labels during coverage evaluation.

### Docs
- Updated `Detect` fields reference table and added "Excluding labels with a deny list" section in `detection.md`
- Added `excluded_entity_labels` to the judge scoping list in `evaluation.md`
- Added `excluded_entity_labels` section and corrected the `data_summary` soft-exclusion claim in `choosing-a-strategy.md`
- Added `excluded_entity_labels` as the recommended fix for systematically noisy label types in `troubleshooting.md`
- Updated `skills/anonymizer/SKILL.md` with denylist behavior and precedence rule

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] CI, release, or contributor workflow update

## Contributor Checklist
- [ ] PR title follows Conventional Commits, for example `fix: handle empty entity list`
- [ ] Related issue is linked, or a maintainer-owned no-issue reason is documented above
- [ ] For non-trivial changes, a plan document is linked above, or the no-plan reason is documented above
- [ ] Public API impact checked; `skills/anonymizer/SKILL.md` updated if needed
- [ ] No real PII added to tests, docs, notebooks, fixtures, or artifacts
- [ ] No API keys, service tokens, private keys, credentials, or real endpoint secrets added

## Validation

- Commands run:

- [ ] Docs updated, or not needed
- [ ] If tutorial sources changed: notebooks regenerated with `make convert-notebooks`
- [ ] If e2e, benchmark, or model-provider behavior changed: relevant validation is listed above

closes issue #164


src/anonymizer/engine/detection/detection_workflow.py
```diff
@@ -143,7 +146,7 @@ def _build_detection_spec(
         and :meth:`build_detection_config` (which exports it for an external runtime),
         so both paths run exactly the same workflow.
         """
-        labels = _resolve_detection_labels(entity_labels)
+        labels = _resolve_detection_labels(entity_labels, set(entity_label_denylist) if entity_label_denylist else None)
```
binaryaaron: **[P1] Filter denied augmented labels before resolving overlaps**

In permissive mode, the augmenter can still return denied labels. These labels enter overlap resolution before the final deny-list filter, so a denied span can replace an allowed span and then be removed, leaving the allowed PII undetected.

Please filter denied augmented entities before merging them, and add a regression test confirming an overlapping allowed span survives.

lipikaramaswamy:
The param name `entity_label_denylist` is not precise and is not likely to be intuitive to a user. `excluded_entity_labels` (preferred) or `ignored_entity_labels` sounds better to me; open to other suggestions. We should agree on the name before we make changes to the public API.

src/anonymizer/engine/detection/detection_workflow.py
lipikaramaswamy: Yes, this is critical to fix. Do note that #238 will change the mechanism for overlap resolution in `resolve_overlaps`, but regardless of the resolution mechanism we should filter out undesired entity labels after.

src/anonymizer/interface/anonymizer.py
```diff
@@ -455,6 +458,7 @@ def evaluate(
             raise InvalidConfigError(str(exc)) from exc
 
         entity_labels = getattr(output, "entity_labels", None)
+        entity_label_denylist = getattr(output, "entity_label_denylist", None)
```
lipikaramaswamy: We should preserve the existing result constructor contract. 

These result classes are public, so their constructor parameters and attributes form part of the user-facing API. Adding the new field before `data_summary` changes the positional signature, even though positional construction is probably uncommon.

Could we place the new optional field after `data_summary` to preserve the existing parameter order? Longer term, it may be worth grouping evaluation provenance into a dedicated context object, but replacing the existing fields would require a separately planned API change.

src/anonymizer/engine/detection/detection_workflow.py
memadi-nv: Good catch! added a filter for excluded entity labels after augment and a follow up test.
Thanks both.

memadi-nv:
> The param name `entity_label_denylist` is not precise and is not likely to be intuitive to a user. `excluded_entity_labels` (preferred) or `ignored_entity_labels` sounds better to me; open to other suggestions. We should agree on the name before we make changes to the public API.

`excluded_entity_labels` is reasonable. Updated. Thanks!

src/anonymizer/engine/detection/detection_workflow.py
memadi-nv, line 486: when `entity_labels` and `excluded_entity_labels` fully overlap, `_resolve_detection_labels` returns [] and we warn, but `_inject_detector_params` still sends `extra_body["labels"] = []` to the GLiNER endpoint as-is.

Our reference server `(tools/serve_gliner.py)` treats empty labels as "detect nothing" — but the default production endpoint (`nvidia/gliner-pii` on build.nvidia.com) is a NIM we don't control, and its empty-label behavior is unverified. If it treats [] as "no filter," we'd silently detect everything instead of nothing — the opposite of what we log and promise.

I'm leaning towards hard-error, not skip-and-warn: `validate_entity_labels` already rejects `entity_labels=[] `directly ("must not be empty, use None"). Full overlap reaches the exact same empty-effective-label-set state through a second path, but only warns — an inconsistency, not a deliberate design choice. There's also no real use case for a config that fully cancels itself out; it's always a mistake (stale exclusion, copy-paste, merge error), and warn-and-continue lets that mistake run silently through a whole job with zero detections.

Please let me know your thoughts on this and what you think is the better approach.

lipikaramaswamy: Yup I think it makes sense to have a validation error when an explicitly provided `entity_labels` list is fully canceled by `excluded_entity_labels`. Failing early.

memadi-nv: changes were applied.

lipikaramaswamy: The explicit allowlist case looks good. `entity_labels=None` with all default labels excluded can still make `_resolve_detection_labels` return `[]`. Can we guard that path too?

binaryaaron (changes requested):
> **agent** (`review-pr`): Blocking empty effective-label path remains

**[P1] Reject exclusions that remove every default label**

The explicit-overlap validator does not run when `entity_labels=None`. On the current head (`f3cffb5`), this remains valid:

```python
Detect(
    entity_labels=None,
    excluded_entity_labels=list(DEFAULT_ENTITY_LABELS),
)
```

Both the standard and latent detection paths call `_resolve_detection_labels()`. It returns `[]`, logs a warning, and the workflow injects that empty list into the detector configuration.

A local regression test confirms the missing guard:

```text
Failed: DID NOT RAISE <class 'ValueError'>
```

With the bundled GLiNER server, this silently disables detection. Other detector endpoints may interpret an empty label list differently. Either outcome violates the intended fail-early behavior for an empty effective detection set.

Please treat this as blocking and ensure no workflow can send `labels=[]` to the detector.

<details>
<summary>Prompt for an implementation agent</summary>

Fix the remaining empty-effective-label defect in PR #243.

Requirements:

- Reproduce the defect with `entity_labels=None` and `excluded_entity_labels` containing every value in `DEFAULT_ENTITY_LABELS`.
- Reject that configuration with a clear `ValueError` before a detector workflow executes.
- Ensure both callers of `_resolve_detection_labels()`—the standard detection specification and latent-entity detection—cannot inject `labels=[]`.
- Add a regression test for the default-label case. Update the existing warning test if the new invariant makes it obsolete.
- Preserve partial default exclusions, label normalization, deduplication, and explicit-label overlap behavior.
- Keep the invariant in the narrowest owning layer; do not duplicate the effective-label calculation across callers.
- Use `astnav refs _resolve_detection_labels` to verify the complete caller set before editing.
- Run the focused config and detection-workflow tests, Ruff format/lint, and `git diff --check`.

Report the changed files, the ownership decision, and the exact verification results.

</details>

src/anonymizer/engine/detection/detection_workflow.py
memadi-nv: Good catch, just addressed this. Thanks!

docs/concepts/detection.md
```diff
@@ -104,6 +105,22 @@ Detect(entity_labels=["first_name", "last_name", "email"])
...
+```python
+# Detect all defaults except occupation and gender
+Detect(excluded_entity_labels=["occupation", "gender"])
+
+# Combine with an explicit allowlist — exclusions always win
+Detect(entity_labels=["first_name", "email", "city"], excluded_entity_labels=["city"])
+```
+
+!!! warning
+    `excluded_entity_labels` is always checked against the effective allowlist — `entity_labels` if set, otherwise `DEFAULT_ENTITY_LABELS`. A partial overlap just drops the shared labels and logs a warning. If the overlap is total, leaving an empty effective detection set, `Detect` raises a `ValueError` at config time instead of silently detecting nothing.
```
lipikaramaswamy: Small mismatch: this says a partial overlap always logs a warning, but the validator only warns when `entity_labels` is explicitly set. Partial exclusions from `DEFAULT_ENTITY_LABELS` are accepted without a warning, which is also the common use case. Could we clarify that distinction here and in the matching warning in `docs/concepts/choosing-a-strategy.md`?

lipikaramaswamy (approved):
Small nit, but looks great overall!

docs/concepts/detection.md
memadi-nv: addressed. Thank you!

binaryaaron (approved):

PR #246, opened 2026-08-12
fix: harden rewrite replacement and leak metrics

binaryaaron:
## Summary

- apply replacement-map entries to validated source spans before rewrite generation and carry that baseline through repair
- emit schema-v2 metrics that distinguish map entries, targeted/applied/skipped spans, unique leaked values, and associated source occurrences
- separate post-validation detection artifacts from materialized final entities and keep legacy schema-v1 ingestion isolated
- consolidate local and generated-map DataFrame orchestration around one span-application helper, without `iterrows()` or per-row Series construction

## Why

The schema-v1 leak metric used raw substring matching for longer values and counted every same-valued source entity when one match appeared in output. This produced false positives such as `lawyer` in `lawyers`. Adjacent replacement metrics also described map membership rather than actual span execution, and detection sidecars exported post-validation detections under `final_*` names.

The rewrite path additionally relied on the model to apply deterministic replacement-map entries. This change uses the existing offset-based replacement boundary, fails closed when required mappings cannot be applied, and prevents the repair loop from returning to the original sensitive baseline. It supersedes the global-regex runtime approach proposed in #208.

Both local replacement and LLM-generated replacement maps now use the same DataFrame orchestration helper. The helper iterates aligned columns directly, preserves row order and custom column names, and records replacement text and application telemetry together.

## Compatibility

- new measurements emit schema version 2 and explicit metric names
- strict ingress continues to accept schema-v1 records
- mixed schema snapshots and cross-version metric fields are rejected
- benchmark groups include measurement schema version, so v1 and v2 series are not aggregated together
- legacy artifact sidecars are interpreted as detected metrics; final metrics remain unavailable unless actual final entities are present

## Validation

- `make check`
- `make test` (1,219 passed, one existing W&B deprecation warning)
- `uv run pytest tests/engine/test_replace_strategies.py tests/engine/test_replace_runner.py -q` (21 passed)
- `git diff --check`
- Astnav syntax query confirms zero `iterrows()` calls in `strategies.py`


tools/measurement/analyze_benchmark_output.py
```diff
@@ -471,11 +518,16 @@ def _build_case_row(
     records_per_ndd_sec = _safe_rate(record_count, ndd_elapsed_sec_total)
     input_text_tokens_per_pipeline_sec = _safe_rate(input_text_tokens_total, pipeline_elapsed_sec)
     input_text_tokens_per_ndd_sec = _safe_rate(input_text_tokens_total, ndd_elapsed_sec_total)
+    measurement_schema_version = _first_int([measurement_rows], ["schema_version"])
+    leak_presence_column = (
+        "original_value_leak_unique_value_count" if measurement_schema_version == 2 else "original_value_leak_count"
+    )
     final_entity_count = _coalesce_number(
         _sum_or_none(record_rows, "final_entity_count"),
         _sum_or_none(artifact_rows, "final_entity_count"),
```
lipikaramaswamy: It looks like this fallback still treats a legacy sidecar `final_entity_count` as an actual final-entity measurement. Elsewhere in this change, legacy `final_*` fields are correctly interpreted as detected metrics because those sidecars did not contain materialized final entities.

For example, a legacy sidecar with `final_entity_count=2` and no record-level final count produces `artifact_detected_entity_count=2` and `artifact_final_entity_count=None`, but still sets the case-level `final_entity_count=2`. Could we limit this fallback to schema-v2 artifact rows that contain actual final entities, and add a regression test for the legacy-sidecar case?

tests/engine/test_rewrite_generation.py
```diff
@@ -205,6 +212,94 @@ def test_filter_replacement_map_accepts_schema_instance(
...
+    assert result[COL_REPLACEMENT_APPLICATION]["applied_span_count"] == 2
+
+
+def test_prepare_rewrite_tagged_text_fails_closed_for_partial_map() -> None:
+    row = {
+        COL_TEXT: "Alice",
+        COL_FINAL_ENTITIES: {
+            "entities": [{"value": "Alice", "label": "first_name", "start_position": 0, "end_position": 5}]
+        },
+        COL_REPLACEMENT_MAP: {"replacements": []},
```
lipikaramaswamy: Nice to see the fail-closed path covered here. Would it be worth making this a genuinely partial-map case—for example, two required replacement entities with a mapping for only one? The current empty replacement list exercises the missing-map behavior again, but does not independently guard the partially complete-map case.

lipikaramaswamy:
General note: this PR correctly supersedes #208 by using validated entity spans instead of global regex replacement, preventing embedded-substring corruption such as Ann inside Anna. It verifies that every required span was replaced and fails closed when a map is absent, partial, ambiguous, or cannot be applied. It also preserves the sanitized baseline through repair and reports only safe labels and counts in diagnostics. cc @asteier2026

src/anonymizer/engine/rewrite/rewrite_generation.py
```diff
@@ -152,36 +165,125 @@ def _format_rewrite_disposition_block(row: dict[str, Any]) -> dict[str, Any]:
...
+        {"original": replacement.original, "label": replacement.label, "synthetic": replacement.synthetic}
+        for replacement in _parse_replacements(raw_map)
+        if (replacement.original, replacement.label) in replace_pairs
     ]
     row[COL_REPLACEMENT_MAP_FOR_PROMPT] = {"replacements": filtered}
     return row
 
 
-@custom_column_generator(required_columns=[COL_FULL_REWRITE])
+@custom_column_generator(
```
andreatnvidia, line 190: Could we declare `_replacement_application`, `_rewrite_replacement_ready`, and `_rewrite_baseline_text` in `side_effect_columns` here? This generator writes all three, but Data Designer removes undeclared columns after the function returns. `_extract_rewritten_text` then requires `_rewrite_replacement_ready`, so the record gets omitted; the repair baseline and replacement-application metrics also never make it downstream.

The current tests call this function directly, which bypasses that validation. An execution-level test through the Data Designer generator would catch it. It would also be good to add `COL_TAGGED_TEXT` to `required_columns`, since the fallback path reads it.

tools/measurement/analyze_benchmark_output.py
```diff
@@ -383,6 +415,21 @@ def analyze_benchmark_output(
     )
 
 
+def _validate_measurement_schema_versions(measurements: pd.DataFrame) -> None:
+    if "schema_version" not in measurements.columns:
+        return
+    versions = measurements["schema_version"].dropna()
```
andreatnvidia, line 423: Could we validate `schema_version` before `pd.read_json` performs dtype inference? In a file containing versions `1` and `true`, pandas coerces both values to integer `1`, so this check accepts the boolean as schema v1. The same thing can happen with `1.0`.

The existing single-boolean test passes because a bool-only column remains boolean, so it misses the mixed-value case. Reading the raw JSON values and requiring `type(value) is int`, plus adding a `[1, true]` regression test, should cover this.

src/anonymizer/engine/rewrite/rewrite_generation.py
binaryaaron: Fixed in 061752a. The generator now declares _replacement_application, _rewrite_replacement_ready, and _rewrite_baseline_text as side effects and _tagged_text as required. A regression executes the real Data Designer CustomColumnGenerator and verifies all side effects survive.

tools/measurement/analyze_benchmark_output.py
binaryaaron: Fixed in 061752a. Case-level final_entity_count now falls back only to detection-artifact rows with artifact_schema_version=2; legacy final_* sidecars remain detected-only. Added legacy-only and mixed legacy/v2 regressions.

tests/engine/test_rewrite_generation.py
binaryaaron: Fixed in 061752a. The test now uses two required first_name spans with a valid mapping for Alice only. It asserts 2 targeted, 1 applied, 1 skipped, and fail-closed readiness.

tools/measurement/analyze_benchmark_output.py
binaryaaron: Fixed in 061752a. Measurement schema values are now streamed from raw JSONL and required to have exact int type before pandas loads the frame. Mixed [1, true] and [1, 1.0] regressions both reject the input.

lipikaramaswamy (approved):
Reviewed current head 061752a. The follow-up commit addresses the replacement, repair, schema-versioning, Data Designer side-effect, legacy artifact, and partial-map test concerns. No remaining code-review blockers from my review.

andreatnvidia (approved):

PR #265, opened 2026-09-09
feat(detection): add regex entity detection

### LINKED ISSUE #262: feat(detection): add regex and validator entity detection
### Priority Level

Medium (Nice to have)

NeMo Anonymizer currently relies on GLiNER, LLM validation, and LLM augmentation for entity detection. This works well for contextual entities, but users cannot express deterministic formats through supported configuration, and strongly structured identifiers still depend on model behavior.

Several entity classes have conservative candidate patterns and deterministic validity checks. Users also need a supported way to detect organization-specific identifiers by supplying an entity label, regular expression, and optional local validator.

Add regex-backed entity detection as a native seed source alongside GLiNER.

- Provide a curated registry of built-in recognizers. Built-ins are enabled by default and run only when their output labels are in the effective detection label set. `Detect(builtin_regexes=False)` explicitly disables them.
- Begin with benchmark-backed recognizers for `credit_debit_card`, `email`, `ipv4`, `ipv6`, `mac_address`, and `url`.
- Allow built-in recognizers to combine candidate regexes with local parsers, structural checks, or checksums based on documented standards.
- Add public `RegexRule` configuration with a label, pattern, optional validator, and `validate_with_llm` setting.
- Accept a direct validator callable for local Python execution. Serialized and exported workflows use stable validator names supplied by trusted installed packages.
- Default `validate_with_llm=True` for every rule. When set to `False`, a candidate is accepted after regex and local validation; with no local validator, the regex match is authoritative.
- Merge regex and GLiNER candidates with deterministic duplicate, overlap, stable-ID, ordering, provenance, and validation-route semantics.
- Reject invalid or empty-matching patterns during configuration and bound regex execution and candidate counts at runtime.

Built-in usage requires no new rule configuration:

```python
config = AnonymizerConfig(
    detect=Detect(entity_labels=["email", "ipv4", "credit_debit_card"]),
    replace=Redact(),
)
```

Custom rule usage:

```python
def validate_support_case(candidate: RegexCandidate) -> bool:
    return not candidate.groups["number"].startswith("0000")


config = AnonymizerConfig(
    detect=Detect(
        entity_labels=["email", "support_case_id"],
        regex_rules=[
            RegexRule(
                label="support_case_id",
                pattern=r"(?<![A-Z0-9])CASE-(?P<number>[0-9]{8})(?![A-Z0-9])",
                validator=validate_support_case,
                validate_with_llm=False,
            )
        ],
    ),
    replace=Redact(),
)
```

Custom validator callables receive a typed candidate containing the original value, offsets, named capture groups, bounded context, and rule identity. They return either `bool` or a typed result containing a rejection reason and optional normalized value. The original source value and offsets remain authoritative.

Whether a decorator should optionally attach a stable validator name and version will be discussed during PR review. Direct callables remain the primary local Python experience.

Acceptance criteria:

- Public typed configuration supports built-ins, custom label/regex rules, optional validators, and `validate_with_llm` per rule.
- Built-ins activate only for labels in the effective detection scope; custom-label behavior is defined for implicit and explicit `entity_labels`.
- Rules with `validate_with_llm=True` enter the existing contextual validation path; rules with `False` bypass that payload after local acceptance.
- Direct callable validators work locally. Exported workflows resolve stable validator names from installed packages and fail preflight when a validator is unavailable.
- An exact same-label/span deterministic acceptance is not forced back through LLM validation by a duplicate GLiNER candidate.
- Configuration rejects malformed, duplicate, empty, and empty-matching rules with errors identifying the offending value.
- Runtime safeguards cover pathological patterns and excessive candidate production.
- Behavior-focused tests use fabricated data and cover matching, built-in and custom validation, both LLM routes, merge conflicts, serialization, workflow integration, and unchanged behavior when built-ins are disabled.
- Boundary tests cover structured values directly adjacent to Han characters without relying on whitespace or Unicode `\b` behavior.
- Quality evaluation reports per-label precision and recall, detector disagreement, language/script/locale slices, latency, and LLM candidate/call volume.
- Public configuration and detection documentation are updated, including geography, script, and standards caveats.
- The bundled anonymizer skill is updated because the public `Detect` surface changes.

- Continue using GLiNER and LLM augmentation alone. This does not provide deterministic support for structured or organization-specific identifiers.
- Merge the benchmark implementation from PR #183. That work is intentionally benchmark-focused and does not provide the public configuration or production integration contract.
- Treat validator callables as directly serializable data. The selected design separates local callable execution from stable installed identifiers used by exported workflows.
- Replace GLiNER for regex-covered labels. Keeping both sources initially provides disagreement evidence and model recovery of unusual valid formats.

### Additional context

- PR #183 explores regex-backed benchmark strategies and provides reusable test ideas.
- PR #258 documents the existing hybrid detection pipeline and benchmark context.
- Issue #259 proposes per-label prompt examples; examples complement deterministic rules but do not enforce a format.
- The internal `validated-pii-benchmark` provides deterministic curation evidence and fixtures for card, email, IPv4, IPv6, MAC, and URL validation.
- The Chinese GLiNER findings guide motivates non-whitespace boundary tests, exact-offset checks, and language/script/locale evaluation slices.

The implementation plan is stored at `plans/262/hybrid-regex-detection.md` and will be linked from the implementation PR. It cites primary documentation for recognizer design and each built-in validator.

lipikaramaswamy:
## Related Issue

Fixes #262.

## Plan Document

[Hybrid regex entity detection plan](https://github.com/NVIDIA-NeMo/Anonymizer/blob/lipikaramaswamy/feature/262-regex-entity-detection/plans/262/hybrid-regex-detection.md)

## Summary

Adds regex-based entity detection as a first-class detection source alongside GLiNER and LLM detection.

- Enables built-in rules by default for requested credit/debit card, email, IPv4, IPv6, MAC address, and URL labels.
- Introduces one unified `Detect.regex_rules` configuration surface for built-in customization and user-defined rules.
- Supports optional local validators and per-rule LLM validation controls, defaulting LLM validation to enabled.
- Allows user-defined rules to replace built-in behavior for the same label without a separate overrides API.
- Adds safe bounded matching, stable content-derived rule IDs, deterministic source-aware merging, and DataDesigner serialization/plugin support.
- Documents the user experience and updates the bundled Anonymizer skill.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update
- [ ] Refactoring
- [ ] CI, release, or contributor workflow update

## Contributor Checklist

- [x] PR title follows Conventional Commits (`type(scope): description`).
- [x] Related issue is linked.
- [x] Plan document is linked.
- [x] Tests cover the new behavior.
- [x] Public API and documentation changes are included.
- [x] Commits include a DCO sign-off.

## Validation

- Commands run:
  - `TMPDIR=/tmp make test` — 1312 passed, 1 warning.
  - `make format-check` — passed.
  - Targeted `ty check` for changed files — passed.
  - Full commit hooks in a clean uv environment — passed, including format, typecheck, lock, and copyright checks.
  - `make docs-build` from a clean worktree at this commit — passed.
- Skipped checks or known failures: None.

- [x] Docs updated, or not needed.
- [x] If docs changed: `make docs-build` passes locally.
- [x] If tutorial sources changed: notebooks regenerated and checked.
- [x] If end-to-end behavior changed: relevant e2e checks completed, or not applicable.


src/anonymizer/engine/detection/custom_columns.py
```diff
@@ -172,7 +191,9 @@ def apply_validation_and_finalize(row: dict[str, Any]) -> dict[str, Any]:
         entities=merged,
         validation_output=row.get(COL_VALIDATED_ENTITIES, {}),
     )
-    expanded = expand_entity_occurrences(text=text, entities=validated)
+    accepted_regex = _parse_entity_spans(row.get(COL_REGEX_ACCEPTED_ENTITIES, {}))
+    protected = merge_entity_sources(accepted_regex, validated)
```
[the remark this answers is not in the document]
lipikaramaswamy: Fixed in 46907f81. Validation routes are now recombined by candidate provenance, preserving regex_user > regex_builtin > other detector sources regardless of validate_with_llm. Regression tests cover both route permutations and the finalization merge.

src/anonymizer/engine/detection/regex_detection.py
```diff
@@ -0,0 +1,416 @@
...
+                    continue
+                match_count += 1
+                if match_count > max_matches_per_rule:
+                    raise RuntimeError(
+                        f"Regex rule {rule.rule_id!r} exceeded the maximum of "
+                        f"{max_matches_per_rule} matches for one record."
+                    )
+                if rule.label == "url":
+                    trimmed = text[start:end].rstrip(_URL_TRAILING_PUNCTUATION)
+                    end = start + len(trimmed)
```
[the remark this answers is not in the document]
lipikaramaswamy: Fixed in 46907f81. URL suffix trimming is now delimiter-balance-aware: balanced delimiters remain part of the URL, while unmatched prose closers are removed. Tests cover nested ASCII delimiters, CJK pairs, unmatched closers, IPv6 host brackets, and exact offsets.

src/anonymizer/engine/detection/regex_detection.py
```diff
@@ -0,0 +1,438 @@
...
+        label="credit_debit_card",
+        pattern=r"(?<![0-9])(?:[0-9][ -]?){12,18}[0-9](?![0-9])",
+        validator_id="nemo.credit-card-luhn.v1",
+        source="regex_builtin",
+    ),
+    ResolvedRegexRule(
+        rule_id="nemo.email.v1",
+        label="email",
+        pattern=(
+            r"(?<![A-Za-z0-9.!#$%&'*+/=?^_`{|}~-])"
```
andreatnvidia, line 412: Some valid international email domains are missed because these character classes exclude Unicode combining marks. I reproduced this with `x@उदाहरण.भारत` and decomposed `x@éxample.com`; both domains encode successfully with IDNA but produce no match here. Could we allow IDNA-valid marks, possibly after normalization, and add a non-CJK IDN test?

src/anonymizer/engine/detection/regex_detection.py
```diff
@@ -0,0 +1,438 @@
...
+        rule_id="nemo.ipv4.v1",
+        label="ipv4",
+        pattern=r"(?<![0-9.])(?:[0-9]{1,3}\.){3}[0-9]{1,3}(?![0-9]|\.[0-9])",
+        validator_id="nemo.ipv4.v1",
+        source="regex_builtin",
+    ),
+    ResolvedRegexRule(
+        rule_id="nemo.ipv6.v1",
+        label="ipv6",
+        pattern=r"(?<![0-9A-Fa-f:])(?:[0-9A-Fa-f]{0,4}:){2,7}[0-9A-Fa-f]{0,4}(?![0-9A-Fa-f:])",
```
andreatnvidia: IPv4-embedded IPv6 addresses get split unexpectedly here. For example, `::ffff:192.0.2.128` is detected as `::ffff:192`, which `IPv6Address` accepts as a different valid address. With IPv4 detection enabled, the dotted portion may be replaced while `::ffff:` remains visible. Could we extend the pattern to consume dotted IPv4 tails and add an exact-span test?

src/anonymizer/engine/detection/regex_detection.py
```diff
@@ -0,0 +1,438 @@
...
+
+def _register_or_resolve_validator(
+    validator: RegexValidatorCallable | str | None,
+) -> str | None:
+    if validator is None:
+        return None
+    if isinstance(validator, str):
+        _resolve_validator(validator)
+        return validator
+    module = getattr(validator, "__module__", "")
```
andreatnvidia, line 254: One edge case with the local validator registry: two closures returned by the same factory share the same module and qualified name, so registering the second raises a duplicate-registration error even though it is a different callable. An opaque identity-based key for direct callables would avoid the collision while keeping stable names for explicit entry-point validators.

src/anonymizer/config/regex.py
```diff
@@ -0,0 +1,121 @@
...
+            raise ValueError("Regex rule pattern must not be empty.")
+        if len(value) > MAX_REGEX_PATTERN_LENGTH:
+            raise ValueError(
+                f"Regex rule pattern length {len(value)} exceeds the maximum of {MAX_REGEX_PATTERN_LENGTH}."
+            )
+        try:
+            compiled = regex.compile(value)
+        except regex.error as exc:
+            raise ValueError(f"Invalid regex pattern {value!r}: {exc}") from exc
+        match = compiled.search("")
```
andreatnvidia: I think contextual zero-width rules can slip through this check. Patterns such as `(?=CASE)` and `(?<=A)` construct successfully, but detection later skips all their zero-length matches, so the rule silently does nothing. It would be safer to reject these during configuration validation, with a couple of lookahead and lookbehind tests.

src/anonymizer/engine/detection/regex_detection.py
```diff
@@ -0,0 +1,438 @@
...
+    r"|^(?:[0-9A-Fa-f]{4}\.){2}[0-9A-Fa-f]{4}$"
+)
+
+
+def _validate_mac(candidate: RegexCandidate) -> bool:
+    return _MAC_VALIDATION_RE.fullmatch(candidate.value) is not None
+
+
+def _validate_url(candidate: RegexCandidate) -> bool:
+    target = candidate.value if not candidate.value.startswith("www.") else f"https://{candidate.value}"
```
andreatnvidia: Small case-sensitivity mismatch here: the regex accepts `WWW.example.com`, but this prefix check only recognizes lowercase `www.`. That leaves the value without a scheme, and `urlsplit` rejects it. A case-insensitive check such as `candidate.value.lower().startswith("www.")` should cover it.

skills/anonymizer/BENCHMARK.md
```diff
@@ -1,85 +1,125 @@
-<!-- SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved. -->
-<!-- SPDX-License-Identifier: Apache-2.0 -->
+# Skill Benchmark: anonymizer
```
andreatnvidia, line 1: The copyright check currently flags this file and `skills/anonymizer/skill-card.md` because they are missing the repository's SPDX headers. Adding the standard headers to both should clear the aggregate CI failure.

plans/262/hybrid-regex-detection.md
```diff
@@ -0,0 +1,660 @@
...
+
+# Hybrid Regex Detection
+
+## Status
+
+Implementation plan for
+[issue #262](https://github.com/NVIDIA-NeMo/Anonymizer/issues/262). This
+document scopes the deterministic regex and validator feature as a focused
+slice of the broader
+[Multi-Pole Detection](../detection-poles/multi-pole-detection.md) design.
```
andreatnvidia: This link points to `plans/detection-poles/multi-pole-detection.md`, which isn't present in the PR. Since the broader framework is larger than this change, could we replace it with a follow-up issue or design doc? A short scope covering common detector output, configurable fan-in, and execution/failure contracts would make the intended direction clear without expanding this implementation.

plans/262/hybrid-regex-detection.md
```diff
@@ -0,0 +1,660 @@
...
+- LLM keep/drop/reclass counts for regex candidates;
+- accepted candidate counts by `validate_with_llm` route;
+- final entity counts by source and label;
+- local detection duration;
+- regex timeout and match-limit failures;
+- validation token/call changes relative to baseline.
+
+Avoid logging raw entity values in aggregate telemetry.
+
+The source presentation asks whether deterministic and model detection can run
```
andreatnvidia: Data Designer 0.9.1 already schedules independent graph columns concurrently, so the regex and GLiNER branches should overlap today. I think this paragraph can describe parallel poles as part of the current graph contract and leave the shared resource, timeout, cancellation, and failure model for the follow-up design.

plans/262/hybrid-regex-detection.md
```diff
@@ -0,0 +1,660 @@
...
+
+## Merge and Overlap Policy
+
+Add a source-aware `merge_detection_sources()` helper rather than globally
+changing `resolve_overlaps()`, because the latter also affects augmentation,
+name splitting, and occurrence propagation.
+
+Merge policy:
+
+1. Reject malformed or out-of-bounds spans.
```
andreatnvidia, line 361: One edge case to consider before resolving overlaps here: the winning candidate hasn't necessarily been contextually validated yet. A longer candidate can remove an overlapping fallback, then get dropped by the LLM, leaving neither candidate. Coalescing only exact same-label/span matches before validation, while retaining all their origins, would let final overlap resolution happen after the decisions are known.

plans/262/hybrid-regex-detection.md
```diff
@@ -0,0 +1,660 @@
...
+    reason: str | None = None
+```
+
+This is needed for rules such as URLs, where trailing punctuation may need to
+be removed while preserving exact source offsets.
+
+## Workflow Architecture
+
+The current seed path parses GLiNER output directly into `COL_SEED_ENTITIES`.
+Split candidate generation from seed fan-in:
```
andreatnvidia, line 264: This feels like the right extension seam. It may be worth noting that GLiNER and regex are the first two producers of a future common detector-output contract, with fan-in eventually accepting a configured set of producer columns. The current source-specific implementation can stay, but the note would help keep the next detector from adding another parallel set of constants and merge paths.

andreatnvidia:
The overall direction makes sense, and regex plus GLiNER as independent Data Designer columns feels like a solid first step toward layered detection. I don't think this PR needs to build the full multi-detector framework. Keeping the public regex API and focusing here on correct two-source behavior seems like the right scope.

Could we capture a follow-up issue or design for the general framework? I'd include a shared candidate/outcome shape with multiple origins, one independently schedulable column per detector, generic fan-in, overlap resolution after validation, and per-detector preparation, resource, timeout, and failure contracts. A useful acceptance test would be adding a dummy third detector without changing fan-in, while producing identical results regardless of completion order. That follow-up could also replace the currently missing multi-pole design referenced by this plan.

src/anonymizer/engine/detection/regex_detection.py
lipikaramaswamy: Fixed in f0c37811. Domain labels now allow Unicode combining marks and are NFC-normalized before IDNA validation, while detected values and offsets remain unchanged. Added Devanagari and decomposed-Latin regression cases.

src/anonymizer/engine/detection/regex_detection.py
lipikaramaswamy: Fixed in f0c37811. The IPv6 rule now consumes dotted IPv4 tails before considering the hexadecimal-only form. Added an exact value and offset regression test for ::ffff:192.0.2.128 with IPv4 detection enabled.

src/anonymizer/engine/detection/regex_detection.py
lipikaramaswamy: Fixed in f0c37811. The www prefix check is now case-insensitive, with a regression test for WWW.example.com/path.

skills/anonymizer/BENCHMARK.md
lipikaramaswamy: Fixed in f0c37811. Added the repository standard SPDX headers to BENCHMARK.md and skill-card.md. The full copyright check now passes.

plans/262/hybrid-regex-detection.md
lipikaramaswamy: Implemented in 6bd28e0c. Pre-validation fan-in now coalesces only exact same-label/span duplicates, retains a deterministic origin chain, and preserves partial overlaps and same-span label alternatives as independently identified validation candidates. Final source-aware overlap resolution runs after keep/drop/reclass decisions. Regression coverage confirms that a shorter fallback survives when the longer candidate is dropped, while the longer candidate wins when both survive.

src/anonymizer/engine/detection/custom_columns.py
```diff
@@ -55,16 +60,18 @@
...
 def parse_detected_entities(row: dict[str, Any]) -> dict[str, Any]:
     """Parse detector payload and produce seed entities."""
     text = str(row.get(COL_TEXT, ""))
-    entities = parse_raw_entities(
+    detector_entities = parse_raw_entities(
         raw_response=str(row.get(COL_RAW_DETECTED, "")),
         text=text,
     )
+    regex_entities = _parse_entity_spans(row.get(COL_REGEX_ENTITIES, {}))
+    entities = coalesce_exact_entity_candidates(regex_entities, detector_entities)
```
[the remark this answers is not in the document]
lipikaramaswamy: Fixed in ce66d823 without adding crossing markers to the readable tagged text. Every validation manifest entry now carries its existing context_before and context_after fields into the prompt, and the validator instructions treat that manifest as authoritative when an overlapping candidate cannot be rendered inline. Regression coverage confirms tagged_text stays clean while both overlapping candidates retain independently identifiable context.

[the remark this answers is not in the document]
lipikaramaswamy: Superseded the context-window approach in 6c8a877c to avoid per-candidate token duplication. Validation text now renders each connected overlap region exactly once as a neutral CANDIDATE_GROUP, and the prompt manifest adds only a compact group-id to candidate-id mapping. Non-overlapping entities retain their normal readable tags, and final overlap decisions remain independent. Tests cover nested/crossing spans, overlaps with deterministic accepted candidates, and overlap groups split across validation chunks.

src/anonymizer/config/regex.py
lipikaramaswamy: Fixed in 52cdee3a. RegexRule validation now rejects contextual zero-width matches, including the reported lookahead and lookbehind forms, while allowing lookarounds that consume a non-empty match. Runtime detection now fails explicitly with the rule ID and offset if an unprobed context-dependent zero-width match is encountered instead of silently skipping it. Tests cover both configuration cases, consuming lookarounds, and the runtime backstop.

plans/262/hybrid-regex-detection.md
lipikaramaswamy: Updated in 18f7e9e2. Removed the missing multi-pole design reference so this document stands on its own as the implementation plan for issue #262.

skills/anonymizer/skill-card.md
```diff
@@ -1,139 +1,89 @@
...
-
-Use NeMo Anonymizer through an interactive agent workflow: inspect text data,
-choose Replace or Rewrite, select a replacement strategy, draft a runnable
-Python script, preview before full execution, diagnose failed records first, and
-configure self-hosted GLiNER when detection must stay local.
-
-This skill package is prepared for NVSkills publication review. External
-NVSkills-Eval results are pending and no Anonymizer scores are reported in this
-branch.
+This skill is ready for commercial/non-commercial use. <br>
```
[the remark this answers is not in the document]
lipikaramaswamy: Reverted in 4e928f39. The skill card again describes the package as prepared for NVSkills publication review and explicitly notes that external NVSkills-Eval results are pending, removing the unsupported readiness claim.

src/anonymizer/engine/detection/regex_detection.py
lipikaramaswamy: Fixed in 4e928f39. Direct callable validators now use an opaque object-identity key. Added regression coverage showing that two distinct closures from the same factory register independently and that reusing the same callable reuses its registry identity.

src/anonymizer/config/regex.py
```diff
@@ -0,0 +1,144 @@
...
+    dialect. Runtime detection remains the authoritative backstop for a
+    context-dependent assertion not exercised by these bounded probes.
+    """
+    for probe in ("", pattern, *_ZERO_WIDTH_CONTEXT_PROBES):
+        try:
+            for match in compiled.finditer(probe, timeout=_ZERO_WIDTH_PROBE_TIMEOUT_SECONDS):
+                if match.start() == match.end():
+                    return True
+        except TimeoutError:
+            continue
```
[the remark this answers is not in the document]
lipikaramaswamy: Fixed in 4e928f39. Configuration now parses the regex AST and computes a conservative minimum consumed width for every branch, instead of probing sample strings. This rejects (?<=Z{20}), mixed alternatives such as CASE-[0-9]+|(?=Z{20}), and optional consuming suffixes. Consuming lookarounds remain valid, and the runtime guard remains as defense in depth.

plans/262/hybrid-regex-detection.md
lipikaramaswamy: Corrected in 4e928f39. The plan now states the current graph contract: COL_REGEX_ENTITIES and COL_RAW_DETECTED both depend only on COL_TEXT and are scheduled concurrently; COL_SEED_ENTITIES is the deterministic fan-in barrier.

plans/262/hybrid-regex-detection.md
lipikaramaswamy: Thanks, Andre. We decided to keep this plan focused on the implemented regex and GLiNER graph and removed the forward-looking multi-pole references, including the missing design-doc link. The current plan now documents the concrete two-producer contract: regex and GLiNER run as independent columns, COL_SEED_ENTITIES is their deterministic fan-in barrier, and final overlap resolution occurs after validation.

src/anonymizer/engine/detection/regex_detection.py
```diff
@@ -0,0 +1,433 @@
...
+) -> str | None:
+    if validator is None:
+        return None
+    if isinstance(validator, str):
+        _resolve_validator(validator)
+        return validator
+    module = getattr(validator, "__module__", "")
+    qualified_name = getattr(validator, "__qualname__", "")
+    validator_id = f"local:{module}:{qualified_name}:{id(validator):x}"
+    _LOCAL_VALIDATORS[validator_id] = validator
```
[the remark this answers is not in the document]
lipikaramaswamy: Fixed in 183e97af. Direct callable validators now live in a WeakValueDictionary, so discarding the owning configuration releases the closure and its captured state. Installed entry-point validators use a separate persistent cache. Regression tests verify captured-state collection and that repeatedly creating temporary closure validators does not grow the local registry.

src/anonymizer/engine/detection/regex_detection.py
```diff
@@ -0,0 +1,439 @@
...
+
+@dataclass(frozen=True)
+class RegexDetectionResult:
+    """Regex candidates partitioned by contextual validation route."""
+
+    llm_entities: list[EntitySpan]
+    accepted_entities: list[EntitySpan]
+
+
+_LOCAL_VALIDATORS: WeakValueDictionary[str, RegexValidatorCallable] = WeakValueDictionary()
```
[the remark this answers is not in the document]
lipikaramaswamy: Fixed in b381e46a. Resolved in-process rules now hold their callable validator strongly in an excluded non-serialized field, so deferred execution remains valid after the original RegexRule is collected. Once the resolved workflow is discarded, the weak registry releases the callable and captured state. Tests force GC before execution, verify successful validation, then confirm cleanup afterward; non-weak-referenceable callable objects are covered as well.

src/anonymizer/config/regex.py
andreatnvidia: The reported lookahead and lookbehind cases look good now. There is still a similar gap for `\K`: `RegexRule(label="case_id", pattern=r"CASE\K")` passes validation, then produces a `(4, 4)` span and fails the row at runtime. Could the sequence-width calculation reset when it encounters `core.Keep`, with a test covering this form?

skills/anonymizer/BENCHMARK.md
andreatnvidia: `f0c37811` added the headers, but the regeneration in `49ea1652` removed them again. The current `Check` job is failing on `BENCHMARK.md` and `skill-card.md` as a result. It looks like the generation step needs to preserve the SPDX headers when rewriting these files.

binaryaaron (changes requested):
**agent** (`review-pr`): Request changes — correctness fixes and integration work remain

local human-driven agent review found a few more bits, in addition to the outstanding comments:

1. **Medium: URL validation accepts malformed hosts.**  
   `regex_detection.py:_validate_url()` accepts any IDNA-encodable host containing a dot, including `http://-bad.com` and `http://999.999.999.999/path`, with `validate_with_llm=False`. These false positives reduce detection precision and can cause unnecessary replacement of text. Define the intended hostname policy, validate IP-shaped hosts with `ipaddress`, and validate DNS-style labels before accepting them.

2. **Low: callable serialization succeeds but produces an unusable configuration.**  
   `RegexRule` serializes a callable as `module:qualname`, but restored configurations cannot resolve that representation. Direct callables are documented as in-process only; serialization should fail clearly at that boundary instead of emitting a value that later fails as an unknown validator.

3. **Previously reported: `CASE\K` still permits a zero-width match.**  
   Configuration accepts the pattern, then runtime matching produces `(4, 4)` and drops the row. Width analysis needs to account for `core.Keep`.

4. **Previously reported: generated files still lack SPDX headers.**  
   Signing regeneration removed headers from `BENCHMARK.md` and `skill-card.md`, and the copyright check still fails. Fix the generator before restoring the headers. The restored files will have different hashes, so regenerate and sign `skill.oms.sig` after their final contents are settled.

The PR also conflicts with current `main` across detection code, configuration, tests, interface code, and skill artifacts. This is an integration gate. Conflict resolution must preserve both regex detection and the newer entity-exclusion behavior.

Local validation at `49ea1652`: `1341 passed`; format and type checks passed; copyright check failed on the two generated Markdown files.

## Prompts for managed agents

Agents 1–3 own the fixes below; Agent 4 owns final integration. Each agent should preserve other agents’ changes and report the files changed, validation results, and any unresolved issues.

### Agent 1 — URL validation

Fix built-in URL host validation in `src/anonymizer/engine/detection/regex_detection.py`. Reproduce acceptance of `http://-bad.com` and `http://999.999.999.999/path`, and establish the current underscore-host behavior, using `BuiltinRegex(label="url", validate_with_llm=False)`. Define the intended hostname policy, validate IP-shaped hosts with `ipaddress`, and validate DNS-style labels consistently with IDNA support. Preserve valid Unicode domains, bracketed IPv6 hosts, ports, balanced delimiters, and exact offsets. Add focused positive and negative tests. Keep changes scoped to URL validation.

### Agent 2 — regex configuration boundaries

Fix two configuration issues without enabling arbitrary imports. First, reject patterns such as `r"CASE\K"` during `RegexRule` validation by accounting for `regex._regex_core.Keep` semantics; retain the runtime zero-width backstop. Second, make JSON serialization of direct callable validators fail clearly, preserving the documented rule that only registered entry-point names are portable. Add tests covering `RegexRule` and `Detect` JSON round trips with registered names, serialization failures for top-level callables, and `\K` validation. Preserve validator trust boundaries.

### Agent 3 — generated skill artifacts

Identify why NVSkills regeneration removes repository SPDX headers from `skills/anonymizer/BENCHMARK.md` and `skills/anonymizer/skill-card.md`. Fix the producing template or generation step so future runs retain the headers, then regenerate both files. Renew `skills/anonymizer/skill.oms.sig` only after the final bytes are stable, and run the copyright and skill-package checks. Coordinate with Agent 4 so any artifact changes during integration are followed by regeneration and signing.

### Agent 4 — rebase and integration

Integrate the fixes and rebase PR 265 onto current `main`, resolving conflicts semantically. Preserve both regex detection and the newer entity-exclusion/denylist behavior through `AnonymizerConfig`, interface plumbing, detection workflow columns, postprocessing, and tests. Regenerate signed skill artifacts after source conflicts are resolved, and coordinate final signing with Agent 3. Run `make test`, `make format-check`, `make typecheck`, and `make copyright-check`. Report any remaining conflicts or failed checks without discarding either feature to complete the merge.

lipikaramaswamy:
Addressed review items 1–3 in dd5bf2eb:

- URL validation now distinguishes valid IP literals from invalid IP-shaped hosts and validates every IDNA-normalized DNS label, rejecting leading/trailing hyphens, underscores, empty labels, and malformed numeric addresses. Positive coverage preserves Unicode domains, IPv4, IPv6, and direct deterministic acceptance.
- Regex width validation now handles `\\K` by resetting sequence width and rejects `CASE\\K` at configuration time.
- Direct callable validators remain available in Python-mode configuration but now fail clearly on JSON serialization; registered validator names round-trip through both `RegexRule` and `Detect`.
 
Focused tests and the full 1,352-test suite pass; formatting, lint, and type checks pass.

src/anonymizer/config/regex.py
lipikaramaswamy: Fixed in dd5bf2eb. Sequence width analysis now resets accumulated width when it encounters regex._regex_core.Keep, so CASE\\K is rejected during RegexRule configuration. The runtime zero-width guard remains in place, and the reported form is covered by the configuration regression test.

lipikaramaswamy:
Merge-resolution note on `excluded_entity_labels` + regex detection:

- Exclusions are applied to the effective label set before GLiNER, built-in regex rule selection, and LLM prompts are constructed. Custom regex labels are added to the effective scope first, then exclusions are applied.
- Exclusions are applied again after locally accepted regex entities and LLM-validated entities are merged. This second boundary is a safety net for regex rules using `validate_with_llm=False`, so that route cannot bypass `excluded_entity_labels`.
- The same final filtering remains in place during finalization to protect against reclassification or any later detection route producing an excluded label.

Added integration coverage for excluded custom-regex labels and locally accepted regex matches at both the pre-augmentation and finalization boundaries. The merged suite passes locally (1,467 tests), and the Python 3.11/3.12/3.13 CI jobs pass.

@memadi tagging you for visibility since you authored the excluded-labels change. The merge keeps its original early filtering behavior and extends the safety-net filtering to the new regex routes.

src/anonymizer/engine/detection/regex_detection.py
```diff
@@ -0,0 +1,457 @@
...
+    rules: list[BuiltinRegex | RegexRule],
+) -> list[ResolvedRegexRule]:
+    """Resolve active built-in and custom rules into a serializable form."""
+    active_labels = set(labels)
+    resolved: list[ResolvedRegexRule] = []
+    custom_rules = [rule for rule in rules if isinstance(rule, RegexRule)]
+    builtin_settings = {rule.label: rule for rule in rules if isinstance(rule, BuiltinRegex)}
+    for rule in custom_rules:
+        if not rule.enabled:
+            continue
```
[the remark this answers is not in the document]
lipikaramaswamy: Addressed in dd1a0869 and covered by 94dc7a44. Custom rules are now skipped unless their label is in the active detection scope, so an excluded rule is not compiled, does not resolve or register its validator, and cannot reach LLM validation. The added test uses an intentionally unknown validator to verify that filtering happens before validator resolution.

src/anonymizer/config/anonymizer_config.py
```diff
@@ -111,6 +112,14 @@ class Detect(BaseModel):
             "validator sees per chunk; it is NOT the LLM's context window limit."
         ),
     )
+    builtin_regexes: bool = Field(
+        default=True,
+        description="Run built-in regex recognizers for labels in the effective detection scope.",
+    )
+    regex_rules: list[BuiltinRegex | RegexRule] = Field(
```
binaryaaron: > **agent** (`review-pr`): Invalid custom-rule input silently changes detection policy.

**P1 — Reject malformed custom rules instead of falling back to built-ins**

`Detect.model_validate({"regex_rules": [{"label": "email", "pattern": "("}]})` succeeds as `BuiltinRegex`, discarding the invalid pattern. Direct `RegexRule` construction correctly rejects it. The union can fall back because the built-in model ignores the custom fields; with built-ins disabled, the accepted configuration produces no recognizer for this rule. Please reject extra fields and make the serialized alternatives unambiguous. Add dictionary/JSON tests for invalid patterns and validators alongside direct-constructor tests.

src/anonymizer/engine/detection/postprocess.py
```diff
@@ -329,30 +421,119 @@ def build_tagged_text(
...
-    occurrence of each unique entity value in the text, creating new spans
-    for positions not already covered. Overlaps are resolved by preferring
-    longer spans.
+    Regex spans are not propagated because their pattern or validator may
+    intentionally accept only some occurrences. Other detected values are
+    expanded as before, and overlaps prefer longer spans.
     """
     entity_map: dict[str, str] = {}
-    for entity in entities:
+    propagatable_entities = [entity for entity in entities if not entity_has_source_prefix(entity, "regex_")]
```
binaryaaron: > **agent** (`review-pr`): Adding a regex origin can leave a previously protected occurrence exposed.

**P1 — Preserve GLiNER propagation when exact candidates are coalesced**

For `allow:ABC deny:ABC`, let GLiNER detect the first `ABC` and add `RegexRule(label="token", pattern=r"(?<=allow:)ABC")`. When both candidates survive validation, coalescing produces a source containing both `regex_user:…` and `detector`. This filter then disables propagation for the whole entity. Replacement changes from `allow:[REDACTED] deny:[REDACTED]` on the base to `allow:[REDACTED] deny:ABC`. Please preserve propagation eligibility for the accepted non-regex origin rather than excluding every entity with any regex provenance, and cover the case through final replacement.

src/anonymizer/engine/detection/postprocess.py
```diff
@@ -287,6 +345,40 @@ def resolve_overlaps(entities: list[EntitySpan], *, prefer_highest_score: bool =
...
+            ranked.append((priority, entity))
+
+    ordered = sorted(
+        ranked,
+        key=lambda item: (
+            -(item[1].end_position - item[1].start_position),
+            item[1].start_position,
+            item[1].end_position,
+            item[0],
+            item[1].label,
```
binaryaaron: > **agent** (`review-pr`): The new conflict ordering changes GLiNER-only output.

**P2 — Preserve same-detector confidence ordering with regex disabled**

With regex disabled, let GLiNER return the same `5551234567` span as `phone_number` at `0.95` and `account_number` at `0.4`, and keep both validation decisions. The base selects `phone_number`; this path selects `account_number` alphabetically. That changes label-dependent replacement and annotation without opting into regex. Please retain confidence ordering for comparable candidates from the same detector, while keeping cross-source precedence explicit. Add a regex-disabled pipeline regression test.

src/anonymizer/engine/detection/regex_detection.py
```diff
@@ -0,0 +1,457 @@
...
+                        f"Regex rule {rule.rule_id!r} produced a zero-width match at offset {start}. "
+                        "Regex rules must consume at least one character."
+                    )
+                match_count += 1
+                if match_count > max_matches_per_rule:
+                    raise RuntimeError(
+                        f"Regex rule {rule.rule_id!r} exceeded the maximum of "
+                        f"{max_matches_per_rule} matches for one record."
+                    )
+                if rule.label == "url":
```
binaryaaron: > **agent** (`review-pr`): The output label changes a custom rule's explicitly matched span.

**P2 — Apply URL trimming only to the built-in recognizer**

A custom pattern matching all of `https://example.com/reset?token=abc!` emits a shortened span without `!` when labeled `url`, even with built-ins disabled. The identical pattern labeled `token` preserves the full match. Please attach punctuation trimming to the built-in URL recognizer rather than to the output label, so user-defined match boundaries remain authoritative.

src/anonymizer/config/regex.py
```diff
@@ -0,0 +1,181 @@
...
+            return _minimum_node_width(parsed, _regex_core)
+        except _regex_core._UnscopedFlagSet:
+            global_flags = info.global_flags
+
+
+def _minimum_node_width(node: Any, core: Any) -> int:
+    """Return the minimum characters consumed by a parsed regex node."""
+    if isinstance(node, (core.ZeroWidthBase, core.LookAround)):
+        return 0
+    if isinstance(node, core.Sequence):
```
binaryaaron: > **agent** (`review-pr`): The width analysis accepts a pattern that fails rows with empty matches.

**P2 — Handle nested match resets at pattern admission**

`RegexRule(label="token", pattern=r"CASE(?:\K)")` passes configuration validation, but matching `CASE` yields an empty span and fails the row. The nested sequence's `Keep` effect is lost when its width is added to the enclosing sequence. The runtime guard correctly reports failure, but this contradicts configuration-time rejection of empty matches. Please either reject unsupported reset constructs explicitly or account for their effects compositionally, document the supported dialect, and retain the runtime guard.

binaryaaron (changes requested):
> **agent** (`review-pr`): Configuration and composed detection regressions need fixes before merge.

I recommend fixing the configuration and pipeline defects below before merging. Invalid custom rules can silently become built-in settings, and adding a regex candidate can stop an independently detected GLiNER value from being protected at its other occurrences. The PR also changes same-detector label selection with regex disabled and trims explicitly matched custom URL spans.

The existing DataDesigner producer integration can stay. The architectural changes needed are bounded: make public rule parsing unambiguous, give effective label resolution one owner, and preserve explicit origin and propagation information through deduplication, validation, name splitting, and overlap resolution. Keep competing candidates until their applicable validation decisions are known, then apply one conflict policy.

Before publishing the new API, please also clarify the portable-versus-local validator contract and the supported regex dialect. `RegexValidationResult.reason` and `normalized_value` currently have no execution effect; either define their behavior or omit them from the initial API. Portable descriptors should be validated independently of live callable bindings. Normal local closure and callable-object execution passed; the concern is restoration and the guarantees of the intermediate configuration, not ordinary local execution. Duplicate validator-name handling can be a documented, tracked hardening item.

Validation: 295 detection/configuration tests and 78 replacement/rewrite tests passed. Additional offline probes reproduced the defects below; the September 21 rerun produced the same 12 failing contract assertions and three passing controls. Those assertions include multiple cases for a single issue and a proposed plugin-collision policy, so they are not twelve separate bugs. Real local DataDesigner preview/create probes passed for live validators and surfaced an empty-match row as a `FailedRecord`. No live LLM evaluation was run.

For acceptance, test final protection as well as candidate output: repeated occurrences, exact labels and boundaries, custom-only label scopes, and regex-disabled compatibility. The benchmark approach in #183 is useful here: compare value protection separately from exact span/label parity, and measure model cost only after those checks. This does not require adopting that draft's experimental router.

CI is a separate issue: the latest rollup still has the NVSkills evidence gate failing and DCO requiring action; Python 3.11/3.12/3.13 checks pass. The pre-existing augmentation tie-break behavior is not counted as a new general regression in this review.

### 3. P2 — Carry regex occurrence restrictions through name splitting

[src/anonymizer/engine/detection/postprocess.py:276](https://github.com/NVIDIA-NeMo/Anonymizer/blob/94dc7a445a14cf46f8705bc7980d2cc27d4ed377/src/anonymizer/engine/detection/postprocess.py#L276)

> **agent** (`review-pr`): Name splitting bypasses the new regex occurrence restriction.

A regex-only `full_name` match restricted to `allow:John Smith` in `allow:John Smith deny:John Smith and John` still produces `name_split` entities in the denied occurrence and at the later standalone `John`, with an empty augmentation response. These survive under the default permissive label scope. Name splitting predates this PR, but it now needs to preserve the occurrence restrictions of regex-derived candidates. Please restrict derived spans to eligible occurrences and test this together with mixed-origin propagation.

### 6. P2 — Validate the same effective label scope that execution uses

[src/anonymizer/config/anonymizer_config.py:191](https://github.com/NVIDIA-NeMo/Anonymizer/blob/94dc7a445a14cf46f8705bc7980d2cc27d4ed377/src/anonymizer/config/anonymizer_config.py#L191)

> **agent** (`review-pr`): Configuration rejects a nonempty custom-only detection scope.

With `entity_labels=None`, excluding every default label and adding a custom `ticket` rule raises an empty-effective-set error. The workflow resolver adds custom regex labels before exclusions and would retain `ticket`. Please use one effective-scope resolver for configuration validation and execution, preserving the distinction between an explicit allowlist and permissive default augmentation.

src/anonymizer/engine/detection/regex_detection.py
```diff
@@ -0,0 +1,457 @@
...
+    except ValueError:
+        pass
+    else:
+        return True
+
+    if ":" in host or _IPV4_SHAPED_HOST_RE.fullmatch(host) is not None:
+        return False
+    if len(host) > 253 or "." not in host:
+        return False
+    return all(_DNS_LABEL_RE.fullmatch(label) is not None for label in host.split("."))
```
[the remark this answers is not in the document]
binaryaaron: > **agent** (`review-pr`): The root-dotted URL issue still reproduces at the reviewed head.

`https://example.com./private/alice` is rejected while the equivalent URL without the terminal DNS root dot is detected. Please allow the terminal root dot during hostname validation while preserving the original matched text and offsets. Add the paired URLs as a regression case.

src/anonymizer/config/anonymizer_config.py
lipikaramaswamy: Addressed in 01fa5b6d. `BuiltinRegex` and `RegexRule` now both use `extra="forbid"`, so malformed custom-rule dictionaries cannot fall back to `BuiltinRegex` by discarding `pattern` or `validator`. Added dictionary and JSON regressions for invalid patterns and validators, direct unknown-field tests for both alternatives, and retained valid mixed-rule JSON round trips. The 111 relevant configuration, workflow-serialization, and regex tests pass in a branch-local environment.

src/anonymizer/config/anonymizer_config.py
```diff
@@ -143,13 +204,18 @@ def validate_entity_label_overlap(self) -> "Detect":
...
+            self.entity_labels,
+            regex_rules=self.regex_rules,
+            excluded_entity_labels=excluded_set,
+        )
 
         if self.entity_labels is not None:
             entity_labels_set = set(self.entity_labels)
             overlap = sorted(entity_labels_set & excluded_set)
             if not overlap:
                 return self
```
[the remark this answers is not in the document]
lipikaramaswamy: Fixed in 5eb43ac1. Explicit-scope overlap detection now uses the same strip + Unicode casefold normalization as effective-label resolution, so equivalent labels such as `straße` and `STRASSE` cannot bypass the empty-scope guard. The shared normalization helper is used by both paths, and a focused configuration test covers the reported case. The 117 relevant config/workflow tests pass, along with the repository pre-commit checks.

binaryaaron (changes requested):
> **agent** (`review-pr`): Close the detection-policy gaps through final output before adding further routing behavior.

This review is pinned to `929d1784`; the branch has advanced since that verification. It does not assess subsequent changes.

# PR #265 — implementation review

For the author and implementation agents. Reviewed `929d1784bff551735943dc0cc2eeed5f61638513`, compared with the previously reviewed `94dc7a445a14cf46f8705bc7980d2cc27d4ed377`. Findings below apply to that revision. References to the PR merge base mean `86ec54d8075245fa44c7a23582605766fef568c4`, not the previously reviewed head.

---

The configuration fixes are verified, but the PR still has two paths that lose protection: mixed regex/GLiNER provenance suppresses propagation, and the new all-regex route silently loses accepted matches during DataDesigner execution. Please prioritize these before extending the feature further.

The recurring issue is that policy is enforced at one stage and then reinterpreted or bypassed at the next. A shared label resolver helps, but activation still uses a different equality rule. Removing labels from model prompts helps, but validation and name splitting can reintroduce them. Skipping a model column helps avoid generation, but its skip state also disables the transforms needed to retain regex results.

Please organize the remaining changes around the contracts below. The required outcomes are explicit; the suggested internal shapes are implementation guidance. Keep the existing `Detect` entry point and `NddAdapter` execution boundary. A generic detector framework is unnecessary for these fixes.

Validation at `929d1784`: 399 targeted repository tests passed; separate reproductions confirmed the defects below, including real DataDesigner preview/create execution. No live LLM inference was used.

## What is already fixed

- Both rule models forbid extra fields. Invalid custom-rule dictionaries and JSON no longer fall back to a built-in configuration by discarding the pattern or validator.
- Configuration validation and workflow construction share effective-label resolution. Excluding all defaults while retaining an enabled custom rule now works.
- Disabled custom rules no longer extend the effective scope or trigger the enabled-rule duplicate/scope checks.
- The specific Unicode exclusion example, `entity_labels=['straße']` with `excluded_entity_labels=['STRASSE']`, now raises the expected empty-scope error.

Retain these fixes and their tests. The Unicode activation defect below is a separate inconsistency introduced by combining normalized scope deduplication with exact activation comparisons.

## 1. Preserve accepted regex matches across skipped model columns — P1

**Owner:** detection graph construction in [detection_workflow.py](https://github.com/NVIDIA-NeMo/Anonymizer/blob/929d1784bff551735943dc0cc2eeed5f61638513/src/anonymizer/engine/detection/detection_workflow.py#L238), particularly seed parsing and merge inputs.

**Reproduction:** use explicit `entity_labels=['ticket']`, disable built-ins, and configure `RegexRule(label='ticket', pattern=r'TKT-\d+', regex_only=True)`. Disable model health checks (`skip_health_check=True`) and guard HTTP transports to isolate row execution, then process `TKT-123` through the real adapter. In both preview and full creation, the regex producer emits the correct span, but `_detected_entities` becomes `None`, final entities are empty, and no failed record is reported.

**Cause:** DataDesigner columns inherit `propagate_skip=True`. The unconditional detector skip propagates into seed parsing; the augmenter skip independently propagates into merging. Supplying an empty skip value does not prevent this propagation.

**Required invariant:** intentionally skipping an optional detector must not skip the domain operations that combine other detectors' accepted output.

**Bounded fix:** make the two fan-in transforms consume intentional empty model inputs without inheriting their skip state, or provide those empty inputs through local producers. Setting `propagate_skip=False` on seed parsing and merging restored the final match in a test-only control. Limit this change to transforms that can safely consume the configured empty inputs.

**Acceptance:** execute preview and create through `NddAdapter`, with one matching row and one nonmatching row. Assert the exact final span, empty results only for the nonmatching row, and no silent loss of a successful producer's output. Exercise replacement on the matching result. A test asserting only that `skip` is configured does not cover this defect.

**Separate runtime contract:** model health checks still run before row skips. The reproduction isolated row execution by disabling health checks and guarding HTTP transports. With default health checks, the same graph attempts remote model calls. Document the distinction between skipping row generation and running without a model service; if the latter is intended, remove unused model dependencies. Disabling health checks alone does not fix the lost-match defect.

## 2. Preserve provenance and occurrence permissions through every transform — P1

**Owners:** exact coalescing, occurrence expansion, and name splitting in [postprocess.py](https://github.com/NVIDIA-NeMo/Anonymizer/blob/929d1784bff551735943dc0cc2eeed5f61638513/src/anonymizer/engine/detection/postprocess.py).

For `allow:ABC deny:ABC`, let GLiNER detect the first `ABC` and add a contextual regex matching only that first occurrence. After both candidates are accepted, their merged source contains both origins. The current “has any regex source” check suppresses GLiNER propagation. Actual replacement leaves the second `ABC` exposed; the PR merge-base implementation protects both.

The opposite error also remains: a regex-origin `full_name` restricted to `allow:John Smith` can generate name parts in `deny:John Smith` and at a later standalone `John`. Here “regex-origin” means that regex supplied the candidate; it does not imply the newly added `regex_only=True` configuration flag. Fixing the explicit flag's full-name case does not repair the existing default-mode behavior.

**Required invariants:**

- Adding a redundant accepted regex origin must not remove propagation permission supplied by an independently accepted detector origin.
- An occurrence restriction must survive derived-span creation. Name splitting cannot grant a regex-origin candidate permission to detect arbitrary new occurrences.
- Provenance must survive validation and coalescing sufficiently to explain why a final span was admitted.

**Suggested architecture:** distinguish origin, admission route, and occurrence permission internally. A joined `source` string is suitable for display but is insufficient as their only representation. A small internal record or explicit helper contract is enough; preserve the existing external source representation where compatibility requires it.

**Acceptance:** mixed-origin replacement protects both `ABC` occurrences; a contextual regex without an independent propagatable origin protects only its accepted occurrence; derived name parts respect that scope. Test both LLM-validated and locally accepted regex routes. Do not make users opt into a new mode to recover the old GLiNER behavior.

## 3. Use one label identity throughout configuration and execution — P2

**Owners:** [effective scope](https://github.com/NVIDIA-NeMo/Anonymizer/blob/929d1784bff551735943dc0cc2eeed5f61638513/src/anonymizer/config/anonymizer_config.py#L52), rule activation, and regex-only policy grouping in `regex_detection.py` and `Detect`.

With implicit labels, enabled `straße`/`ABC` and `strasse`/`DEF` custom rules now detect only `ABC` from `ABC DEF`. Scope resolution casefolds and retains one spelling; activation compares exact strings and drops the other rule. Both patterns executed at the previous reviewed head.

**Required invariant:** every enabled rule is either activated according to the declared scope or rejected with a clear configuration error. Normalization must not silently remove an enabled pattern.

Use the same canonical identity for scope, activation, exclusions, built-in overrides, and per-label policy conflicts. Keep display spelling separate if necessary. Alternatively, reject ambiguous equivalent spellings explicitly. Apply the decision consistently to Python and serialized input.

**Acceptance:** test both rule orders, explicit and implicit scopes, exclusion of the equivalent label, and conflicting `regex_only` values on equivalent spellings. Both patterns must execute under an accepted shared identity, or construction must clearly reject the ambiguity. Merely fixing the original exclusion example is insufficient.

## 4. Enforce regex-only authority after reclassification and derivation — P2

**Owners:** validation application in `custom_columns.py` and name-part generation in [postprocess.py](https://github.com/NVIDIA-NeMo/Anonymizer/blob/929d1784bff551735943dc0cc2eeed5f61638513/src/anonymizer/engine/detection/postprocess.py#L301).

Two routes currently bypass the new policy:

- Configure regex-only `ticket` matching `TKT-\d+` plus model-enabled `identifier`. A validator can reclassify GLiNER's `identifier` candidate for `OTHER-123` into `ticket`; it survives finalization without a regex match.
- Configure regex-only `first_name` matching `Alice` plus model-enabled `full_name`. Detecting `John Smith` in `John Smith met John` creates a final `first_name` entity for the trailing `John`. The name-splitting guard checks the parent label, not the emitted part label.

**Required invariant:** final candidates under a regex-only label require eligible regex evidence. Removing the label from a prompt is not enforcement of that invariant.

Carry the policy through both validation-application stages and derived-span creation. Preserve accepted regex candidates while excluding model-only introductions. Decide how a reclassification into a forbidden route is handled, and test that decision; do not silently treat it as an authorized detection.

**Acceptance:** both examples above must stop producing unauthorized regex-only entities. Include positive controls proving that genuine regex matches survive and unrelated model-enabled labels still work. Test the full-name parent and each generated name-part label separately.

## 5. Give overlap resolution one explicit policy — P2

**Owner:** `merge_entity_sources`, `resolve_overlaps`, and their callers.

With regex disabled, same-span GLiNER candidates `phone_number` at `0.95` and `account_number` at `0.4`, both retained by validation, now resolve to `account_number` alphabetically. The PR merge base selects `phone_number`. This changes label-dependent replacement without enabling regex.

Retain alternatives until their applicable validation decisions are known. Then apply an explicit policy that preserves confidence ordering within a comparable detector origin and defines cross-source precedence separately. Regex's synthetic score of `1.0` should not be interpreted as a confidence comparable to GLiNER's score.

Audit the earlier augmentation overlap resolver as part of this ownership change: a later merge cannot restore a candidate already discarded upstream. The analogous augmentation limitation predates this PR; it is not a newly discovered general regression.

**Acceptance:** regex-disabled output preserves the higher-confidence label; a rejected higher-priority candidate allows an eligible alternative to survive; rule ordering does not alter the agreed outcome. Assert labels and replacement behavior, not only entity counts.

## 6. Fix remaining matcher defects — P2

These cases are reproduced output or pattern-admission defects:

| Boundary | Observed problem | Required outcome |
| --- | --- | --- |
| Custom span ownership | A custom rule matching `https://example.com/reset?token=abc!` loses `!` solely because its label is `url`, even with built-ins disabled. | Attach punctuation trimming to the built-in recognizer. Preserve explicit custom match boundaries. |
| Built-in hostname validation | `https://example.com./private/alice` still fails while its non-root-dotted equivalent passes. | Accept the terminal DNS root dot during hostname validation without changing source text or offsets; keep malformed empty-label hosts invalid. |
| Pattern admission | `CASE(?:\K)` passes configuration checks and then fails a row with an empty match. | Define the supported dialect and reject unsupported constructs clearly, or make the width adapter account for nested effects. Retain runtime timeout, match-count, and empty-span guards. |

For pattern analysis, keep use of private `regex` parser internals isolated and version-tested. The current partial width walk is not sufficient evidence that all admitted patterns consume text. A finite supported subset is acceptable if its restrictions are intentional and documented.

## 7. Settle public-validator and transport contracts

These API-contract concerns need explicit decisions before publication. The internal separation below is a design recommendation; it is not a claim that supported in-process execution fails.

| Boundary | Observed limitation | Recommended contract |
| --- | --- | --- |
| Validator result | `reason` and `normalized_value` are public fields but execution consumes only `valid`. | Define observable semantics or omit unused fields from the initial API. Never substitute normalized text without an offset-preserving contract. |
| Portable configuration | Resolved rules admit unsupported source strings and serialized local IDs whose restoration depends on process/object lifetime. | Separate validated portable descriptions from executable bindings; reject invalid restored state before row processing. Keep local callable ownership explicit. |

Normal local closure and callable-object execution passed. The serialization concern does not mean those supported in-process paths are broken. Account for DataDesigner's own artifact serialization when changing this boundary; an unconditional serializer exception can break normal execution.


Duplicate installed-validator names currently select the first discovered implementation. Deterministic ambiguity handling is recommended hardening; it can be tracked separately if the initial policy is documented. It is not the reason for the blocking recommendation.

## Suggested implementation sequence

1. Fix the two loss-of-protection paths: skip propagation and mixed-origin occurrence propagation. Add final-output regressions before broader restructuring.
2. Unify label identity and enforce regex-only admission through reclassification and name splitting. These share the per-label policy boundary and should use the same canonical identity.
3. Consolidate overlap ordering while preserving regex-disabled behavior and candidate fallback after validation.
4. Close matcher and transport contracts, then update public documentation and the bundled skill to describe the implemented behavior.

Each change should identify the affected invariant, include a failing-before/passing-after reproduction, and retain unrelated behavior. Avoid adding another public routing switch to work around an internal policy inconsistency.

PR #268, opened 2026-09-15
fix(detection): normalize null proposed labels

lipikaramaswamy:
## Summary

- normalize an explicit `proposed_label: null` from model output to the existing empty-string default
- keep the downstream `RawValidationDecisionSchema.proposed_label` contract typed as `str`
- add focused regression coverage for the null response

## Testing

- `uv run pytest tests/engine/test_schemas.py -q` (44 passed)
- repository pre-commit checks passed

## Contribution checks

- No linked issue: maintainer-owned, narrowly scoped bug fix identified while diagnosing benchmark model-output drift.
- No plan required: this is a two-file internal schema normalization with no public API or data-flow changes.
- Public API impact: none.
- PII fixtures: synthetic only.
- Secrets: none.


memadi-nv (approved):
LGTM. Thanks for the fix!

PR #269, opened 2026-09-15
fix(rewrite): normalize null evaluator fields

lipikaramaswamy:
## Summary

- normalize explicit null quality answers to the existing conservative unknown fallback
- normalize explicit null privacy fields to the existing highest-confidence leak fallback
- normalize explicit null comparison scores to the existing conservative zero-score fallback
- preserve strict validation for IDs, enum values, score bounds, and answer coverage

## Testing

- uv run pytest tests/engine/test_schemas.py tests/engine/test_evaluate.py -q (77 passed)
- make format-check
- make typecheck
- repository pre-commit checks passed

## Contribution checks

- No linked issue: narrowly scoped follow-up to null model-output drift handled in #268.
- Public API impact: none.
- PII fixtures: none.
- Secrets: none.


asteier2026 (approved):
Looks good to me

src/anonymizer/engine/schemas/rewrite.py
```diff
@@ -383,6 +389,30 @@ class PrivacyAnswerItemSchema(BaseModel):
     reason: str = Field(min_length=1, max_length=200)
     evidence: list[str] = Field(default_factory=list)
 
+    @field_validator("answer", mode="before")
+    @classmethod
+    def normalize_null_answer(cls, value: object) -> object:
+        """Default an explicit JSON null to the conservative leak verdict."""
+        return PrivacyAnswer.yes if value is None else value
```
memadi-nv: The privacy fields are normalized independently, which can undermine the intended fail-closed behavior. For example:

`{"answer": null, "confidence": 0.0}`
becomes `answer="yes"` with `confidence=0.0`. Because leakage mass is multiplied by confidence, this contributes zero leakage rather than representing a highest-confidence leak.

The fallback `reason` can also become inaccurate when only reason is null—for example, an `answer="no"` result could receive a reason claiming it was defaulted to a highest-confidence leak.

**Proposed solution:** normalize privacy answers atomically. When `answer` is null, set a`nswer="yes", confidence=1.0,` and the fallback reason together. When only `reason` is null, use neutral text that does not claim the answer or confidence was changed. Please also add tests covering individual and partial-null combinations.

memadi-nv (changes requested):

src/anonymizer/engine/schemas/rewrite.py
lipikaramaswamy: Addressed in 08275df4. Privacy fallback normalization is now atomic: when answer is null, the schema sets answer=yes, confidence=1.0, and the fallback reason together, overriding conflicting partial values. A standalone null reason now gets neutral text without claiming the verdict changed. I also added individual and partial-null regression cases for answer, confidence, reason, and evidence.

Validation: 81 focused tests passed; format/lint, typecheck, and commit hooks all passed.

memadi-nv (approved):

PR #212, opened 2026-07-06
feat(dev): add local model serving tool

binaryaaron:
## Summary

- Add a source-tree developer tool that compiles and manages local vLLM inference services from pinned TOML profiles.
- Support stock Hugging Face generation models and NVIDIA GLiNER or GLiNER2 through the pinned external [vLLM Factory](https://github.com/latenceainew/vllm-factory) integration.
- Preserve pure compilation, digest-protected plans, task-aware capability probes, PID-reuse protection, and bounded process-group cleanup.
- Preserve vLLM API-key authentication on the Factory detector compatibility route.
- Ship seven profiles: NVIDIA GLiNER, GLiNER2, TinyLlama, GPT-OSS 20B and 120B, Qwen3 30B A3B Instruct, and NVIDIA Nemotron 3.5 Lightning.
- Add deployment guidance and a checked-in NVIDIA GPU container image that runs the same lifecycle inside the container boundary.

## Scope and lifecycle

The tool owns one deployment domain: a managed local-process vLLM server.

- The v2 input schema has four sections: `task`, `model`, `vllm`, and `local`.
- Lifecycle commands are `compile`, `launch`, `probe`, `status`, and `stop`.
- Factory-backed detection and stock vLLM generation share one compiler and lifecycle.
- The Factory detector route enforces `VLLM_API_KEY` before request parsing or inference and returns `401` for missing, malformed, or incorrect credentials.
- The supplied container runs that unchanged local-process lifecycle. It does not add a second compiler path.
- Docker-specific compiler branches, native Transformers GLiNER serving, and Hugging Face cache discovery remain outside the design because no shipped profile requires them.
- Plan SHA-256 digests detect accidental storage or transport changes. They do not authenticate plan writers or prove semantic validity.

The tool remains under `tools/` and is not part of the distributable Anonymizer package. It does not attach to existing endpoints or manage remote compute.

## Validation

Validation on current HEAD `8562bdd`:

- Focused inference-service and vLLM Factory tests: 58 passed and 3 expected skips.
- `make format-check`: passed.
- `make typecheck`: passed.
- `make test`: 1,341 passed and 3 expected skips.

Earlier full-branch validation on rebased HEAD `836a13f`:

- `make check`: passed, including formatting, lint, `ty`, lock consistency, and SPDX checks.
- `make docs-build`: strict build passed.
- Range-diff matched all 17 product commits exactly after rebasing onto current `main`.
- Every rebased product commit has a DCO trailer and a valid repository ED25519 signature.

[...]

PR #271, opened 2026-09-15
feat: add notebook-local GLiNER2 runtime

lipikaramaswamy:
## Related Issue

No linked issue required: maintainer-owned recovery of the notebook path after the NVIDIA Build GLiNER detector became unavailable.

## Plan Document

[`plans/notebook-local-gliner/notebook-demo-runtime.md`](https://github.com/NVIDIA-NeMo/Anonymizer/blob/lipikaramaswamy/feature/notebook-local-gliner2/plans/notebook-local-gliner/notebook-demo-runtime.md)

## Summary

- Add a notebook-owned, OpenAI-compatible GLiNER2 runtime using the pinned `fastino/gliner2-privacy-filter-PII-multi` revision.
- Start and stop the local detector through `anonymizer.notebooks.create_anonymizer()` and `stop_local_runtime()`, with automatic CUDA, MPS, or CPU selection.
- Isolate the GLiNER2 server dependencies from the notebook environment to accommodate its older Hugging Face Hub requirement.
- Route notebook detection to loopback by default while preserving configurable external LLM providers for augmentation, validation, replacement, rewrite, and evaluation.
- Update tutorials, generated notebooks, documentation, tests, packaging, and notebook CI for the local runtime.
- Keep production guidance focused on a separately managed compatible endpoint. This PR is intended to land before #212; #212 can become the canonical production vLLM Factory guidance when it lands.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update
- [x] Refactoring
- [x] CI, release, or contributor workflow update

## Contributor Checklist

- [x] PR title follows Conventional Commits, for example `fix: handle empty entity list`
- [x] Related issue is linked, or a maintainer-owned no-issue reason is documented above
- [x] For non-trivial changes, a plan document is linked above, or the no-plan reason is documented above
- [x] Public API impact checked; `skills/anonymizer/SKILL.md` does not require an update because its generated workflow continues to use the core `Anonymizer` API and does not use the notebook-only helper
- [x] No real PII added to tests, docs, notebooks, fixtures, or artifacts
- [x] No API keys, service tokens, private keys, credentials, or real endpoint secrets added

## Validation

- Commands and checks run:
  - Focused notebook/runtime and affected workflow tests: 168 passed
  - Full test suite: 1307 passed; two host-specific failures noted below
  - `make format-check`
  - `make typecheck`
  - `make copyright-check`
  - `make lock-check`
  - `make docs-build`
  - Built the wheel and verified a clean installation
  - Loaded the exact pinned GLiNER2 revision on CPU and verified real inference and exact entity offsets
  - Executed all five notebooks end-to-end with a compatible OpenAI endpoint: 73/73 code cells completed with zero notebook errors and no credential material embedded in outputs
  - The full suite had two environment-specific host failures: an AF_UNIX path-length limit in the isolated worktree and a W&B sandbox log-path permission failure. Neither failure exercises the changed runtime behavior.

- [x] Docs updated, or not needed
- [x] If tutorial sources changed: notebooks regenerated with `make convert-notebooks`
- [x] If e2e, benchmark, or model-provider behavior changed: relevant validation is listed above


src/anonymizer/interface/anonymizer.py
```diff
@@ -827,6 +829,41 @@ def _validate_preflight_config(self, config: AnonymizerConfig) -> None:
...
+        detector_alias = self._selected_models.detection.entity_detector
+        detector_config = next(
+            (config for config in self._model_configs if config.alias == detector_alias),
+            None,
+        )
+        if detector_config is None or detector_config.provider not in {"local-gliner2", "notebook-local-gliner2"}:
+            return
+        provider = next(
+            (provider for provider in self._resolved_providers if provider.name == detector_config.provider),
+            None,
```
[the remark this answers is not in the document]
lipikaramaswamy: Fixed in b2f88bbe. Anonymizer now records whether it manages the DataDesigner instance and skips the local-detector endpoint preflight when a caller-supplied DataDesigner owns provider registration. Added a regression test asserting that no bundled-provider HTTP probe occurs for a supplied instance.

src/anonymizer/notebooks/_runtime.py
```diff
@@ -0,0 +1,333 @@
...
+
+
+def _start_runtime(requested_device: str) -> _LocalRuntime:
+    global _previous_token
+    server_packages = _ensure_server_environment()
+    port = _find_available_port()
+    token = secrets.token_urlsafe(32)
+    if _previous_token is None:
+        _previous_token = os.environ.get(LOCAL_TOKEN_ENV)
+    os.environ[LOCAL_TOKEN_ENV] = token
```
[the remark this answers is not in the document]
lipikaramaswamy: Fixed in b2f88bbe. Token restoration now tracks capture state separately from the original value, preserving the distinction between an originally absent variable and an uncaptured one across startup retries. Added a regression test covering repeated generated-token assignment followed by restoration to absence.

src/anonymizer/notebooks/_runtime.py
```diff
@@ -0,0 +1,342 @@
...
+            requested_device,
+        )
+        stop_local_runtime()
+    elif _runtime is not None:
+        _runtime = None
+
+    last_error: Exception | None = None
+    for _ in range(_MAX_PORT_ATTEMPTS):
+        runtime = _start_runtime(requested_device)
+        try:
```
andreatnvidia, line 123: Stopping a slow first launch raises `KeyboardInterrupt`, so this `except Exception` block is skipped. Since `_runtime` is assigned only after readiness succeeds, the child can remain alive and hold the accelerator while normal cleanup has no reference to it. Could the interrupt path terminate the candidate and restore the token before propagating?

src/anonymizer/interface/anonymizer.py
```diff
@@ -827,6 +830,41 @@ def _validate_preflight_config(self, config: AnonymizerConfig) -> None:
...
 
+    def _validate_local_detector_endpoint(self) -> None:
+        if not self._uses_default_detection_workflow or not self._manages_data_designer:
+            return
+        detector_alias = self._selected_models.detection.entity_detector
+        detector_config = next(
+            (config for config in self._model_configs if config.alias == detector_alias),
+            None,
+        )
+        if detector_config is None or detector_config.provider not in {"local-gliner2", "notebook-local-gliner2"}:
```
andreatnvidia: The self-hosting example names its provider `my-gliner2-service`, but that name never reaches this preflight. A bad or mismatched endpoint then fails later inside Data Designer instead of producing the actionable GLiNER2 error, and servers that ignore the requested model may run the wrong detector unnoticed. Basing the guard on the selected detector model or contract would cover custom provider names.

tools/serve_gliner.py
```diff
@@ -1,486 +1,10 @@
...
-    )
-    parser.add_argument(
-        "--port",
-        type=int,
-        default=DEFAULT_PORT,
-        help=f"Listen port (default: {DEFAULT_PORT})",
-    )
-    return parser.parse_args()
-
+from anonymizer.notebooks._runtime import run_development_server
```
andreatnvidia, line 7: `tools/serve_gliner.py` is still the command in the B300 archival instructions, including the old device and batching environment variables and the pinned `nvidia/gliner-pii` model. After this change, the same command ignores those knobs and serves Fastino GLiNER2, so the documented benchmark cannot be reproduced faithfully. Please either preserve a legacy entry point or pin the devnote to the older implementation.

src/anonymizer/notebooks/local_inference/gliner2/backend.py
```diff
@@ -0,0 +1,261 @@
...
+        enumerate(entities),
+        key=lambda row: (
+            -float(row[1]["score"]),
+            int(row[1]["start"]),
+            int(row[1]["end"]),
+            row[0],
+        ),
+    )
+    distinct: list[tuple[int, dict[str, Any]]] = []
+    seen_boundaries: set[tuple[int, int]] = set()
```
andreatnvidia: Two predictions can cover the same characters while carrying different labels, but `seen_boundaries` keeps only one of them here. The previous server retained both labels, and dropping one can alter the replacement category selected downstream. Keying this set by span and label, with a matching regression case, would avoid that loss.

src/anonymizer/notebooks/local_inference/gliner2/backend.py
```diff
@@ -0,0 +1,261 @@
...
+    ends = [int(entity["end"]) for _, entity in by_end]
+    predecessors = [
+        bisect.bisect_right(ends, int(entity["start"]), 0, index) - 1 for index, (_, entity) in enumerate(by_end)
+    ]
+    best: list[tuple[float, tuple[int, ...]]] = [(0.0, ())]
+    for index, (_, entity) in enumerate(by_end):
+        previous_score, previous_selection = best[predecessors[index] + 1]
+        with_entity = (previous_score + float(entity["score"]), previous_selection + (index,))
+        without_entity = best[index]
+        if with_entity[0] > without_entity[0] or (
```
andreatnvidia: On a tied total score and span count, this branch retains the earlier-ending selection rather than GLiNER2's confidence/start/end ordering. With equal-score outer `[2,10]` and inner `[5,8]` candidates, it keeps the inner span, which can leave part of a sensitive value uncovered. Adding the upstream tie-break and an equal-score overlap test would align the post-chunk resolver with the pinned model.

andreatnvidia (changes requested):
Requesting changes for the blocking issues noted inline. I also left two non-blocking comments on overlap handling.

src/anonymizer/notebooks/local_inference/gliner2/backend.py
```diff
@@ -0,0 +1,261 @@
...
+                raise
+        return target
+    finally:
+        shutil.rmtree(temporary_root, ignore_errors=True)
+
+
+def _link_or_copy(source: str, destination: str) -> str:
+    """Hard-link cached model artifacts when possible, copying as a fallback."""
+    try:
+        os.link(source, destination)
```
andreatnvidia: The pinned Hugging Face snapshot stores its files as relative symlinks into the blob cache. On Linux, this `os.link()` call preserves those symlinks under `model-adapters`, where their relative targets no longer resolve; a real GB300 cold start fails at `config_path.read_text()` before the model loads. Please resolve symlink sources before hard-linking, or copy while dereferencing them, and add a symlink-backed snapshot test.

lipikaramaswamy: Fixed in 25aabc7e. Snapshot sources are resolved to their real blob paths before hard-linking or copying, so the adapter cache contains usable regular files rather than relocated relative symlinks. Added a Hugging Face-style relative-symlink snapshot regression test.

src/anonymizer/notebooks/local_inference/gliner2/backend.py
lipikaramaswamy: Fixed in 25aabc7e. Document-level deduplication now keys predictions by span and label. For the disallow-overlap policy, same-span labels are grouped for interval selection and all labels on the winning span are retained. Regression coverage exercises both overlap policies.

src/anonymizer/notebooks/_runtime.py
lipikaramaswamy: Fixed in 25aabc7e. The startup wait now handles KeyboardInterrupt and SystemExit explicitly, terminates the candidate child, restores the original token environment, and then re-raises. Added a regression test that verifies termination, token restoration, and that no managed runtime is retained.

src/anonymizer/interface/anonymizer.py
lipikaramaswamy: Fixed in 25aabc7e. The preflight guard is now based on the selected Fastino GLiNER2 detector model rather than a provider-name allowlist, so custom provider names such as my-gliner2-service receive the same model and endpoint validation. Added a custom-provider regression test.

tools/serve_gliner.py
lipikaramaswamy: Fixed in 25aabc7e. The B300 devnote now downloads and runs the legacy server from the exact pre-change commit a2cc91847d2674649f1a30a518851e25773010fa, preserving its old environment knobs and NVIDIA GLiNER PII contract. The note also explicitly distinguishes that archived server from the current GLiNER2 entry point.

src/anonymizer/notebooks/local_inference/gliner2/backend.py
lipikaramaswamy: Fixed in 25aabc7e. Equal total-score and span-count selections now use the established confidence, start, end, and original-order ranking as the final tie-break. Added the equal-score outer 2-to-10 versus inner 5-to-8 regression case, which now selects the outer span.

src/anonymizer/interface/anonymizer.py
```diff
@@ -827,6 +832,41 @@ def _validate_preflight_config(self, config: AnonymizerConfig) -> None:
...
+        api_key = os.getenv(provider.api_key, provider.api_key) if provider.api_key else None
+        headers = {"Authorization": f"Bearer {api_key}"} if api_key else None
+        try:
+            response = httpx.get(f"{provider.endpoint.rstrip('/')}/models", headers=headers, timeout=2.0)
+            response.raise_for_status()
+            payload = response.json()
+            if not isinstance(payload, dict):
+                raise ValueError("endpoint returned malformed model metadata")
+            models = payload.get("data", [])
+            if not any(isinstance(model, dict) and model.get("id") == detector_config.model for model in models):
```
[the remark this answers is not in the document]
lipikaramaswamy: Clarified in 2f7b0d52. The approved runtime plan intentionally includes GET /v1/models for readiness and model discovery, and Andrea specifically requested that custom providers be checked so a server cannot silently ignore the configured detector. The public Server contract had omitted that requirement; it now documents the two-second discovery call, exact configured model ID, response shape, and rationale. Managed endpoints that do not implement discovery are therefore not compatible with this detector contract.

tools/serve_gliner.py
lipikaramaswamy: Follow-up in 55bda9b6: cleaned up the archival guidance. Instead of downloading a script into /tmp, the devnote links the historical implementation and tells repro users to run the original command from an Anonymizer checkout at that commit. It also distinguishes the source pin from the separate Hugging Face model-weight revision.

andreatnvidia:
After the latest fixes, I reran the committed `docs/notebooks/01_your_first_anonymization.ipynb` end to end at `12527a5` on a GB300, using the notebook-managed local GLiNER2 runtime and `google/gemma-4-31b-it`. All 13 code cells completed in 2m54s with no cell errors. The full run processed 25 records, detected 698 entities, completed substitutions with zero failed records, and finished the 3-record evaluation. The model endpoint handled 102 chat requests without a non-200 response, averaging 8.9s per request. This closes the notebook runtime concern for me.

andreatnvidia (approved):

PR #275, opened 2026-09-17
add plan for entity label examples

### LINKED ISSUE #259: Expose per-label detection examples in the public configuration API
### Priority Level

Medium (Nice to have)

Entity-label examples are an important quality lever for datasets containing domain-specific formats, such as vendor API keys, account handles, or organization-specific identifiers. Today these examples are stored in the internal `anonymizer.engine.constants.ENTITY_LABEL_EXAMPLES` mapping.

Users can technically mutate that mapping before calling `run()` or `preview()`, but this is not a supported API. It also changes process-global state, can leak between jobs with different requirements, may affect detection, substitution, and evaluation unexpectedly, and does not automatically activate a newly added label in the default detection set.

This came up while tuning detection for developer-agent task transcripts. Narrow positive examples could help distinguish vendor-prefixed API keys and actual account handles, while broad examples such as generic hexadecimal strings, UUIDs, and complete file paths could reinforce false positives such as `message_hash` values or repository names.

Expose per-label positive examples as supported, per-run user configuration. One possible API is:

```python
config = AnonymizerConfig(
    detect=Detect(
        entity_labels=["api_key", "user_name", "email"],
        entity_label_examples={
            "api_key": [
                "sk-ant-api03-abc123def456xyz",
                "OPENAI_API_KEY=sk-proj-abc123def456",
            ],
            "user_name": ["jsmith", "alice.chen", "@trudy-spies"],
        },
    ),
    replace=Annotate(),
)
```

Suggested behavior:

- Configuration is resolved independently for each run and never mutates module-level defaults.
- User examples merge with built-in examples by default. The merge/replace behavior should be documented explicitly.
- Label names are normalized consistently with `Detect.entity_labels`.
- When `entity_labels` is explicit, example-map keys must belong to that list; unmatched keys should produce a clear validation error.
- Adding examples for a new label must not silently add that label to the active detection set.
- Resolved examples are passed to the detection validator and augmenter prompts.
- Any use by substitute or evaluation workflows should be intentional and documented rather than occurring as a side effect of global state.
- Examples are documented as positive examples, not as a mechanism for expressing guaranteed exclusions.

Acceptance criteria:

- A public typed configuration field supports per-label examples.
- Two configurations with different examples can run in the same process without cross-run contamination.
- Validation covers empty labels/examples, normalized duplicate keys, and keys outside an explicit `entity_labels` list.
- Detection prompt tests verify that configured examples appear only for the corresponding active labels.
- Public documentation includes a dataset-specific example and explains when deterministic exclusions or `data_summary` are more appropriate.
- The bundled anonymizer agent skill is updated if its public configuration template is affected.

- Mutating `anonymizer.engine.constants.ENTITY_LABEL_EXAMPLES`. This works only as an internal, process-global workaround and is unsuitable as a supported recommendation.
- Encoding positive examples in `data_summary`. This mixes dataset context with label ontology and provides no structured validation.
- Using only custom label names. Label names constrain the ontology but do not demonstrate domain-specific value formats.
- Using a larger validation model. This may improve contextual reasoning but increases cost and does not replace explicit domain examples.

### Additional context

This complements, rather than replaces, the existing quality controls:

- Explicit `entity_labels` constrain which labels augmentation may emit.
- `data_summary` supplies contextual guidance, including negative distinctions, but does not enforce exclusions.
- `validation_excerpt_window_chars` provides additional surrounding context.
- Deterministic rules remain preferable for guaranteed exclusions such as known non-sensitive JSON fields.

memadi-nv:
## Summary
Adds an implementation plan for [issue #259](https://github.com/NVIDIA-NeMo/Anonymizer/issues/259), covering per-run entity-label examples for built-in and custom labels.
## Key decisions
- Merge configured examples with built-in examples without mutating global defaults.
- Automatically activate custom example labels alongside defaults when `entity_labels=None`.
- Keep explicit `entity_labels` lists strict.
- Pass:
  - label names only to GLiNER;
  - full resolved examples to the validator;
  - user-configured examples only to the augmenter.
- Let excluded labels take precedence, warning unless no effective labels remain.
- Keep substitution and evaluation behavior unchanged.
- Document prompt-size, typo-activation, export-sensitivity, and telemetry considerations.
The plan intentionally diverges from the issue in two places to improve usability and prompt efficiency: automatic custom-label activation and configured-only augmenter examples.

## Contributor Checklist
- [ ] PR title follows Conventional Commits, for example `fix: handle empty entity list`
- [ ] Related issue is linked, or a maintainer-owned no-issue reason is documented above
- [ ] For non-trivial changes, a plan document is linked above, or the no-plan reason is documented above
- [ ] Public API impact checked; `skills/anonymizer/SKILL.md` updated if needed
- [ ] No real PII added to tests, docs, notebooks, fixtures, or artifacts
- [ ] No API keys, service tokens, private keys, credentials, or real endpoint secrets added

## Validation

- Commands run:

- [ ] Docs updated, or not needed
- [ ] If tutorial sources changed: notebooks regenerated with `make convert-notebooks`
- [ ] If e2e, benchmark, or model-provider behavior changed: relevant validation is listed above


asteier2026 (approved):
Looks good to me

binaryaaron (commented):
> **agent** (`review-pr`): Two small observability/documentation suggestions and one nonblocking logging follow-up.

The shared resolver and execution/export paths look sound. I have two small changes to suggest:

1. **Report the effective detection scope.** With `entity_labels=None`, configuring examples for a new label activates it alongside the defaults, but measurement metadata and the scope log still report only the default set. For one added label, I reproduced 66 resolved labels versus 65 reported. Please report the effective count using the existing resolution logic, preserving existing metadata field meanings if needed.

2. **Document the repeated prompt overhead.** Each validator chunk receives the full resolved ontology, while the augmenter receives only configured examples. Please mention this alongside the examples documentation: reducing candidates per chunk does not bound example-related prompt overhead or total request cost. This overlaps Greptile’s comment.

One nonblocking follow-up: existing sync/async validation handlers and outer dependency paths log raw exception messages, which can contain sensitive response content. The new metadata-omission test doesn’t cover those paths. Please distinguish that test’s coverage from an end-to-end no-content-logging guarantee; broader sanitation can be tracked separately.

I also reviewed this against the proposed graph-native SDK design and found no architectural reason to delay the feature or add graph-specific abstractions here. The current result/evaluation behavior and run-local resolver can remain unchanged.

Validation: 251 targeted tests passed. No live provider tests were run.

plans/259-add-entity-label-examples.md
```diff
@@ -0,0 +1,330 @@
...
+
+- **Explicit label set:** labels supplied through `entity_labels`; it may include default and non-default labels.
+- **Default label:** a label present in `DEFAULT_ENTITY_LABELS`.
+- **Non-default label:** a label absent from `DEFAULT_ENTITY_LABELS`.
+- **Built-in examples:** examples shipped with Anonymizer in `ENTITY_LABEL_EXAMPLES`.
+- **Configured examples:** user-supplied positive examples in `entity_label_examples`; they may target default or non-default labels.
+
+## Intentional divergence and known limitations
+
+- Follow issue #259 by requiring every non-excluded example key to belong to the active label set. A non-default label must therefore be declared in an explicit `entity_labels` set; examples never activate it implicitly.
```
lipikaramaswamy: I think this is reasonable, especially since agents can read the rules and do this pretty easily when writing a confg.

plans/259-add-entity-label-examples.md
```diff
@@ -0,0 +1,330 @@
...
+- Document one intentional difference from issue #259: the augmenter receives only configured examples rather than the full resolved mapping of built-in and configured examples.
```
lipikaramaswamy: IIUC the reasoning for this is that currently the augment prompt includes no examples, so if we do have a custom list, then the augmenter should know what to look for beyond just a new label name

plans/259-add-entity-label-examples.md
```diff
@@ -0,0 +1,330 @@
...
+- Keep evaluation out of scope. Configured examples are not persisted on `AnonymizerResult`/`PreviewResult` or passed to evaluation, but an explicit label set containing non-default labels is persisted and scopes entity coverage consistently.
+- Treat example values as potentially sensitive configuration. They are embedded in validator/augmenter prompts, included in exported detection builders, and sent to configured model providers.
+  - Example: use synthetic `acme_live_abc123`, never a real production credential or customer identifier.
```
lipikaramaswamy: What do we mean here? Is it that users should be discouraged from passing real values? Or something else? The telemetry note below makes sense to me

memadi-nv: Yes, since the examples are sent to model providers. More explained as a warning in the docs [here](https://github.com/NVIDIA-NeMo/Anonymizer/pull/275/changes#diff-cedf06304c73b9284caa1756a133cd51c9bf223dfa30254800e22c87349bd395R164).

plans/259-add-entity-label-examples.md
memadi-nv: correct. Also clarified this in the docs, [here](https://github.com/NVIDIA-NeMo/Anonymizer/pull/275/changes/4b4a7fa343baa1ffcb1800274e3db8fd7caa07cf#diff-cedf06304c73b9284caa1756a133cd51c9bf223dfa30254800e22c87349bd395R169).

memadi-nv:
Thanks for the great reviews. I believe I addressed them all. Next steps after PR approval: 
- Remove the plan from the PR to make the CI/ Check pass.
- Make NVSkills CI tests pass.

